PR #45, opened 2020-07-21
[WIP] Enable seccomp filters

flxo:
Setting seccomp filters is enabled. Additional flush after dumping the
seccomp config ensures that the file is flushed before passing to
libminijail.


flxo:
Superseded by #114

PR #65, opened 2020-07-28
List implemented features in README

marcmo:


README.md
```diff
@@ -16,8 +16,13 @@ So far we tested Northstar on
 * 64-bit ARM
 * x86_64
 
-### Missing from implementation
+### Everything that is checked is implemented
```
flxo, line 19: Why not just name this sections features?

marcmo, line 19: @flxo we should list also the implemented features from a birds eye view...please help me to complete this list

README.md
```diff
@@ -16,8 +16,13 @@ So far we tested Northstar on
...
 
+- [x] On-the-fly verification of container content
```
flxo, line 21: Transparent integrity checks backed by [dm-verity](https://www.kernel.org/doc/html/latest/admin-guide/device-mapper/verity.html)

README.md
```diff
@@ -16,8 +16,13 @@ So far we tested Northstar on
...
+- [x] Process supervision: memory control
+- [x] Process supervision: cpu control
```
flxo, line 23: [CGroups](https://www.kernel.org/doc/Documentation/cgroup-v1/cgroups.txt) configuration support (memory, CPU)

README.md
```diff
@@ -16,8 +16,13 @@ So far we tested Northstar on
...
+- [x] Limiting system calls (whitelist)
```
flxo, line 24: [Seccomp](https://www.kernel.org/doc/Documentation/prctl/seccomp_filter.txt) configuration with whitelist approach

README.md
```diff
@@ -16,8 +16,13 @@ So far we tested Northstar on
 * 64-bit ARM
 * x86_64
```
flxo, line 17: I'd add Android as a target because it's quite different from `aarch64-unknown-linux-gnu`.
Why not list the target triples?

README.md
flxo: @marcmo 

I'd add `chroot` and `Process supervision`

README.md
marcmo: dm-verity is what we use...it's not the feature

README.md
marcmo: same thing...I'd say the memory and cpu control is the feature...cgroups are just what we use to implement it

README.md
marcmo: same

README.md
marcmo: yeah....better

README.md
flxo: Yes, but still worth to mention that it makes use of cgroups.

PR #68, opened 2020-07-28
Sextant ruby port

inorick:


sextant/src/main.rs
```diff
@@ -66,9 +75,360 @@ enum Opt {
...
+) -> Result<()> {
+    log::debug!("");
+    log::debug!("pack called with");
+    log::debug!("src_dir={}", src_dir.display());
+    log::debug!("registry_dir={}", registry_dir.display());
+    log::debug!("signing_key_name={}", signing_key_name);
+
+    // load manifest
+    let manifest_file_path = src_dir.join("manifest").with_extension("yaml");
+    let arch = "x86_64-unknown-linux-gnu"; // TODO: get as CLI parameter
```
flxo, line 193: We need to think how we handle this. Passing the arch can be fine. Checking for `root-<arch>`dirs in the sources can be one approach. This doesn't necessarily work for resources.
Open point.

sextant/src/main.rs
```diff
@@ -66,9 +75,360 @@ enum Opt {
...
+    let manifest_file = std::fs::File::open(&manifest_file_path)?;
+    log::debug!(
+        "read manifest file {}",
+        manifest_file_path.as_os_str().to_str().unwrap()
+    );
+    let manifest: Manifest = serde_yaml::from_reader(manifest_file)
+        .with_context(|| format!("Failed to parse {}", manifest_file_path.display()))?;
```
flxo, line 200: Yes. This is the way to do it ;-)

sextant/src/main.rs
```diff
@@ -66,9 +75,360 @@ enum Opt {
...
+
+    let tmp_dir = TempDir::new("")?;
```
flxo, line 202: context is missing here.

sextant/src/main.rs
```diff
@@ -66,9 +75,360 @@ enum Opt {
...
+        log::debug!(
+            "copy {} to {}",
+            root_dir.as_os_str().to_str().unwrap(),
+            tmp_dir.path().as_os_str().to_str().unwrap()
+        );
+        copy(&root_dir, &tmp_dir, &options)?;
+    }
+    if !tmp_root_dir.exists() {
+        log::debug!("mkdir {}", tmp_root_dir.as_os_str().to_str().unwrap());
+        fs::create_dir(&tmp_root_dir)?;
```
flxo, line 230: Add context.

sextant/src/main.rs
```diff
@@ -66,9 +75,360 @@ enum Opt {
...
+     * Add /lib and lib64 on Linux systems.
+     * Add /system on Android. */
+    let mut pseudo_files = vec![
+        ("/tmp", 444),
+        ("/proc", 444),
+        ("/dev", 444),
+        ("/sys", 444),
+        ("/data", 777),
+    ];
+    if arch == "aarch64-unknown-linux-gnu" || arch == "x86_64-unknown-linux-gnu" {
```
flxo, line 312: Maybe should introduce a `Arch` enum in `north_common`. The strings are used over and over and this get's too much. There should be `Display` implementation for this. Maybe there's a target triple crate? Maybe reexporting things from the `platforms` crate is a good idea.

sextant/src/main.rs
```diff
@@ -66,9 +75,360 @@ enum Opt {
...
+        pseudo_files.push(("/system", 444));
+    }
+
+    // create filesystem image
+    let squashfs_comp = match std::env::consts::OS {
+        "linux" => "gzip",
+        _ => "zstd",
+    };
+    if fs_type == &FsType::SQUASHFS {
+        let mut cmd = Command::new("mksquashfs");
```
flxo, line 325: The `mksquashfs` call shall be wrapped in a struct. Use the Builder pattern to add arguments like `.pseudo_file(t: PseudoType, p: &Path, permissions: ....)`

sextant/src/main.rs
```diff
@@ -66,9 +75,360 @@ enum Opt {
...
+                uid,
+                gid
+            ));
+        }
+        dbg!(&cmd);
+        let cmd_output = cmd.output()?;
+        println!("status={}", cmd_output.status);
+        println!("stdout={}", String::from_utf8_lossy(&cmd_output.stdout));
+        println!("stderr={}", String::from_utf8_lossy(&cmd_output.stderr));
+    } else if fs_type == &FsType::EXT4 {
```
flxo, line 348: Skip `ext4` for now. Movement clearly goes into the direction of squashfs. The ext4 support is barely tested and probably removed and added in a clean way when we need it (maybe);

sextant/src/main.rs
```diff
@@ -66,9 +79,455 @@ enum Opt {
...
+        Opt::Pack { dir, out } => pack_cmd(&dir, &out),
+        _ => {
+            unimplemented!();
+        }
+    }
+}
+
+#[derive(Copy, Clone, PartialEq)]
+enum FsType {
+    SQUASHFS,
```
marcmo, line 101: we only need to consider SquashFS

sextant/src/main.rs
```diff
@@ -66,9 +75,360 @@ enum Opt {
...
+fn pack_containers(
+    registry_dir: &Path,
+    container_src_dir: &Path,
+    key_dir: &Path,
+    signing_key_name: &str,
+    fs_type: &FsType,
+    uid: u32,
+    gid: u32,
+) -> Result<()> {
+    log::debug!("");
```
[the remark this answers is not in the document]
marcmo: like this:
```
use log::debug;
...
{
    debug!(""):
...
```

sextant/src/main.rs
```diff
@@ -66,9 +79,455 @@ enum Opt {
...
+    log::debug!("pack_containers called with");
+    log::debug!("registry_dir={}", registry_dir.display());
+    log::debug!("container_src_dir={}", container_src_dir.display());
+    log::debug!("key_dir={}", key_dir.display());
+    log::debug!("signing_key_name={}", signing_key_name);
+
+    let sign_key_path = key_dir.join(signing_key_name).with_extension("key");
+    log::debug!("open file {}", sign_key_path.display());
+    let sign_key = File::open(&sign_key_path)?;
+    log::debug!("sign_key.len()={}", sign_key.metadata().unwrap().len());
```
marcmo, line 142: no `unwrap()` should remain in the code. If you use this in a function that returns a result, you should be able to use the `?` operator

sextant/src/main.rs
```diff
@@ -66,9 +79,455 @@ enum Opt {
...
+    signing_key: &File,
+    signing_key_name: &str,
+    fs_type: FsType,
+    uid: u32,
+    gid: u32,
+) -> Result<()> {
+    let component_name = src_dir
+        .components()
+        .last()
+        .unwrap()
```
marcmo, line 177: same

bzld:
Related: A feature we don't have currently, but is almost required:
Have a `sextant generate_key $NAME` command which calls `sodiumoxide::crypto::sign::ed25519::gen_keypair()` and writes the `keys/$NAME.key` and `keys/$NAME.key.pub` files.

sextant/Cargo.toml
```diff
@@ -9,3 +9,10 @@ anyhow = "1.0.31"
 log = "0.4.8"
 env_logger = "0.7.1"
 structopt = "0.3.14"
+glob = "0.3.0"
```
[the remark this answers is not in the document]
inorick: Thanks, appreciate it :+1:

north/src/manifest.rs
```diff
@@ -140,22 +142,31 @@ pub struct Manifest {
     /// Target arch
     pub arch: String,
     /// Path to init
+    #[serde(skip_serializing_if = "Option::is_none")]
```
marcmo, line 145: why is there any case where we want to skip serialization of the manifest @inorick ?

flxo: For yaml a option can be either not preset or set to `null`. Semantically this is the same. Skipping `None`s during serialisation reduces the size of the manifest. In the examples we also do not have `None`s.

examples/build_examples.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/sh
```
marcmo, line 1: I'd add
```
set -o errexit
set -o pipefail
set -o nounset
```
to fail hard and prevent usage of undeclared variables

rakefile.rb
```diff
@@ -182,11 +183,17 @@ def examples
...
+        # create_arch_package(target_arch, target_dir, dir, registry, package_config)
       end
     end
   end
 
+  def create_arch_package_sextant(container_src, out_dir, target_arch)
+    puts "target arch was: #{target_arch}"
+    # sh "target/release/sextant pack -d #{container_src} -o #{out_dir} -k #{KEY_FILE}"
+  end
+
```
marcmo, line 196: I wouldn't change this in the pr.
If the shellscript is to be used, get rid of the rake task

sextant/src/npk.rs
```diff
@@ -0,0 +1,427 @@
...
+};
+use tempdir::TempDir;
+use uuid::Uuid;
+
+const DIGEST_SIZE: usize = 32;
+const BLOCK_SIZE: u64 = 4096;
+const UID: u32 = 1000;
+const GID: u32 = 1000;
+
+/// out_dir: where the npk should be packaged to (usually the registry directory)
```
marcmo, line 26: missing: documentation of parameters

sextant/src/npk.rs
```diff
@@ -0,0 +1,427 @@
...
+
+    // create filesystem image
+    let fsimg_path = &tmp_dir.path().join("fs").with_extension("img");
+    let pseudo_files = gen_pseudo_files_list(&manifest)
+        .with_context(|| anyhow!("Cannot generate list of pseudo files"))?;
+    create_squashfs(&tmp_root_dir, &fsimg_path, &pseudo_files)
+        .with_context(|| anyhow!("Cannot generate squashfs"))?;
+    let fsimg_size = fs::metadata(&fsimg_path)
+        .with_context(|| anyhow!("Cannot get fsimg file metadata"))?
+        .len();
```
marcmo, line 58: extract to it's own function

sextant/src/npk.rs
```diff
@@ -0,0 +1,427 @@
...
+        &File::open(&fsimg_path).with_context(|| anyhow!("Cannot open file fsimg.img"))?,
+        fsimg_size,
+        BLOCK_SIZE,
+        &salt,
+        &hash_level_offsets,
+        tree_size,
+    )
+    .with_context(|| anyhow!("Error while generating hash tree"))?;
+    write_verity_header(&fsimg_path, fsimg_size, &salt, &hash_tree)
+        .with_context(|| anyhow!("Error while writing verity header"))?;
```
marcmo, line 78: same...create function for it

sextant/src/npk.rs
```diff
@@ -0,0 +1,427 @@
...
+    let hashes_yaml = gen_hashes_yaml(&tmp_manifest_dir, &fsimg_path, fsimg_size, verity_hash)?;
+    let signatures_yaml = sign_hashes(&signing_key, &hashes_yaml);
+    write_zip_file(
+        &out_dir,
+        &platform,
+        &manifest,
+        &fsimg_path,
+        &signatures_yaml,
+    )
+    .with_context(|| anyhow!("Error while writing zip file"))?;
```
marcmo, line 92: same...create a function for it

sextant/src/npk.rs
```diff
@@ -0,0 +1,427 @@
...
+use uuid::Uuid;
+
+const DIGEST_SIZE: usize = 32;
+const BLOCK_SIZE: u64 = 4096;
+const UID: u32 = 1000;
+const GID: u32 = 1000;
+
+/// out_dir: where the npk should be packaged to (usually the registry directory)
+pub fn pack(src_dir: &Path, out_dir: &Path, key_file_path: &Path, platform: &str) -> Result<()> {
+    // write platform to manifest
```
marcmo, line 28: Add a check that verifies that the squashfs tools are installed, otherwise fail with a descriptive error message

sextant/src/npk.rs
```diff
@@ -0,0 +1,427 @@
...
+    let options =
+        zip::write::FileOptions::default().compression_method(zip::CompressionMethod::Stored);
+    let mut zip = zip::ZipWriter::new(&npk_file);
+    zip.start_file("signature.yaml", options)?;
+    zip.write_all(signatures_yaml.as_bytes())?;
+    zip.start_file("manifest.yaml", options)?;
+    let manifest_string = serde_yaml::to_string(&manifest)?;
+    zip.write_all(manifest_string.as_bytes())?;
+
+    let offset = 43 + manifest_string.len() + 44 + signatures_yaml.len() + 36; // stored
```
marcmo, line 401: introduce constants for the magic numbers

sextant/src/npk.rs
```diff
@@ -0,0 +1,427 @@
...
+    let padding = (offset / 4096 + 1) * 4096 - offset;
```
marcmo, line 402: same as above

sextant/src/npk.rs
inorick: It is not clear if adding this offset is even necessary. The containers I tested work even without extra data.

sextant/src/npk.rs
inorick: It is not clear if adding this offset is even necessary. The containers I tested work even without extra data.

sextant/src/main.rs
inorick: Arch type is gone now.

flxo:
@inorick `sodium` has some optional features which probably can be turned off in this use case:

examples/build_examples.sh
```diff
@@ -0,0 +1,85 @@
...
```
marcmo, line 1: you are using bash syntax -> you need to use `#!/bin/bash`

examples/build_examples.sh
```diff
@@ -0,0 +1,85 @@
...
+    fi
+
+    echo "Creating NPK (${name})"
+    cargo run --bin sextant -- \
+      pack \
+      --dir "${TMP_DIR}" \
+      --out "${REGISTRY_DIR}" \
+      --key "${EXAMPLE_PRV_KEY}" \
+      --platform "${platform}"
+  done
```
marcmo, line 84: for whatever reason I need to use `sudo` to execute the script. This should not be so.
this is what I get on the "Creating NPK..." step:
```
Creating NPK (cpueater)
    Finished dev [unoptimized + debuginfo] target(s) in 0.28s
     Running `sudo -E target/debug/sextant pack --dir /tmp/tmp.fyBxWO6XjE --out ./target/north/registry --key ./examples/keys/north.key --platform aarch64-linux-android`
[sudo] password for muellero: 
```

flxo: See #126 . The runner is currently global for `x86_64-unknown-gnu-linux`.

rakefile.rb
```diff
@@ -182,7 +183,7 @@ def examples
           sh "cross build --release --bin #{name} --target #{target_arch}"
           cp "target/#{target_arch}/release/#{name}", target_dir
         end
-        create_arch_package(target_arch, target_dir, dir, registry, package_config)
+        create_arch_package_sextant(dir, registry, target_arch)
```
marcmo, line 186: the rake task is now broken. Until we fully commit to get rid of rake, this should stay the entry point to using sextant

marcmo (changes requested):
1.Add description of `sextant pack` command: missing: where is the manifest expected?

2.the error messages for `pack` are not clear enough:
```
sextant     pack     --dir /home/muellero/dev/git/northstar/examples/container/cpueater/root-aarch64-linux-android     --out /home/muellero/dev/git/northstar/target/north/registry     --key /home/muellero/dev/git/northstar/examples/keys     --platform aarch64-linux-android 
Error: Cannot read manifest

Caused by:
    No such file or directory (os error 2)
```
and
```
sextant     pack     --dir /home/muellero/dev/git/northstar/examples/container/crashing     --out /home/muellero/dev/git/northstar/target/north/registry     --key /home/muellero/dev/git/northstar/examples/keys/north.key     --platform aarch64-linux-android
Error: Cannot copy root dir to tmp

Caused by:
    Src root dir does not exist
```
3. Call sextant from rakefile. It's not deprecated yet.

marcmo (approved):

PR #71, opened 2020-07-31
Support versioned resources

### LINKED ISSUE #57: Resource mounting relies on `cwd`
The code for mounting resources relies on a specific `cwd` which should not. This is bad. Next need for having multiple bind_mounts is not obvious and needs some documentation.
A minor: the logs doesn't match the log style used in `north`.

```rust
        if let Some(resources) = &container.manifest.resources {
            for res in resources {
                if let Ok(cwd) = std::env::current_dir() {
                    let dir_in_container_path = std::path::Path::new(&res.dir);
                    let first_part_of_path = cwd
                        .join(SETTINGS.directories.run_dir.to_owned())
                        .join(res.name.to_owned());
                    let src_dir = match dir_in_container_path.strip_prefix("/") {
                        Ok(dir_in_resource_container) => {
                            first_part_of_path.join(dir_in_resource_container)
                        }
                        Err(_) => first_part_of_path,
                    };
                    info!(
                        "Mounting from src_dir {} to target {:?}",
                        src_dir.display(),
                        res.mountpoint
                    );
                    mount_bind(&mut jail, src_dir.as_ref(), &res.mountpoint, false)?;
                }
            }
        }
```
@marcmo Can you please take care?

### LINKED ISSUE #67: Resource dependencies ignore version
When making use of a resource the version is ignored. To be able to respect the version the mount points of *all* containers must be done in a version subdir to allow multiple in parallel e.g `<run_dir>/hello/v0.0.1`. The bind mounts of resources must take care to select the correct version.

marcmo:
* Resources can now exist in multiple versions (fixes #67)
* Fix tooling to put resources in version folder
* Added rake task to view registry
* removed double mounting (resources were mounted twice)
* Updated examples using resources
* Resource mounting does not rely on cwd anymore (fixes #57)


tooling.rb
```diff
@@ -158,17 +166,17 @@ def create_npk(src_dir, npk, manifest, arch_dir, pack_config)
...
-    when 'aarch64-unknown-linux-gnu', 'x86_64-unknown-linux-gnu'
-      pseudofiles << ['/lib', 444]
-      pseudofiles << ['/lib64', 444]
-    when 'aarch64-linux-android'
-      pseudofiles << ['/system', 444]
+      # The list of pseudofiles is target specific.
+      # Add /lib and lib64 on Linux systems.
+      # Add /system on Android.
+      case arch
+      when 'aarch64-unknown-linux-gnu', 'x86_64-unknown-linux-gnu'
```
bzld, line 174: Maybe this should be the default branch, so it won' fail for other targets.

tooling.rb
```diff
@@ -188,7 +196,7 @@ def create_npk(src_dir, npk, manifest, arch_dir, pack_config)
     end.join(' ')
     # TODO: The compression algorithm should be target and not host specific!
     squashfs_comp = OS.linux? ? 'gzip' : 'zstd'
-    shell "mksquashfs #{tmp_root} #{fsimg} -all-root -comp #{squashfs_comp} -no-progress -info #{pseudofiles}"
+    shell "mksquashfs #{folder_to_package} #{fsimg} -all-root -comp #{squashfs_comp} -no-progress -info #{pseudofiles}"
```
bzld, line 199: So we put all files inside a folder `/$VERSION/`? Does that make sense?
Shouldn't we keep them in the root of the filesystem and just have the squashfs mount target folder versioned?

Example:
- Resource container `resource` contains just `test.json`, versions `1.0.0` and `1.2.0` exist.
- Resource container `1.0.0` NPK filesystem content: `/test.json`
- Resource container `1.2.0` NPK filesystem content: `/test.json`
- North mounts `resource_v1.0.0.npk` to `<run_dir>/resource/1.0.0/` (file at `<run_dir>/resource/1.0.0/test.json`)
- North mounts `resource_v1.2.0.npk` to `<run_dir>/resource/1.2.0/` (file at `<run_dir>/resource/1.2.0/test.json`)
- Container A uses `1.0.0` of resource container at `/res`: North bind-mounts `<run_dir>/resource/1.0.0/` to `<container A root>/res`
- Container B uses `1.2.0` of resource container at `/res`: North bind-mounts `<run_dir>/resource/1.2.0/` to `<container B root>/res`

marcmo: in the original resource container we do not have a version folder, just the usual structure. So it always looks like this:
```
examples/container/resource/hello_message
├── manifest.yaml
└── root
    ├── hello
    └── pad
```
if we increase the version, the version field in the manifest is changed along with all the changes the new version brings.
but if we put it into the squashfs image, it looks like this:
```
squashFS-image: fs.img
Parallel unsquashfs: Using 2 processors
2 inodes (41 blocks) to write

squashfs-root
squashfs-root/0.1.2
squashfs-root/0.1.2/hello
squashfs-root/0.1.2/pad
```
does that make sense?

north/src/process.rs
```diff
@@ -440,11 +416,12 @@ fn wait_for_exit(name: &str, pid: u32, exit_tx: sync::Sender<i32>, event_tx: Eve
...
-    let cwd = std::env::current_dir()?;
+    let run_dir: &std::path::Path = std::path::Path::new(&SETTINGS.directories.run_dir);
     let dir_in_container_path = std::path::Path::new(&res.dir);
-    let first_part_of_path = cwd
-        .join(SETTINGS.directories.run_dir.to_owned())
-        .join(res.name.to_owned());
+    let first_part_of_path = run_dir
+        .join(res.name.to_owned())
+        .join(format!("{}", res.version));
+
```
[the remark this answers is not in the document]
marcmo: 👍

north/src/npk/linux.rs
```diff
@@ -127,7 +127,7 @@ async fn install_internal(
     let start = time::Instant::now();
 
     if let Some(npk_name) = npk.file_name() {
-        info!("Loading {}", npk_name.to_string_lossy());
+        info!("Installing {}", npk_name.to_string_lossy());
```
bzld, line 130: To be honest, "Installing" sounds like a one-time-operation after download, not something done on every boot...

north/src/npk/linux.rs
```diff
@@ -370,12 +376,12 @@ async fn losetup(
         .context("Failed to acquire free loopdev")?;
 
     debug!(
-        "Using loop device {}",
+        "Created loop device {}",
```
bzld, line 379: It could be "just" reused, if it was already precreated by kernel...

PR #78, opened 2020-08-12
Simple GHA Workflow to build docker images

meatlink:


.github/workflows/build.yml
```diff
@@ -0,0 +1,39 @@
...
+              source $HOME/.cargo/env
+
+              cargo install cross
+
+              echo "${TOKEN}" | docker login https://docker.pkg.github.com \
+                -u "$GITHUB_REPOSITORY_OWNER" --password-stdin
+              
+              sudo apt install -y gnupg2
+
+              gpg2 --keyserver hkp://pool.sks-keyservers.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
```
marcmo, line 32: what is this needed for?

meatlink: I think it's required to install rvm
https://rvm.io/rvm/install

ruby can also be installed without rvm if necessary

PR #79, opened 2020-08-13
Verify NPK signatures

bzld:
- Copy public keys to target/keys folder in rakefile
- Add `key_dir` setting
- Load public keys on runtime startup (TODO: Move this out of main.rs?)
- Verify `signature.yaml` signature


north/src/npk/mod.rs
```diff
@@ -84,8 +80,18 @@ impl Hashes {
...
         debug!("Using key {}", signature.key);
-
-        // TODO verify hashes
+        let key = state
+            .signing_key(&signature.key)
+            .ok_or_else(|| anyhow!("Key {} not found", &signature.key))?;
+        let key = sign::PublicKey::from_slice(&key).ok_or_else(|| anyhow!("Key {} is invalid"))?;
+
+        if !sign::verify_detached(
+            &sign::Signature::from_slice(&signature.signature).unwrap(),
```
inorick, line 89: Could be read like 'key' with '?' and no '.unwrap()'

north/Cargo.toml
```diff
@@ -17,6 +17,7 @@ async-trait = "0.1.36"
 base64 = "0.12.3"
 bitflags = "1.2.1"
 config = "0.10.1"
+ed25519-dalek = "1.0.0-pre.4"
```
marcmo, line 20: is there a non-alpha version also?

north/src/npk/mod.rs
```diff
@@ -63,13 +64,8 @@ struct Signature {
...
 impl Hashes {
     #[allow(unused)]
-    pub fn from_str(file: &str) -> Result<Hashes, Error> {
-        // sha256 of signature.yaml
-        let mut hasher = sha2::Sha256::new();
-        hasher.update(file.as_bytes());
-        let _digest = format!("{:x}", hasher.finalize());
-
-        let mut docs = file.split("---");
+    pub fn from_str(file: &str, state: &State) -> Result<Hashes, Error> {
```
marcmo, line 67: Passing the complete state seems unnecessary...why not only the needed keys?

north/src/main.rs
```diff
@@ -197,6 +203,28 @@ async fn main() -> Result<()> {
...
+                    let key = PublicKey::from_bytes(&key_bytes)?;
+                    signing_keys.insert(key_id.to_string_lossy().to_string(), key);
+                    info!("Loaded signing key {:?}", key_id);
+                }
+            }
+        }
+    }
+    Ok(signing_keys)
+}
+
```
marcmo, line 227: Do we need to support more than 1 key?

north/src/npk/mod.rs
marcmo: I see...we don't know what key to pass without reading the signature file

north/src/main.rs
```diff
@@ -19,10 +19,14 @@
 extern crate structure;
 
 use anyhow::{Context, Error, Result};
+use async_std::path::PathBuf;
+use async_std::prelude::*;
```
marcmo, line 23: should be formated like `use async_std::{fs, path::PathBuf, prelude::*, sync};`
best use `rake format`

north/src/npk/mod.rs
marcmo: this breaks our tests, best run `rake check`
```
error[E0061]: this function takes 2 arguments but 1 argument was supplied
   --> north/src/npk/mod.rs:180:13
    |
66  |     pub fn from_str(file: &str, state: &State) -> Result<Hashes, Error> {
    |     ------------------------------------------------------------------- defined here
...
180 |     let s = Hashes::from_str(signature).expect("Failed to parse signature");
    |             ^^^^^^^^^^^^^^^^ --------- supplied 1 argument
    |             |
    |             expected 2 arguments
```

north/Cargo.toml
bzld: The 0.9 version seems broken because rand crate has upgraded meanwhile.
Also its API is different than for 1.0(-pre).
Let's just hope they get to final 1.0.0 soon.

north/src/main.rs
bzld: I think that's a nice flexibility ;)

north/src/main.rs
bzld: I just ran `cargo fmt`, but apparently `rake format` is slightly stricter in exactly this spot.

north/src/npk/mod.rs
bzld: Fixed - by just passing the HashMap

PR #83, opened 2020-08-17
Support musl targets

bzld:


bzld:
Seems to work fine on aarch64 target: both musl-based runtime and musl-based container binaries work.

Issue #86, opened 2020-09-04
Detect group mount points by parsing /proc/mounts


flxo:
Found this in the firecracker docs:

```
 (they should be mounted by the user otherwise). The jailer will parse /proc/mounts to detect where each of the three controllers can be found (multiple controllers may share the same path). For each identified location (referred to as <cgroup_base>), the jailer creates the
```

This sounds like a good approach and we can get rid of at least two config options.

PR #88, opened 2020-09-15
Detect group mount points by parsing /proc/mounts

mmoessmer:


flxo:
@mmoessmer can you have a look at [proc-mounts](https://crates.io/crates/proc-mounts). This crate internal caches a list of mounts that are cheap to access. This avoids an additional `lazy_static` in our codebase. From configuration point of view the approach to configure the sub directories aka groups is fine. I would just have a dedicated one per group subsystem and not just one. Reason: the cgroup subsystem operate very independent from each other and therefore should also be treated individually.

flxo:
Fixes #86

PR #104, opened 2020-09-29
Separate north bin and lib. Refactor module layout

### LINKED ISSUE #91: Split runtime into lib and bin part
Split into lib and bin. The iterating part here is the handling of the configuration that cannot be statically read somewhere and must be passed from the bin.

flxo:
Fixes #91


north/src/runtime/linux/cgroups.rs
```diff
@@ -12,7 +12,8 @@
 //   See the License for the specific language governing permissions and
 //   limitations under the License.
 
-use crate::{Event, EventTx, SETTINGS};
+use super::super::super::runtime::{config, Event, EventTx};
```
marcmo, line 15: maybe nicer:
```
use crate::runtime::{config, Event, EventTx};
```

north/src/runtime/linux/mount.rs
```diff
@@ -120,44 +120,36 @@ pub async fn mount(
...
+    let c_data = data.as_ref().map_or(ptr::null(), |d| d.as_ptr()) as *const c_void;
+    let result = unsafe {
+        libc::mount(
+            c_source.as_ptr(),
+            c_target.as_ptr(),
+            c_fstype.as_ptr(),
+            flags.bits(),
+            c_data,
+        )
+    };
```
marcmo, line 133: why not spawn on the thread pool without blocking the runtime?

north/src/runtime/state.rs
```diff
@@ -283,27 +297,27 @@ impl State {
...
+    //         npk::uninstall(app.container())
+    //             .await
+    //             .map_err(Error::UninstallationError)?;
+    //         self.applications.remove(&app.manifest().name);
+    //         Ok(())
+    //     } else {
+    //         warn!("Cannot uninstall running container {}", app);
+    //         Err(Error::ApplicationRunning(vec![app.manifest().name.clone()]))
+    //     }
+    // }
```
marcmo, line 320: missing in commit message

north/src/runtime/linux/cgroups.rs
flxo: The idea is to have inner module `use` directive relative and only needed ones absolute based on the crate root.

north/src/runtime/linux/mount.rs
flxo: There's a issue with calling `libc::mount` from a `spawn_blocking` future. @alfonsoros88 investigates here. The mounting code will be refactored in #106 so calling the mounts in the non blocking context here is temporary fine.

north/src/runtime/state.rs
flxo: Yes. I think I commented it because the install/uninstall/update thing is a subject to be totally reworked.

marcmo: agreed...not part of the pr though

north/src/runtime/linux/cgroups.rs
marcmo: ok...looks weird but you are the expert

marcmo:
could you please increase the version and add a changelog entry @flxo ?

flxo:
> could you please increase the version and add a changelog entry @flxo ?

Done and rebased.

PR #107, opened 2020-10-01
pr34: Add shell script to check kernel config

pclesr:
The shell script is meant as a guide for people configuring
their kernel to run northstar. It attempts to check specific
functionality provided by the various kernel config options.


flxo:
Check on one of our regular targets.

PR #114, opened 2020-10-07
Enable seccomp filter for minijailed processes

flxo:


flxo:
Weird: something is wrong with the configuration:

PR #117, opened 2020-10-08
Return non zero code when errors happen in north

### LINKED ISSUE #42: Make north exit with a non zero exit code on fatal errors
The runtime exits with exit code `0` even in error conditions. Extend the handling of `Event::Error` and `break` with e.g `1`. Change the signature of `main`. Do *not* do a `std::process::exit` in  the main loop when receiving a `Event::Error`.

alfonsoros88:
fixes #42

### DIFF
--- a/north/src/main.rs
+++ b/north/src/main.rs
@@ -14,11 +14,12 @@
 
 #![deny(clippy::all)]
 
-use anyhow::Result;
+use anyhow::{Context, Result};
 use async_std::{fs::read_to_string, path::PathBuf};
-use log::info;
+use log::{error, info};
 use north::runtime;
 use runtime::config::Config;
+use std::process;
 use structopt::StructOpt;
 
 #[derive(Debug, StructOpt)]
@@ -34,9 +35,22 @@ struct Opt {
 }
 
 #[async_std::main]
-async fn main() -> Result<()> {
+async fn main() {
+    process::exit(match run().await {
+        Ok(()) => 0,
+        Err(err) => {
+            error!("{}", err);
+            1
+        }
+    })
+}
+
+async fn run() -> Result<()> {
     let opt = Opt::from_args();
-    let config: Config = toml::from_str(&read_to_string(&opt.config).await?)?;
+    let config: Config = toml::from_str(&read_to_string(&opt.config).await?).context(format!(
+        "Failed to read configuration file {}",
+        opt.config.display()
+    ))?;
 
     let log_filter = if opt.debug || config.debug {
         "north=debug"


north/src/main.rs
```diff
@@ -33,10 +33,27 @@ struct Opt {
...
 }
 
+async fn parse_config(path: &PathBuf) -> Result<Config> {
+    toml::from_str(&read_to_string(path).await?).context(format!(
+        "Failed to read configuration file {}",
+        path.display()
+    ))
+}
+
+fn unwrap_or_exit<T>(result: Result<T>) -> T {
```
flxo, line 43: A very common approach is to have a 
```rust
async run() -> Result<i32> {
    // read the config
   // run the runtime...
   // etc
}
```
Within that you can use `?` as usual and the match on the `Result` with the `exit` appears only once.

north/src/main.rs
```diff
@@ -33,10 +33,27 @@ struct Opt {
     pub debug: bool,
 }
 
+async fn parse_config(path: &PathBuf) -> Result<Config> {
```
flxo, line 36: We can get rid of this one line `fn` with the `run` function mentioned below.

north/src/main.rs
```diff
@@ -33,10 +33,27 @@ struct Opt {
...
+        path.display()
+    ))
+}
+
+fn unwrap_or_exit<T>(result: Result<T>) -> T {
+    match result {
+        Ok(t) => t,
+        Err(err) => {
+            eprintln!("{:?}", err);
+            std::process::exit(1);
```
flxo, line 48: Add a `use std::process`.

north/src/main.rs
```diff
@@ -33,10 +33,27 @@ struct Opt {
...
+        "Failed to read configuration file {}",
+        path.display()
+    ))
+}
+
+fn unwrap_or_exit<T>(result: Result<T>) -> T {
+    match result {
+        Ok(t) => t,
+        Err(err) => {
+            eprintln!("{:?}", err);
```
flxo, line 47: Think the `anyhow::Error` impls `Display` which give nicer output.

north/src/main.rs
alfonsoros88: The config.debug is needed to configure the logging, that's why we have this double check for a possible exit(1).

north/src/main.rs
alfonsoros88: But in this case we need to parse the config to set the logging in main before calling run :(

flxo: Why? Printing the error from `run` should be done via a direct `eprintln` anyway to be independent from the logger.

alfonsoros88: because the log filter is set via either opt.debug || config.debug. The latter coming from the parsed config.

PR #120, opened 2020-10-09
Shutdown north when SIGINT or SIGTERM is received

### LINKED ISSUE #103: Make sure all resources are cleaned up when the north runtime exits
Exit scenarios:
- graceful shutdown
- killed by SIGINT
- crash because of panic (panic handler needed)

alfonsoros88:
fixes #103

### DIFF
--- a/north/src/runtime/mod.rs
+++ b/north/src/runtime/mod.rs
@@ -35,6 +35,9 @@ use tokio::{
     sync::{self, oneshot},
 };
 
+#[cfg(any(target_os = "android", target_os = "linux"))]
+use tokio::signal::unix::{signal, SignalKind};
+
 pub type EventTx = mpsc::Sender<Event>;
 
 #[allow(clippy::large_enum_variant)]
@@ -123,6 +126,9 @@ pub async fn run(config: &Config) -> Result<(), Error> {
         }
     }
 
+    #[cfg(any(target_os = "linux", target_os = "android"))]
+    enable_signaled_shutdown(event_tx.clone()).await;
+
     // Initialize console
     let console = console::Console::new(&config.console_address, &event_tx);
     // Start to listen for incoming connections
@@ -158,8 +164,7 @@ pub async fn run(config: &Config) -> Result<(), Error> {
     }
 
     info!("Shutting down...");
-
-    Ok(())
+    state.tear_down().await
 }
 
 /// This is a starting point for doing something meaningful with the childs outputs.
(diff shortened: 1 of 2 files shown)


north/src/runtime/mod.rs
```diff
@@ -113,6 +113,22 @@ pub async fn run(config: &Config) -> Result<()> {
         state.applications.len()
     );
 
+    // handle SIGINT and SIGTERM
+    let tx_interrupt = tx.clone();
+    ctrlc::set_handler(move || {
```
flxo, line 118: Why not moving a copy of the tx handle of the main event loop to the handler? This would be way simpler.

flxo, line 118: Installing a signal handler catches a signal. Think this means that the signal is not sent to childs? This is something we would need.

north/src/runtime/mod.rs
```diff
@@ -113,6 +113,22 @@ pub async fn run(config: &Config) -> Result<()> {
...
+            payload: api::Payload::Request(api::Request::Shutdown),
+        };
+        let (tx, rx) = sync::channel::<api::Message>(1);
+        async_std::task::block_on(tx_interrupt.send(Event::Console(request, tx)));
+
+        if async_std::task::block_on(rx.recv()).is_err() {
+            panic!("Failed to shutdown north");
+        }
+    })
+    .expect("Failed to set signal handler");
```
flxo, line 130: No need to expect something in a fn that return `Result`.

north/src/runtime/mod.rs
alfonsoros88: this is exactly what this is doing, isn't it?

north/Cargo.toml
```diff
@@ -45,6 +45,7 @@ toml = { version = "0.5.6", optional = true }
 url = { version = "2.1.1", optional = true }
 uuid = { version = "0.8.1", features = [ "v4"], optional = true }
 zip = { version = "0.5.6", optional = true }
+ctrlc = { version = "3.1.6", features = ["termination"] }
```
flxo, line 48: nit: sort dependencies alphanumerically

Why is this a feature? Think we can have this all the time.

north/src/runtime/mod.rs
```diff
@@ -113,6 +113,13 @@ pub async fn run(config: &Config) -> Result<()> {
         state.applications.len()
     );
 
+    // Handle SIGINT and SIGTERM
+    let tx_interrupt = tx.clone();
+    ctrlc::set_handler(move || {
```
flxo, line 118: This won't compile if feature `termination` is not enabled and it's not enabled by default.

north/src/runtime/state.rs
```diff
@@ -319,6 +311,20 @@ impl State {
         }
     }
 
+    /// Stop and uninstall all the applications
+    pub async fn stop_and_uninstall_all(&mut self) -> result::Result<(), Error> {
```
flxo, line 315: I'd rename this to `shutdown`

flxo (changes requested):

north/src/runtime/state.rs
alfonsoros88: That would conflict with the console shutdown function. The difference is that shutdown won't do anything if any container is running. Maybe "cleanup" or "tear_down" is a better name for this function?

north/src/runtime/mod.rs
alfonsoros88: It does compile without the `termination` feature. It will just not handle `SIGTERM` as well.

north/Cargo.toml
alfonsoros88: This  feature makes the handle catch `SIGTERM` in addition to `SIGINT`.

PR #128, opened 2020-10-16
Merge resources and mounts in manifest

### LINKED ISSUE #123: Merge resources and mounts in manifest
The `resource` section is defecto a list of bind mounts and should be merged with `mounts`

A proposal is: 
```
mounts:
  /lib:
    host: /lib
  /lib64:
    host: /lib64
  /system:
    host: /system
  /bin:
    resource: ferris:0.0.2
  /message:
    resource: hello_message:0.1.2/some_subdir
  /cache:
    tmpfs: size=25M
```

In order to safely load the manifest the mounts need to parsed and "manually" converted to the struct/enum used in the manifest to fail early on malformed manifests.

alfonsoros88:
closes #123 supersedes #127

### DIFF
--- a/north/src/manifest.rs
+++ b/north/src/manifest.rs
@@ -14,9 +14,10 @@
 
 use anyhow::{anyhow, Context, Error, Result};
 use async_std::{path::Path, task};
+use lazy_static::lazy_static;
 use serde::{
     de::{Deserializer, Visitor},
-    ser::Serializer,
+    ser::{SerializeMap, Serializer},
     Deserialize, Serialize,
 };
 use std::{
@@ -123,31 +124,6 @@ pub struct CGroups {
     pub cpu: Option<CGroupCpu>,
 }
 
-// TODO: Remove?
-#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
-pub struct Resource {
-    pub name: Name,
-    pub version: Version,
-    pub dir: std::path::PathBuf,
-    pub mountpoint: std::path::PathBuf,
-}
-
-impl fmt::Display for Resource {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        write!(f, "Resource \"{} ({})\"", self.name, self.version)
-    }
-}
-
-#[derive(Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
-pub enum MountType {
-    /// Bind mount
-    #[serde(rename = "bind")]
-    Bind,
-    /// A container specific directory created by the runtime in directories.data_dir
-    #[serde(rename = "data")]
-    Data,
-}
-
 #[derive(Clone, Eq, PartialEq, Debug, Hash, Serialize, Deserialize)]
 pub enum MountFlag {
     /// Bind mount
@@ -161,16 +137,23 @@ pub enum MountFlag {
     // NoSuid,
 }
 
-#[derive(Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
-pub struct Mount {
-    /// Mount source
-    pub source: Option<std::path::PathBuf>,
-    /// Mount target
-    pub target: std::path::PathBuf,
-    /// Mount type
-    pub r#type: MountType,
-    /// Mount flags,
-    pub flags: Option<HashSet<MountFlag>>,
+#[derive(Clone, Eq, PartialEq, Debug, Serialize)]
+pub enum Mount {
+    Resource {
+        target: std::path::PathBuf,
+        name: String,
+        version: Version,
+        dir: std::path::PathBuf,
+    },
+    Bind {
+        target: std::path::PathBuf,
+        host: std::path::PathBuf,
+        flags: HashSet<MountFlag>,
+    },
+    Persist {
+        target: std::path::PathBuf,
+        flags: HashSet<MountFlag>,
+    },
 }
 
 #[derive(Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
@@ -191,9 +174,6 @@ pub struct Manifest {
     /// Environment passed to container
     #[serde(skip_serializing_if = "Option::is_none")]
     pub env: Option<HashMap<String, String>>,
-    /// List of consumed resources
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub resources: Option<Vec<Resource>>,
     /// Autostart this container upon north startup
     #[serde(skip_serializing_if = "Option::is_none")]
     pub autostart: Option<bool>,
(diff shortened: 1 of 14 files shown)


flxo:
@alfonsoros88 There are some open points in the PR. I'll mark it as draft.

north/src/manifest.rs
```diff
@@ -122,12 +122,80 @@ pub struct CGroups {
...
     pub name: Name,
     pub version: Version,
     pub dir: std::path::PathBuf,
-    pub mountpoint: std::path::PathBuf,
+}
+
+impl FromStr for Resource {
+    type Err = anyhow::Error;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let re = regex::Regex::new(r"(?P<name>\w+):(?P<version>[\d.]+)(?P<dir>[\w/]+)?")
```
flxo, line 135: Move this into a lazy static to avoid the overhead to construct and parse the regex upon every manifest parsing run.

north/src/manifest.rs
```diff
@@ -122,12 +122,80 @@ pub struct CGroups {
...
+            .context("Wrong regular expression")?;
+
+        let caps = re
+            .captures(&s)
+            .ok_or_else(|| anyhow!("Invalid resource: {}", s))?;
+
+        let name = caps
+            .name("name")
+            .map(|m| m.as_str().to_owned())
+            .unwrap_or_default();
```
flxo, line 145: Can this happen? I think not because otherwise the regex wouldn't have matched but I'm not sure.

north/src/manifest.rs
```diff
@@ -122,12 +122,80 @@ pub struct CGroups {
...
+
+        let version: Version =
+            Version::parse(caps.name("version").map(|m| m.as_str()).unwrap_or(""))?;
```
flxo, line 148: Same as above with `name`.

north/src/manifest.rs
```diff
@@ -122,12 +122,80 @@ pub struct CGroups {
...
+
+        let dir = caps
+            .name("dir")
+            .map(|m| std::path::Path::new(m.as_str()).to_owned())
+            .unwrap_or_else(|| std::path::Path::new("/").to_owned());
```
flxo, line 153: This is a again a `unreachable`.

north/src/manifest.rs
```diff
@@ -160,15 +218,22 @@ pub enum MountFlag {
 }
 
 #[derive(Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
+pub struct DirMount {
```
flxo, line 221: I'd rename this to `BindMount` to clarify what it actually is.

north/src/manifest.rs
```diff
@@ -160,15 +218,22 @@ pub enum MountFlag {
...
+    pub host: Option<std::path::PathBuf>,
```
flxo, line 222: Why is this a `Option`? Think a bind mount is not valid without the host dir.

north/src/manifest.rs
```diff
@@ -160,15 +218,22 @@ pub enum MountFlag {
...
+    pub flags: Option<HashSet<MountFlag>>,
+}
+
+#[derive(Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum MountSource {
+    Resource(Resource),
```
flxo, line 229: Think there's a third type missing: `Persist`. The container specific (optionally rw) dir we create for the app.

north/src/manifest.rs
```diff
@@ -196,8 +259,37 @@ pub struct Manifest {
...
+        {
+            let mut entries = Vec::new();
+            while let Some((target, source)) = map.next_entry()? {
+                entries.push(Mount { target, source });
+            }
+            Ok(entries)
+        }
+
+        fn expecting(&self, formatter: &mut std::fmt::Formatter) -> ::std::fmt::Result {
+            formatter.write_str("A dictionary of mounts")
```
flxo, line 288: The `expecting` normally return a string that shows a example

north/src/manifest.rs
```diff
@@ -221,6 +313,28 @@ impl Manifest {
...
+
+    /// Returns a list of tuples `(target, DirMount)`
+    pub fn mounts(&self) -> Vec<(&std::path::Path, &DirMount)> {
+        self.mounts
+            .iter()
+            .filter_map(|e| match &e.source {
+                MountSource::Directory(m) => Some((e.target.as_path(), m)),
+                _ => None,
+            })
+            .collect()
```
flxo, line 336: Changing `mounts` to be a enum this two methods are obsolete. The user of a `Manifest` then do not need to do a extra call to extract the lists.

north/src/manifest.rs
```diff
@@ -247,24 +361,22 @@ name: hello
...
           - rw
-    - target: /data
-      type: data
-resources:
-    - name: bla
-      version: 1.0.0
-      dir: /bin/foo
-      mountpoint: /here/we/go
+    /data: {}
+    /data_with_flags:
```
flxo, line 375: change this to `data_rw`

north/src/runtime/process.rs
```diff
@@ -433,64 +455,27 @@ pub mod minijail {
...
-                            );
-                        }
-                    } else {
-                        // TODO: Bind mounts without a source are invalid and should be checked in Manifest::verify
-                        return Err(anyhow!("Cannot mount of type bind without source"));
+        for (target, mount) in container.manifest.mounts() {
+            let source: PathBuf = match mount.host.as_ref() {
+                Some(dir) => {
+                    if !dir.exists() {
+                        warn!("Creating nonexistent input directory {}", dir.display(),);
```
flxo, line 462: Think this is a error. If the mounts distinguish between `Persist` and `BindMount` it is definitely an error.

north/src/runtime/process.rs
```diff
@@ -501,32 +486,29 @@ pub mod minijail {
...
-                    if src_dir.exists().await {
-                        Ok(src_dir)
-                    } else {
-                        Err(anyhow!(format!(
-                            "Resource folder {} is missing",
-                            src_dir.display()
-                        )))
-                    }
-                }?;
+        for (target, res) in container.manifest.resources() {
```
flxo, line 489: The `setup_resources` should be merged in to `setup_mounts` which matches on the mount type:
* `Bind`
* `Persist`
* `Resource`

flxo (changes requested):
Main thing would be to just have on enum for the mounts in the manifest

```rust
enum Mounts {
  Persist { target: ..., flags: ... },
  Bind { source: ..., target: ..., flags: ... }.
  Resource { name: .., version: .... flags: ... }.
}
enum Mounts {
  Persist(PersistMount),
  Bind(BindMount).
  Resource(ResourceMount).
}
```

north/src/manifest.rs
alfonsoros88: this one in particular can actually be empty

north/src/manifest.rs
alfonsoros88: Because "data" mounts or "persist" do not have a host path. I used to split bind and data.

north/src/manifest.rs
alfonsoros88: this one is just for convenience and remove the match expressions in process.rs and mock.rs

north/src/runtime/npk/linux.rs
```diff
@@ -172,7 +172,17 @@ async fn install_internal(
...
+        .iter()
+        .filter_map(|m| match m {
+            Mount::Resource { name, version, .. } => Some(format!("{} ({})", name, version)),
+            _ => None,
+        })
+        .collect();
+
+    if !resources.is_empty() {
         debug!("Referencing {} resources:", resources.len());
         for res in resources {
```
flxo, line 187: Here's no need to collect with some allocation. The two iterations can be easily merged.

north/src/manifest.rs
```diff
@@ -160,15 +136,39 @@ pub enum MountFlag {
...
-    /// Mount source
-    pub source: Option<std::path::PathBuf>,
-    /// Mount target
-    pub target: std::path::PathBuf,
-    /// Mount type
-    pub r#type: MountType,
-    /// Mount flags,
-    pub flags: Option<HashSet<MountFlag>>,
+#[serde(untagged)]
+pub enum MountSource {
```
flxo, line 140: This is just used for the deserialisation and could be declared private and within `deserialize_mounts`.

flxo (changes requested):
Cool! Move `MountSource` and that's it!

flxo:
I get a strange error when starting all app in a clean build from this pr branch:

```
>> start
Error: Failed to parse reply

Caused by:
    invalid type: sequence, expected { /path/a: Bind {} | Persist {} | Resource {}, /path/b: ... } at line 20 column 22
```
@alfonsoros88 Can you please check this?

north/src/runtime/npk/linux.rs
alfonsoros88: The problem here is that if the behavior is not to be changed, it is necessary to know if there are resources in the manifest. Unfortunately there is no way to know without iterating the mounts array. If it is fine to always print the debug string in line 186 independently of the presence of resources, then we can do this in one pass.

alfonsoros88:
> I get a strange error when starting all app in a clean build from this pr branch:
> 
> ```
> >> start
> Error: Failed to parse reply
> 
> Caused by:
>     invalid type: sequence, expected { /path/a: Bind {} | Persist {} | Resource {}, /path/b: ... } at line 20 column 22
> ```
> 
> @alfonsoros88 Can you please check this?

I think the problem was that at some point the manifest it serialize back to be sent. I didn't implement the serialization logic. I'd just add it in the last push. Hopefully it should be fine now.

PR #129, opened 2020-10-19
Remove "sudo -E" runner config for linux x86_64

### LINKED ISSUE #126: Do not run everything on x86_64-unknown-linux-gnu with `sudo`
The runner configured globally for `x86_64-unknown-linux-gnu` in the cargo config should only be applied where necessary which is the `north` package and tests. There's no need to run `nstar` or others with `sudo.

alfonsoros88:
fixes #126

### DIFF
--- a/.cargo/linux_runner_wrapper.sh
+++ b/.cargo/linux_runner_wrapper.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+binary_name=`basename $1`
+
+if [[ $binary_name =~ ^(north|tests-[a-z0-9]{16})$ ]]; then
+    sudo -E $@
+else
+    eval $@
+fi
--- a/.cargo/config
+++ b/.cargo/config
@@ -5,4 +5,4 @@ linker = "arm-linux-androideabi-g++"
 linker = "aarch64-linux-android-clang++"
 
 [target.x86_64-unknown-linux-gnu]
-runner = 'sudo -E'
+runner = '.cargo/linux_runner_wrapper.sh'


flxo:
#126 proposes to find a solution to just run north as root and not everything in the workspace. Isn't there a solution for this? Using the runner feature is quite convenient, since a build invoked via `cargo run` doesn't run as root and leaves nasty root owned files.

alfonsoros88:
> #126 proposes to find a solution to just run north as root and not everything in the workspace. Isn't there a solution for this? Using the runner feature is quite convenient, since a build invoked via `cargo run` doesn't run as root and leaves nasty root owned files.

I haven't found any nicer way to do this. Currently i'm just using the env variable:
export CARGO_TARGET_X86_64_UNKNOWN_LINUX_GNU_RUNNER="sudo -E"

flxo:
The runner just needs to check the crate name for `north`. Not sure what this is for the tests.

.cargo/linux_runner_wrapper.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+case $1 in
+    *north) sudo -E $@
+    ;;
+    *tests*) sudo -E $@
```
flxo, line 6: Can we limit this to the tests of `north`? The tests of everything else do not need `root`.

flxo (changes requested):

.cargo/linux_runner_wrapper.sh
alfonsoros88: This is running only `north` and the `tests` (the integration test one) binaries with `sudo`. No other test matches with `*tests*`. I guess, I could try to narrow the match a bit further.

flxo:
Top. Thanks!

PR #132, opened 2020-10-20
Readd wildcards to nstar

flxo:
The start and stop commands query the runtime for a container list
matching the argument as RE2. If not argument is supplied ".*" is used.

### DIFF
--- a/nstar/Cargo.toml
+++ b/nstar/Cargo.toml
@@ -15,6 +15,7 @@ itertools = "0.9.0"
 log = "0.4.11"
 north = { path = "../north", features = ["api"], default-features = false }
 prettytable-rs = "0.8.0"
+regex = "1.3.9"
 rustyline = "6.2.0"
 rustyline-derive = "0.3.1"
 serde = { version = "1.0", features = ["derive"] }
--- a/nstar/src/main.rs
+++ b/nstar/src/main.rs
@@ -18,10 +18,7 @@ use byteorder::{BigEndian, ByteOrder};
 use itertools::Itertools;
 use log::{info, warn};
 use net::TcpStream;
-use north::api::{
-    InstallationResult, Message, Payload, Request, Response, ShutdownResult, StartResult,
-    StopResult, UninstallResult,
-};
+use north::api::{Message, Payload, Request, Response};
 use prettytable::{format, Attr, Cell, Row, Table};
 use rustyline::{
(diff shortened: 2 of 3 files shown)


nstar/src/main.rs
```diff
@@ -200,32 +200,89 @@ stop <name>:    Stop application"
...
+    mut cmd: I,
+    stream: &mut S,
+) -> Result<Option<Vec<Response>>> {
+    let pattern = cmd.next().unwrap_or(".*");
+    let re = regex::Regex::new(pattern).context("Invalid regex")?;
+    if let Response::Containers(containers) = run(stream, Request::Containers)? {
+        let mut responses = vec![];
+        for container in containers
+            .iter()
+            .filter(|ref c| c.process.is_some()) // Filter not running containers
```
alfonsoros88, line 266: shouldn't the comment be: // Filter running containers ?

nstar/Cargo.toml
```diff
@@ -15,6 +15,7 @@ itertools = "0.9.0"
 log = "0.4.11"
 north = { path = "../north", features = ["api"], default-features = false }
 prettytable-rs = "0.8.0"
+regex = "1.4.1"
```
alfonsoros88, line 18: other packages are using "1.3.9", should we use the same here?

nstar/src/main.rs
```diff
@@ -200,32 +200,89 @@ stop <name>:    Stop application"
         .into()
 }
 
-fn run_cmd<S: Read + Write>(cmd: &str, stream: S) -> Result<Option<Response>> {
+fn run_cmd<S: Read + Write>(cmd: &str, stream: &mut S) -> Result<Option<Vec<Response>>> {
```
alfonsoros88, line 203: Is the optional really needed in the return type? does nstar make a distinction between a None and Some(vec![])?

alfonsoros88 (approved):

nstar/Cargo.toml
flxo: Technically this is not a problem but the double compile (in case building the rt for the host) is not nice. I'll update the other.

nstar/src/main.rs
flxo: The None is originally used to report that nothing was done. Probably this is more an error because the input didn't lead to any action. Will rethink this.

PR #138, opened 2020-10-22
Simplify and enhance example build script

flxo:
The north configuration is adjusted to use the key from the example
folder - it must not be copied there. Extend the script to take a
optional arugment for building just of a single target. If the
arugment is not passed the default set of targets is build.


flxo:
I'll close this because it's superseded by #149

PR #139, opened 2020-10-22
Fix hanging stop calls in integration tests

alfonsoros88:
The integration tests spawn `nstar` to stop containers. If the containers
were not running, `nstar` would not issue any stop command. The tests were
hanging waiting for the message of "Stopped [container name]" in the
runtime log.

This change removes that assertion.


examples/build_examples.sh
```diff
@@ -18,6 +18,7 @@ EXAMPLES=(
   "${EXAMPLES_DIR}/resource/ferris"
   "${EXAMPLES_DIR}/resource/ferris_says_hello"
   "${EXAMPLES_DIR}/resource/hello_message"
+  "./tests/test_container"
```
flxo, line 21: The test container should not be built from the examples script. Use a `build.rs` in the test crate and add `sextant` to the `dev-dependencies` of `test`. Think the `build.rs` should just package. Invoking a `cargo build` from a `build.rs` feels odd.

nstar/src/main.rs
```diff
@@ -292,19 +292,28 @@ fn stop<'a, S: Read + Write, I: Iterator<Item = &'a str>>(
...
-                container.manifest.name, container.manifest.version, response
-            );
+            .map(|ref c| c.manifest.name.clone())
+            .collect();
+
+        if running_container_names.is_empty() {
+            Err(anyhow!(
+                "No running container match the pattern: {}",
+                pattern
+            ))
```
flxo, line 306: Not sure if this is really an error. If there's nothing to do - there's nothing to do.

flxo (changes requested):

examples/build_examples.sh
alfonsoros88: I think a `build.rs` is not the way to go here as the packaging has to occur _after_ the build is done. I will add a bash script similar to the one of `build_examples.sh`.

flxo: ok

tests/src/runtime.rs
```diff
@@ -148,23 +148,10 @@ impl Runtime {
...
-                ))
-                .await
-                .context(format!("Failed to wait for {} to stop", container_name))?;
-
-            Ok(())
-        }
-        .or_timeout(TIMEOUT)
-        .await
-        .context(format!("Failed to stop container {}", container_name))?
+        async { nstar(&format!("stop {}", container_name)).await }
```
flxo, line 151: I'm confused. Isn't `nstar` a async `fn`? Why the additional `async` block around?

alfonsoros88: you are right, this is wrong.

PR #142, opened 2020-10-23
Remove dependency on ruby tooling

marcmo:
* exclude examples and tools from workspace
* repair tasks to build npks


flxo (commented):
In e1e36cd almost everything from the original commit is missing - the message doesn't match anymore at all. Think there went something wrong.

flxo:
I'll close this because it's superseded by #149

PR #143, opened 2020-10-23
Forward minijail log to runtime through pipe

### LINKED ISSUE #116: Minijail logs should be handled and printed by north
Minijail dumps log information to stdout. In order to get a uniform log the logs shall be captured/received by the north runtime and forwarded to the configured logger.

Minijail allows to hand over a fd that is written by minijail for logs but looks like this is missing in the Rust wrapper:

https://github.com/esrlabs/northstar/blob/eee59025034f60142d6df904865e1e9f150c6b25/minijail/util.c#L73-L84

Clarify what minijail does here and evaluate how intrusive setting the logging fd would be.

alfonsoros88:
closes #116

### DIFF
--- a/north/src/runtime/linux/mod.rs
+++ b/north/src/runtime/linux/mod.rs
@@ -15,7 +15,7 @@
 use super::config::Config;
 use anyhow::{Context, Result};
 use async_std::path::PathBuf;
-use log::debug;
+use log::{debug, log, Level};
 use nix::sched;
 
 pub(super) mod cgroups;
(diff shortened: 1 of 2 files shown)


north/src/runtime/linux/mod.rs
```diff
@@ -43,5 +43,52 @@ pub async fn init(config: &Config) -> Result<()> {
...
 
+    // Pipe minijail log to the runtime
+    capture_minijail_log().await?;
+
+    Ok(())
+}
+
+async fn capture_minijail_log() -> Result<()> {
+    use async_std::{os::unix::io::FromRawFd, stream::StreamExt};
+    use futures::io::AsyncBufReadExt;
```
flxo, line 54: We should avoid using `futures`. This crate will be removed with the tokio migration because it's reported by `tokio` and `async-std`. No idea why we added it back then. Think there is a `AsyncBufReadExt` in `async-std` also.

north/src/runtime/linux/mod.rs
```diff
@@ -43,5 +43,52 @@ pub async fn init(config: &Config) -> Result<()> {
...
+    if let Some(log_level) = log::max_level().to_level() {
+        let minijail_log_level = match log_level {
+            Level::Error => SyslogLevel::LOG_ERR,
+            Level::Warn => SyslogLevel::LOG_WARNING,
+            Level::Info => SyslogLevel::LOG_INFO,
+            Level::Debug => SyslogLevel::LOG_DEBUG,
+            Level::Trace => SyslogLevel::MAX,
+        };
+
+        let (readfd, writefd) = nix::unistd::pipe()?;
```
flxo, line 80: How it the pipe remove upon process exit?

north/src/runtime/linux/mod.rs
```diff
@@ -43,5 +43,52 @@ pub async fn init(config: &Config) -> Result<()> {
...
+        minijail::Minijail::log_to_fd(writefd, minijail_log_level as i32);
+
+        async_std::task::spawn(async move {
+            let pipe = unsafe { async_std::fs::File::from_raw_fd(readfd) };
```
flxo, line 84: This can be done outside of the block to handle errors.

flxo (changes requested):

north/src/runtime/linux/mod.rs
alfonsoros88: `futures` is a dependency of `async-std`. In this case, it seems it is what `async-std` is using to get the iterator from `lines()`. I didn't want to put `use async_std::prelude::*`.

north/src/runtime/linux/mod.rs
alfonsoros88: I don't seem to find any specifics, it would probably be cleaned up without much problem. What seems to be important is that if the reading part is closed, the writing part will emit a `SIGPIPE` that could kill the writer, but in this case we are not closing either.

north/src/runtime/linux/mod.rs
flxo: Why not? We're using it so the use is fine. Instead of the wildcard we could also explicitly import what we need -> `BufReadExt`

north/src/runtime/linux/mod.rs
flxo: Are pipes removed when the process exists that created the pipe? Wouldn't make much sense.

alfonsoros88: well, yes, they live in memory.

north/src/runtime/linux/mod.rs
alfonsoros88: that's what I did, i just found that it was a reexport of `futures::io::AsyncBufReadExt`. In the latest change this is replaced with `async_std::io::prelude::BufReadExt`.

north/src/runtime/linux/mod.rs
alfonsoros88: Yes, linux will close any file descriptor created by the exiting process and that should include this. This should be fine.

PR #147, opened 2020-10-27
Tag minijail log strings with log level

alfonsoros88:
- Prepends minijail's log strings with a log level so that it can be log accordingly from the north side.
- Adds a `minijail_log_handle` to `Runtime` so that the spawned task that reads minijail's log is terminated on drop.

### DIFF
--- a/minijail/util.c
+++ b/minijail/util.c
@@ -102,6 +102,18 @@ static void attribute_no_optimize alias(const void *var)
 	(void)var;
 }
 
+void print_log_level(int priority, int fd) {
+	if (priority == LOG_INFO) {
+		dprintf(fd, "INFO ");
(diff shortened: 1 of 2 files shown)


minijail/util.c
```diff
@@ -102,6 +102,18 @@ static void attribute_no_optimize alias(const void *var)
...
+		dprintf(fd, "INFO ");
+	} else if (priority == LOG_WARNING) {
+		dprintf(fd, "WARN ");
+	} else if (priority == LOG_ERR) {
+		dprintf(fd, "ERR ");
+	} else {
+		dprintf(fd, "DEBUG ");
+	}
+}
+
```
marcmo, line 116: In general looks good to me...we just need to be really careful to not touch the minijail code in a way that prevents us from updating it.

PR #150, opened 2020-10-27
Add sextant 'inspect' command

inorick:
The 'inspect' command will print
 * the list of files in an NPK
 * the manifest (manifest.yaml) and signature (signature.yaml) contents
 * the contents of the squashfs image (fs.img)

Thia PR depends on #148.

PR #151, opened 2020-10-27
Add sextant 'gen-key' command

inorick:
The 'gen-key' command will create a 32 byte key seed that can be used to sign an verify NPKs.

Thia PR depends on #148 and #150.


marcmo:
merged manually

PR #155, opened 2020-10-28
Disable memory swap in CGroupMem if available

### LINKED ISSUE #135: Example memeater behaves non-deterministically on x86_64-unknown-gnu-linux
OOM is detected at random times
image
image

alfonsoros88:
fixes #135

Setting the `swappiness` of `memeater` to 0 disables memory swap. If memory swapping is enabled, `memeater` avoids being killed by OOM despite _consuming_ more memory than it is supposed to.

### DIFF
--- a/examples/container/memeater/src/main.rs
+++ b/examples/container/memeater/src/main.rs
@@ -12,11 +12,13 @@
 //   See the License for the specific language governing permissions and
 //   limitations under the License.
 
+use std::fs;
+
 #[allow(clippy::all)]
 fn main() {
     let mut mem = vec![];
     for _ in 0..9_999_999 {
-        println!("Eating a Megabyte... have {}", mem.len());
+        println!("I've consumed {} bytes", get_used_memory_in_bytes());
         let mut chunk = vec![];
         for i in 0..1_000_000 {
             chunk.push((i % 8) as u8);
@@ -29,4 +31,25 @@ fn main() {
     for x in &mem {
         println!("{}", x[0]);
     }
+}
+
+/// Returns the amount in bytes corresponding to all the memory pages maped by the process.
+fn get_used_memory_in_bytes() -> u64 {
+    let stat = fs::read_to_string("/proc/self/stat").expect("Could not read /proc/self/stat");
+    let mut fields = stat.split_whitespace();
+
+    // Read the Resident Set Size (RSS).
+    // This number correspond to the total number of memory pages used by the process.
+    // This includes also the shared libraries which means that the final count is larger
+    // than the real amount of bytes consumed by the process.
+    let rss = fields
+        .nth(23)
+        .expect("Could not read RSS memory usage")
+        .parse::<u64>()
+        .expect("Could not parse RSS");
+
+    // Usually the page size is set to 4kB by default
+    let page_size = 4096;
+
+    rss * page_size
 }
--- a/north/src/runtime/linux/cgroups.rs
+++ b/north/src/runtime/linux/cgroups.rs
@@ -28,6 +28,7 @@ use proc_mounts::MountIter;
 use std::time::Duration;
 use thiserror::Error;
 
+const SWAPPINESS: &str = "memory.swappiness";
 const LIMIT_IN_BYTES: &str = "memory.limit_in_bytes";
 const OOM_CONTROL: &str = "memory.oom_control";
 const UNDER_OOM: &str = "under_oom 1";
@@ -129,6 +130,8 @@ impl CGroupMem {
         );
         write(&limit_in_bytes, &cgroup.limit.to_string()).await?;
 
+        try_disable_memory_swap(&path).await?;
+
         // Configure oom
         let oom_control = path.join(OOM_CONTROL);
         write(&oom_control, "1").await?;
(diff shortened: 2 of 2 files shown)


north/src/runtime/linux/cgroups.rs
```diff
@@ -100,6 +101,14 @@ impl CGroupMem {
                 format!("Failed to set cgroup limit in {}", limit_in_bytes.display())
             })?;
 
+        // Try to disable memory swap
+        // This might be enabled for linux desktops, making the memeater example avoid being killed
+        // when exceeding the memory limit.
+        let swappiness = path.join(SWAPPINESS);
+        if write(&swappiness, "0").await.is_err() {
```
flxo, line 108: Would be better to check for the file existence before trying to write.

examples/container/memeater/Cargo.toml
```diff
@@ -4,4 +4,5 @@ version = "0.1.0"
 authors = ["ESRLabs"]
 edition = "2018"
 
-[dependencies]
\ No newline at end of file
+[dependencies]
+nix = "0.17.0"
```
flxo, line 8: In order to keep the examples small, tiny and examplish we shouldn't add any deps here. Adding `nix` just for the page size is probably a little bit too much.

north/src/runtime/linux/cgroups.rs
```diff
@@ -100,6 +101,14 @@ impl CGroupMem {
                 format!("Failed to set cgroup limit in {}", limit_in_bytes.display())
             })?;
 
+        // Try to disable memory swap
```
flxo, line 104: Good catch!

north/src/runtime/linux/cgroups.rs
```diff
@@ -103,6 +104,11 @@ impl CGroupMem {
         );
         write(&limit_in_bytes, &cgroup.limit.to_string()).await?;
 
+        // Try to disable memory swap
+        // This might be enabled for linux desktops, making the memeater example avoid being killed
+        // when exceeding the memory limit.
```
marcmo, line 109: Better not to refer to a specific example in the comment

north/src/runtime/linux/cgroups.rs
```diff
@@ -144,6 +150,17 @@ impl CGroupMem {
     }
 }
 
+async fn try_disable_memory_swap(path: &Path) -> Result<(), CGroupError> {
+    let swappiness = path.join(SWAPPINESS);
+    if swappiness.exists().await {
+        debug!("Disabling memory swap");
+        write(&swappiness, "0").await
+    } else {
+        debug!("Memory swap is not enabled");
```
marcmo, line 159: In this case, we might not need to log anything

examples/container/memeater/src/main.rs
```diff
@@ -30,3 +32,24 @@ fn main() {
...
+    // This includes also the shared libraries which means that the final count is larger
+    // than the real amount of bytes consumed by the process.
+    let rss = fields
+        .nth(23)
+        .expect("Could not read RSS memory usage")
+        .parse::<u64>()
+        .expect("Could not parse RSS");
+
+    // Usually the page size is set to 4kB by default
+    let page_size = 4096;
```
marcmo, line 52: Can we use `getpagesize()` for that?

alfonsoros88: that would introduce a dependency to either libc or nix, which @flxo recommended to avoid to keep this example simple.

flxo:
Isn't this superseded by #167?

alfonsoros88:
> Isn't this superseded by #167?

Depend's if you want to explicitly set `swappiness` in the manifests, or implicitly disable it for every container. I think explicit is better than implicit but you decide.

flxo:
> > Isn't this superseded by #167?
> 
> Depend's if you want to explicitly set `swappiness` in the manifests, or implicitly disable it for every container. I think explicit is better than implicit but you decide.

I think it's a matter of documentation. Disabling `swapiness` is something one should only do if they know what they're doing.

alfonsoros88:
> > > Isn't this superseded by #167?
> > 
> > 
> > Depend's if you want to explicitly set `swappiness` in the manifests, or implicitly disable it for every container. I think explicit is better than implicit but you decide.
> 
> I think it's a matter of documentation. Disabling `swapiness` is something one should only do if they know what they're doing.

Does that mean that we will add `swappiness: 0` to the memeater example after #167  is done?, or we leave it the way it is and close this pr? What would happen with #135  if the latter?

examples/container/memeater/src/main.rs
marcmo: you are right...skip it

flxo:
> > > > Isn't this superseded by #167?
> > > 
> > > 
> > > Depend's if you want to explicitly set `swappiness` in the manifests, or implicitly disable it for every container. I think explicit is better than implicit but you decide.
> > 
> > 
> > I think it's a matter of documentation. Disabling `swapiness` is something one should only do if they know what they're doing.
> 
> Does that mean that we will add `swappiness: 0` to the memeater example after #167 is done?, or we leave it the way it is and close this pr? What would happen with #135 if the latter?

I'd do exactly that: add `snappiness: 0` in a manifest for `x86_64-unknown-gnu-linux`. Swap is almost never enabled for embedded targets so we probably need to differentiate here.
I just though that this swappiness setting can be easily done when #167 is finalised (still a draft)

PR #156, opened 2020-10-31
Add Tmpfs mount type

### LINKED ISSUE #146: Mount of a `tmpfs` should be optional and configurable
Add another variant to `manifest::Mount` called `Tempfs` with options `target` and `size`. Not sure if the minijail api already allows to set the size of the `tmpfs` created with `mount_tmp` but this is probably easy to extend. A `tmpfs` is always `rw`.

alfonsoros88:
closes #146

### DIFF
--- a/north/src/manifest.rs
+++ b/north/src/manifest.rs
@@ -136,7 +136,7 @@ pub enum MountFlag {
     // NoSuid,
 }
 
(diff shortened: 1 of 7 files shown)


flxo:
Looks good to me. I'd merge it after @marcmo notification path.

PR #158, opened 2020-11-02
Remove redundant Error and Problem terms from Errors

flxo:


north/src/runtime/error.rs
```diff
@@ -116,8 +116,8 @@ pub enum KeyError {
 }
 
 #[derive(Error, Debug)]
-pub enum InstallFailure {
-    #[error("File seems to be corrupted")]
+pub enum Installation {
```
marcmo, line 119: I generally agree that elements in an enum that's called XYZError do not need the Error, but a top level enum? I don't think that `Installation` or `Key` is the right term to describe an `InstallationError` or `KeyProblem`

flxo: I think the additional `Error` can be omitted because they all are in the mod `error`. This pattern is also followed in the stdlib. e.g here: https://doc.rust-lang.org/std/io/enum.ErrorKind.html

I would also strongly suggest to stick to one term: Error or Problem.

marcmo: In ErrorKind the error types are self-descriptive, that's fine.
image
Calling an enum `Key` does not reflect in any way that it actually is an error...even if it currently resides in the `error` module and actually is misleading.

marcmo:
build is still not green

flxo:
I moved the error definitions to the submodules and refactored some things. Still a draft.

north/src/runtime/process/minijail.rs
```diff
@@ -0,0 +1,416 @@
...
+use log::{debug, warn};
+use nix::{
+    sys::signal,
+    unistd::{self, chown},
+};
+use std::{fmt, os::unix::io::AsRawFd, time};
+use stop_token::StopSource;
+
+// We need a Send + Sync version of Minijail
+struct Minijail(::minijail::Minijail);
```
marcmo, line 32: Much nicer to have this in separate modules!

PR #159, opened 2020-11-03
Rework Sextant verity logic

### LINKED ISSUE #55: Ruby npk tooling is broken for fs images < 4k (block size)
The current ruby hack for creating the verity trees doesn't work for fs images sizes < block size (4k default).
To reproduce take a resource container and remove `root*` - make it empty. The fsimage only contains pseudo files and has exactly 4k block size.

@inorick You have to takle this anyway in your port - so I assigned it to you.

inorick:
This change simplifies and documents the dm-verity logic and moves it into its own crate.
Fixes #55.

### DIFF
--- a/north/src/runtime/console.rs
+++ b/north/src/runtime/console.rs
@@ -195,11 +195,11 @@ impl Console {
     }
 
     /// Send a notification to the notification broadcast
-    pub async fn notification(&self, notification: api::Notification) -> Result<(), Error> {
-        self.notification_tx
-            .send(notification)
-            .map_err(|_| Error::Internal("Notification channel tx"))
-            .map(drop)
+    pub async fn notification(&self, notification: api::Notification) {
+        debug!("sending notification: {:?}", notification);
+        if self.notification_tx.send(notification).is_err() {
+            debug!("No subscribers received the notification");
+        }
     }
 
     async fn connection(
--- a/north/src/runtime/mod.rs
+++ b/north/src/runtime/mod.rs
@@ -154,7 +154,7 @@ pub async fn run(config: &Config) -> Result<(), Error> {
             // The runtime os commanded to shut down and exit.
             Event::Shutdown => break,
             // Forward notifications to console
-            Event::Notification(notification) => console.notification(notification).await?,
+            Event::Notification(notification) => console.notification(notification).await,
             // Handle unrecoverable errors by logging it and do a gracefull shutdown.
             Event::Error(ref error) => {
                 error!("Fatal error: {}", error);


sextant/src/npk.rs
```diff
@@ -286,72 +286,86 @@ fn gen_pseudo_files(manifest: &Manifest) -> Result<Vec<(&str, u32)>> {
...
-            }
-            level_output_list.push(sha256.finalize().into());
+            rem_size -= BLOCK_SIZE as u64;
+            level_hashes.push(sha256.finalize().into());
+        }
+
+        // "7. Repeat steps 2-6 using the previous level as the source for the next until you have only a single hash."
+        if level_hashes.len() == 1 {
+            break;
         }
```
marcmo: Why is step 7 before step 4?

sextant/src/npk.rs
```diff
@@ -286,72 +286,86 @@ fn gen_pseudo_files(manifest: &Manifest) -> Result<Vec<(&str, u32)>> {
...
+    if image_size % BLOCK_SIZE as u64 != 0 {
+        return Err(anyhow!(
+            "Failed to generate verity has tree. The image size {} is not a multiple of the block size {}",
+            image_size,
+            BLOCK_SIZE
+        ));
+    }
+
+    loop {
+        level_hashes.clear();
```
marcmo: The loop logic is not clear to me...I mean the exit condition. Breaking from a loop is always harder to read IMHO

sextant/src/npk.rs
```diff
@@ -512,30 +523,30 @@ fn write_verity_header(
...
+    fsimg.write_all(&4096_u32.to_ne_bytes())?; // data block in bytes
+    fsimg.write_all(&4096_u32.to_ne_bytes())?; // hash block in bytes
+    fsimg.write_all(&data_blocks.to_ne_bytes())?; // number of data blocks
+    fsimg.write_all(&32_u16.to_ne_bytes())?; // salt size (SHA256 output length)
+    fsimg.write_all(&[0_u8; 6])?; // padding
     fsimg.write_all(&salt.to_vec())?;
-    fsimg.write_all(&vec![0_u8; 256 - salt.len()])?;
-    fsimg.write_all(&[0_u8; 3752])?;
+    fsimg.write_all(&vec![0_u8; 256 - salt.len()])?; // padding
+    const BYTES_WRITTEN: usize = 8 + 4 + 4 + 16 + 6 + 26 + 4 + 4 + 8 + 2 + 6 + 256;
```
marcmo: maybe we can use `write_vectored` here? (https://doc.rust-lang.org/std/io/trait.Write.html#method.write_vectored)
image
this could give us the bytes written directly

marcmo:
Good idea to split it up!

sextant/src/npk.rs
inorick: This comes straigt from the algorithm description [1] in which steps 2-6 describe a loop and step 7 describes the condition to break out of it. I'll remove the leading '7.' to avoid the confusion.

[1] https://source.android.com/security/verifiedboot/dm-verity#hash-tree

sextant/src/npk.rs
inorick: This attempts to stay close to the original algorithm description (see above). I'll see if this can be simplified.

sextant/src/npk.rs
inorick: I ended up using a vector as a superblock buffer and extending it by each field individually. That way, we have access to the number of bytes to be written. Using IoSlice did not work as we are writing a lot of temporaries.

PR #163, opened 2020-11-04
Refactor keys.rs

alfonsoros88:


north/src/runtime/keys.rs
```diff
@@ -36,40 +32,40 @@ pub enum Error {
...
+    fn context(self, context: String) -> Result<T, Error> {
+        self.map_err(|error| Error::Io { context, error })
+    }
+}
+
+impl<T> MapError<T> for Result<T, ed25519_dalek::SignatureError> {
+    fn context(self, context: String) -> Result<T, Error> {
+        self.map_err(|_| Error::Signature(context))
+    }
+}
```
flxo, line 71: Not sure if we should declare such a general util thing in here. Why not declaring a error variant that uses `#from(io::Error)` and a "context" string e.g

```diff
diff --git a/north/src/runtime/keys.rs b/north/src/runtime/keys.rs
index a8b1378..1f61607 100644
--- a/north/src/runtime/keys.rs
+++ b/north/src/runtime/keys.rs
@@ -20,12 +20,8 @@ use tokio::{fs, io, stream::StreamExt};
 
 #[derive(Error, Debug)]
 pub enum Error {
-    #[error("Io error: {context}")]
-    Io {
-        context: String,
-        #[source]
-        error: io::Error,
-    },
+    #[error("IO error: {0}")]
+    Io(&'static str, #[source] io::Error),
     #[error("Invalid key signature: {0}")]
     Signature(String),
 }
@@ -34,9 +30,9 @@ pub async fn load(key_dir: &Path) -> Result<HashMap<String, PublicKey>, Error> {
     let mut signing_keys = HashMap::new();
     let mut key_dir = fs::read_dir(&key_dir)
         .await
-        .context(format!("Failed to open {}", key_dir.display()))?;
+        .map_err(|e| Error::Io("Failed to read key dir", e))?;
     while let Some(entry) = key_dir.next().await {
-        let path = entry.context("Invalid key dir entry".to_string())?.path();
+        let path = entry.map_err(|e| Error::Io("Failed to read key dir entry", e))?.path();
         if path.extension().filter(|ext| *ext == "pub").is_none() || !path.is_file() {
             continue;
         }
@@ -45,27 +41,11 @@ pub async fn load(key_dir: &Path) -> Result<HashMap<String, PublicKey>, Error> {
             info!("Loading signing key {}", key_id);
             let key_bytes = fs::read(&path)
                 .await
-                .context(format!("Failed to read {}", path.display()))?;
+                .map_err(|e| Error::Io("Failed to read key", e))?;
             let key = PublicKey::from_bytes(&key_bytes)
-                .context(format!("Signature error for key from {}", path.display()))?;
+                .map_err(|e| Error::Io("Failed to read pub key", e))?;
             signing_keys.insert(key_id, key);
         }
     }
```
This is less code with the same amount of information and feels more natural.

alfonsoros88: you cheated a bit ;) you dropped the `format!`s with the context info, that would bump up the code a bit. I will change it as you show but we should look for some way to simplify all the `map_err`s everywhere, they are a bit distracting.

PR #167, opened 2020-11-06
Generic cgroup configuration

### LINKED ISSUE #135: Example memeater behaves non-deterministically on x86_64-unknown-gnu-linux

alfonsoros88:
fixes #135

- Turn manifest cgroup section into a generic HashMap
- Join CGroupMem and CGroupCpu
- Replace config::CGroups with a HashMap
- Add init function to configure root cgroups
- Disable memory swappiness in memeater example

### DIFF
--- a/npk/src/manifest.rs
+++ b/npk/src/manifest.rs
@@ -102,26 +102,6 @@ pub enum OnExit {
     Restart(u32),
 }
 
-#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
-pub struct CGroupMem {
-    /// Limit im bytes
-    pub limit: u64,
-}
-
-#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
-pub struct CGroupCpu {
-    /// CPU shares assigned to this container. CGroups divide the resource CPU into 1024 shares.
-    pub shares: u32,
-}
-
-#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
-pub struct CGroups {
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub mem: Option<CGroupMem>,
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub cpu: Option<CGroupCpu>,
-}
-
 #[derive(Clone, Eq, PartialEq, Debug, Hash, Serialize, Deserialize)]
 pub enum MountFlag {
     /// Bind mount
@@ -164,6 +144,9 @@ pub enum Mount {
     /// Mount a tmpfs with size
     Tmpfs { size: u64 },
 }
+
+pub type CGroup = HashMap<String, String>;
+pub type CGroups = HashMap<String, CGroup>;
 
 #[derive(Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
 pub struct Manifest {
@@ -506,8 +489,8 @@ mounts:
     resource: bla-blah.foo:1.0.0/bin/foo
 autostart: true
 cgroups:
-  mem:
-    limit: 30
+  memory:
+    limit_in_bytes: 30
   cpu:
     shares: 100
 seccomp:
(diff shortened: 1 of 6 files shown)


north/src/runtime/linux/cgroups.rs
```diff
@@ -23,181 +23,93 @@ use std::{
...
-        // Configure oom
-        let oom_control = path.join(OOM_CONTROL);
-        write(&oom_control, "1").await?;
-
-        // Dropping monitor will stop the task below
-        let (monitor, mut rx) = mpsc::channel::<()>(1);
-        let name = name.to_string();
+pub async fn init_root_cgroups(config: &config::CGroups) -> Result<(), LinuxError> {
+    let mount_point = get_mount_point("cpuset")?;
+    let cpuset_root = mount_point.join(get_cgroup_root("cpuset", config));
```
marcmo, line 53: CGroups are now just a hashmap...so they are very generic. But you still treat it in a very non-generic way.
If they can indeed be generic, there should not be any code specific to a concrete cgroup anymore. If we need that anyway it should also be reflected in the datatype.

north/src/runtime/linux/cgroups.rs
```diff
@@ -23,181 +23,93 @@ use std::{
...
+            .map_err(|e| Error::Destroy(self.path.to_string_lossy().to_string(), e).into())
     }
 }
 
 #[derive(Debug)]
-pub struct CGroups {
-    pub mem: Option<CGroupMem>,
-    pub cpu: Option<CGroupCpu>,
-}
+pub struct CGroups(Vec<CGroup>);
```
marcmo, line 112: In `config.rs` the type `CGroups` is a hashmap, here it is a vector...if we need different types they should have different names. But I guess we do not need them.

npk/src/manifest.rs
```diff
@@ -167,6 +146,9 @@ pub enum Mount {
     Tmpfs { size: u64 },
 }
 
+pub type CGroup = HashMap<String, String>;
+pub type CGroups = HashMap<String, CGroup>;
+
```
marcmo, line 151: A 3. `CGroups` type...we should not have 3 different types for the cgroups-concept

north/src/runtime/linux/cgroups.rs
```diff
@@ -206,83 +118,101 @@ impl CGroups {
...
+
+async fn create_if_not_exists(path: &Path) -> Result<(), LinuxError> {
     if !path.exists() {
+        debug!("Creating {}", path.display());
         fs::create_dir_all(&path).await.map_err(|e| {
-            LinuxError::FileOperation(format!("Failed to create {}", path.display()), e)
-        })?;
+            LinuxError::FileOperation(format!("Failed to create directory {}", path.display()), e)
+        })
+    } else {
```
marcmo, line 191: rather:
```
    fs::create_dir_all(&path).await.map_err(|e| {
        LinuxError::FileOperation(format!("Failed to create directory {}", path.display()), e)
    })?;
}
Ok(())
}
```

north/src/runtime/mod.rs
```diff
@@ -94,9 +94,14 @@ impl Runtime {
         let (stop_tx, stop_rx) = oneshot::channel();
         let (stopped_tx, stopped_rx) = oneshot::channel();
 
-        // Initialize minijails static functionality
         #[cfg(any(target_os = "android", target_os = "linux"))]
-        linux::minijail::init().await.map_err(Error::Linux)?;
+        {
```
marcmo, line 98: The `#[cfg...` should not be needed anymore

npk/src/manifest.rs
```diff
@@ -502,8 +484,8 @@ mounts:
     resource: bla-blah.foo:1.0.0/bin/foo
 autostart: true
 cgroups:
-  mem:
-    limit: 30
+  memory:
+    limit_in_bytes: 30
```
marcmo, line 488: nice...better to understand

north/src/runtime/linux/cgroups.rs
alfonsoros88: I agree with you. The `cpuset` controller requires a "especial" treatment and that's why this function got introduced as a consequence. A possible solution to get rid of this would be to enable the possibility to configure the parent `cgroup` directories too, e.g. adding a `(param, value)` to the `cgroups` section in the `north.toml` configuration that are applied to the complete hierarchy. Does that sound reasonable?

north/src/runtime/linux/cgroups.rs
alfonsoros88: There is a `cgroup` section in the `manifest.yml`, one in the `north.toml` config and finally the source `cgroups.rs`. Therefore the 3 definitions, related but not equivalent. I find this confusing too. I think we can't unify them all, as they represent slightly different things, maybe we can resort to just give them different names?

marcmo: A unified type would be nice if possible, but if not practical different types might be an option!

north/src/runtime/linux/cgroups.rs
marcmo: Sounds reasonable if we decide to expose the full cgroups configuration to the user.
I'm still not sure if that is a good thing though since it requires the user to fully understand cgroups to use north.
Might be a no-brainer for those pinguins but me for example I did not configure any cgroup before Northstar ;)
maybe we can discuss this here #234

PR #175, opened 2020-11-10
Implement sextant unpack command

inorick:
This PR implements the sextant unpack command which will unzip a given NPK and unsquash the contained fs.img.

This PR is based on previous work done in #159.


marcmo:
looks good ;)
once the checks are green and everything is properly rebased, we can merge it.

PR #176, opened 2020-11-11
Refactoring some (linux) errors. Sync tests to head. 

flxo:
This PR is a follow up to the tokio migration.

### DIFF
--- a/north/src/api.rs
+++ b/north/src/api.rs
@@ -15,7 +15,6 @@ pub struct Message {
 pub enum Payload {
     Request(Request),
     Response(Response),
-    Installation(usize), // size of npk that will be sent
     Notification(Notification),
 }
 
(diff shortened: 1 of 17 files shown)


north/src/api.rs
```diff
@@ -39,6 +38,7 @@ pub enum Request {
     Containers,
     Start(Name),
     Stop(Name),
+    Install(usize),
```
marcmo, line 41: good idea.
but we should add a descriptive change log entry also.

nstar/src/communication.rs
```diff
@@ -49,7 +65,7 @@ pub fn start_receiving_from_socket(
 ) -> Result<sync::broadcast::Sender<Response>> {
     let (response_sender, _) = sync::broadcast::channel(100);
     let sender = response_sender.clone();
-    let _ = tokio::spawn(async move {
+    task::spawn(async move {
```
marcmo, line 68: is there a difference between `task::spawn` and `tokio::spawn`?

flxo: No. But we use the `task` module all over. The `_` binding is useless.

PR #180, opened 2020-11-12
Replace zip padding with #200 of zip.rs

flxo:
The PR #200 is likely to be merged in master of zip soon.
This patch makes use of the `support-extra-field` branch
from esrlabs/zip.

This contains also a minor improvement when inspecting a
npk which doesn't need to be unpacked.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -193,9 +193,9 @@ dependencies = [
 
 [[package]]
 name = "cc"
-version = "1.0.61"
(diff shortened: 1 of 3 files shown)


flxo:
@inorick As discussed here's the switch to the PR #200 of zip-rs.

inorick (approved):
Looks good to me. Way shorter and ready for the next version of zip-rs.

PR #182, opened 2020-11-12
Rewrite of nstar

flxo:
This rewrite of nstar adds proper reconnects and simplifies the UI.
Futhermore notificaitons are not displayed if running in command mode
(means a single command is supplied via command line).


north/src/api.rs
```diff
@@ -11,14 +26,35 @@ pub struct Message {
...
+
+#[derive(new, Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
 pub enum Payload {
     Request(Request),
     Response(Response),
     Notification(Notification),
 }
 
-#[derive(Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
+#[derive(new, Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
```
marcmo, line 57: Is that a preparation for backward compatibility?

marcmo: the `#[derive(new)]`

nstar/src/codec.rs
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+        if let Some(len) = self.len {
+            if src.remaining() >= len {
+                let message: api::Message = serde_json::from_slice(&src[..len])?;
+                src.advance(len);
+                self.len.take();
+                return Ok(Some(message));
+            }
+        }
```
marcmo, line 71: better with a match expression
``` rust
        match self.len {
            None => {
                if src.remaining() >= 4 {
                    let len = src.get_u32() as usize;
                    self.len = Some(len);
                } else {
                    return Ok(None);
                }
            }
            Some(len) => {
                if src.remaining() >= len {
                    let message: api::Message = serde_json::from_slice(&src[..len])?;
                    src.advance(len);
                    self.len.take();
                    return Ok(Some(message));
                }
            }
        }
```

nstar/src/main.rs
```diff
@@ -12,288 +12,435 @@
...
-    cmd: &str,
-    stream: &mut S,
-    response_receiver: sync::broadcast::Receiver<Response>,
-) -> Result<(), NstarError> {
-    let mut cmd = cmd.trim().split_whitespace();
-    let c = cmd
-        .next()
-        .ok_or_else(|| NstarError::OperationError("Invalid command".to_string()))?;
+        // Spawn a task that reads on the connection and forwards results to the according channel.
+        // If this tasks breaks the main loop wil get a None on notification_rx or request_tx and
```
marcmo, line 122: will

nstar/src/codec.rs
flxo: This is semantically different: If there's a new `len` from the first branch the second one has to be executed as well because fn will not be called again.

north/src/api.rs
flxo: No. It's just to generate `new` methods be cause I'm lazy ;-)

nstar/src/codec.rs
marcmo: you are right...

nstar/src/codec.rs
```diff
@@ -0,0 +1,74 @@
...
+use bytes::{Buf, BufMut, Bytes};
+use north::api::{self};
+use tokio_util::codec;
+
+pub(crate) enum Message {
+    Message(api::Message),
+    Raw(Bytes),
+}
+
+/// A tokio_util Codec for the north api protocol
```
marcmo, line 25: so good to have the tokio library providing help with this

PR #183, opened 2020-11-12
Fix runtime shutdown

flxo:
There's some confusion about the terms uninstall and umount. Turned out that uninstall logically has been a umount. emaned
install/uninstall to mount/umount.

A signal triggers a graceful shutdown performed by the state module.
A shutdown initiated remotly does the same. Errors during shutdown like not terminating apps or failing umounts are logged and ignored since mounts happen anyway because of the (MS_PRIVATE) mount namespace.

The order of the mounts should respect resource dependencies -> not true. Resouces can be unmounted if no application is running.

### DIFF
--- a/north/src/runtime/mod.rs
+++ b/north/src/runtime/mod.rs
@@ -31,7 +31,7 @@ use state::State;
 use std::path::PathBuf;
 use sync::mpsc;
 use tokio::{
-    fs,
+    fs, select,
     sync::{self, oneshot},
 };
 
@@ -49,8 +49,6 @@ pub enum Event {
     Exit(Name, ExitStatus),
     /// Out of memory event occured
     Oom(Name),
-    /// Fatal unhandleable error
-    Error(Error),
     /// North shall shut down
     Shutdown,
     /// Stdout and stderr of child processes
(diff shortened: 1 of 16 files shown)


north/src/runtime/mod.rs
```diff
@@ -121,71 +170,63 @@ pub async fn run(config: &Config) -> Result<(), Error> {
...
+    task::spawn(async move {
+        stop.await.ok();
+        shutdown_tx.send(Event::Shutdown).await.ok();
+    });
+
     // Enter main loop
-    while let Some(event) = event_rx.recv().await {
-        match event {
+    loop {
+        let result = match event_rx.recv().await.unwrap() {
```
marcmo, line 190: We should not use unwraps in the code base. If we really want to panic, we should explicitly use `panic!("bla")` or s.th. like that.

north/src/runtime/npk/linux.rs
```diff
@@ -87,57 +86,62 @@ pub async fn install(state: &mut State, npk: &Path) -> Result<(Name, Version), I
...
-    let (name, version) = install_internal(state, &dm, &lc, npk).await?;
+    let (name, version) = mount_internal(state, &dm, &lc, npk).await?;
 
     Ok((name, version))
 }
 
 #[allow(dead_code)]
-pub async fn uninstall(container: &Container) -> Result<(), InstallationError> {
-    debug!("Unmounting {}", container.root.display());
+pub async fn umount(container: &Container) -> Result<(), InstallationError> {
```
marcmo, line 95: much better!

north/src/runtime/state.rs
```diff
@@ -315,36 +314,42 @@ impl State {
...
+    /// Send a shutdown request to the main loop
+    pub async fn initiate_shutdown(&mut self) {
+        self.events_tx
+            .send(Event::Shutdown)
+            .await
+            .expect("Internal channel error on main");
+    }
+
+    /// Shutdown the runtime: stop running applications and umount npks
+    pub async fn shutdown(&mut self) -> result::Result<(), Error> {
```
marcmo, line 326: Do we still need this in the public interface since we have `initiate_shutdown`?

flxo: Think so because it's used from the console.

flxo:
Thanks!

PR #197, opened 2020-11-13
Run integration tests in check workflow

### LINKED ISSUE #188: Evaluate integrations tests in GitHub Actions
Github actions run as `root`. This would in theory allow to run the integration test there. It's possible that not all features can be used.

=> Smoke test the runtime in GH actions: start and mount the hello example

alfonsoros88:
closes #188 

- Build test container only for host
- Add integration-tests job in check workflow
- Renames `tests` crate to `north_tests`


Cargo.toml
```diff
@@ -16,6 +16,21 @@ members = [
     "examples/container/resource/ferris",
 ]
 
+default-members = [
```
flxo, line 19: This excludes `integration_test` for all other CI checks that do a `cargo check` etc...

Would it be a option to check for the existence of the `north` binary and the test container before starting a test? This is best be done with a proc_macro that wraps `#[tokio::test]`.

.cargo/linux_runner_wrapper.sh
```diff
@@ -2,7 +2,7 @@
 
 binary_name=`basename $1`
 
-if [[ $binary_name =~ ^(north|tests-[a-z0-9]{16})$ ]]; then
+if [[ $binary_name =~ ^(north|integration_tests-[a-z0-9]{16})$ ]]; then
```
flxo, line 5: Is it `integration_test` or `north_test`?

Why not just sticking to `tests`?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,39 @@
+
```
flxo, line 1: This newline is my bad. Sorry - will fix it when merged.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,39 @@
...
+            ./examples/build_examples.sh
+            ./north_tests/build_test_container.sh
+        env:
+            TERM: xterm
+      - name: Build
+        run: cargo build
+      - name: Integration Tests
+        run: cargo test -p north_tests -- --test-threads 1 --ignored --nocapture
+        env:
+            RUST_LOG: debug
```
flxo, line 39: The missing newline is also my bad...

.cargo/linux_runner_wrapper.sh
alfonsoros88: it is `integration_test`. I was looking for examples on how people implement this kind of integration tests. This seems to be the pattern that most projects use.

The reason for the rename is that `tests` has a special treatment as it is the place where integration tests are supposed to go. The content of the `tests` folder is compiled separately from the crate itself. The crate `north_tests` will contain tests utils in general (including possible tests for those features) while the `tests` folder will contain the actual integration tests that's compiled separately to all the crates, including `north_tests`.

Maybe `northstar_tests` is a better name.

Cargo.toml
alfonsoros88: I was looking for a way to remove the `#[ignore]` from the integration tests and not have them executed when running `cargo test`. One possibility is to _exclude_ the integration tests by changing `cargo test` to `cargo test --lib --bins --examples --benches` which is not that nice.

regarding the issue about checking for `north` and so on, i'm experimenting with https://github.com/crate-ci/escargot to see if i can manage to somehow build the binaries as part of running the test. No much luck so far, but maybe it is possible.

.github/workflows/tests.yml
alfonsoros88: i can remove it don't worry :)

PR #198, opened 2020-11-16
Do not terminate on empty subscribers notification

alfonsoros88:
If `nstar` is used in _command line_ mode, its connection with the
runtime will be closed by the time the a notification is sent. This
causes the runtime to terminate when trying to send to the broadcast
channel.

This change re-enables the integration tests.

### DIFF
--- a/north/src/runtime/console.rs
+++ b/north/src/runtime/console.rs
@@ -195,11 +195,11 @@ impl Console {
     }
 
     /// Send a notification to the notification broadcast
-    pub async fn notification(&self, notification: api::Notification) -> Result<(), Error> {
-        self.notification_tx
-            .send(notification)
-            .map_err(|_| Error::Internal("Notification channel tx"))
-            .map(drop)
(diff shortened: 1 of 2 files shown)


north/src/runtime/console.rs
```diff
@@ -195,11 +195,11 @@ impl Console {
...
     /// Send a notification to the notification broadcast
-    pub async fn notification(&self, notification: api::Notification) -> Result<(), Error> {
-        self.notification_tx
-            .send(notification)
-            .map_err(|_| Error::Internal("Notification channel tx"))
-            .map(drop)
+    pub async fn notification(&self, notification: api::Notification) {
+        debug!("sending notification: {:?}", notification);
+        if self.notification_tx.send(notification).is_err() {
+            warn!("No subscribers could receive notification");
```
flxo, line 201: Good catch!
I'd not do a warning, because not always a client is connected - and that's totally ok.

PR #199, opened 2020-11-17
Add test cases to sextant's npk.rs

inorick:


marcmo:
add a macro to exclude tests from mac os or wait until I removed the macos support

marcmo:
Looks good

PR #203, opened 2020-11-18
Add /dev mount options to the manifest

### LINKED ISSUE #194: Android: expose /dev/sockets to applications
Minijail exposes only a subset of `/dev` to the process. On Android at least /dev/socket/logdw` is needed to reach `logd`. This can be accomplished by stupidly bind mounting `/dev` but this is less restrictive than the MJ way.

Another option is to bind mount `/dev/socket` to `/dev/socket`. Or symlink(?) dedicated devices from `/dev` on a manifest configuration basis.

flxo:
Fixes #194

### DIFF
--- a/npk/src/manifest.rs
+++ b/npk/src/manifest.rs
@@ -202,7 +202,7 @@ pub struct Manifest {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub instances: Option<u32>,
     /// List of bind mounts and resources
-    #[serde(with = "MountsSerialization")]
+    #[serde(with = "MountsSerialization", skip_serializing_if = "HashMap::is_empty")]
     #[serde(default)]
     pub mounts: HashMap<PathBuf, Mount>,
 }
--- a/npk/src/manifest.rs
+++ b/npk/src/manifest.rs
@@ -202,7 +202,10 @@ pub struct Manifest {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub instances: Option<u32>,
     /// List of bind mounts and resources
-    #[serde(with = "MountsSerialization", skip_serializing_if = "HashMap::is_empty")]
+    #[serde(
+        with = "MountsSerialization",
+        skip_serializing_if = "HashMap::is_empty"
+    )]
     #[serde(default)]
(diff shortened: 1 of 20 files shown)


npk/src/manifest.rs
```diff
@@ -206,49 +223,68 @@ enum MountSource {
...
+                    "m" => value * 1024 * 1024,
+                    "g" => value * 1024 * 1024 * 1024,
+                    _ => {
+                        return Err(serde::de::Error::custom(format!(
+                            "Invalid tmpfs unit: {}",
+                            unit.as_str()
+                        )))
+                    }
+                };
+                Ok(value)
```
alfonsoros88, line 285: I think this breaks the serialization/deserialization symmetry as we once discussed

flxo: That's true. The problem is not solvable unless we want to have numbers in quotes (strings) in the manifest as you pointed out some weeks ago. The comparison of manifests in yaml needs to be done the numbers must be encoded in bytes. A option would be to remove the unit sugar...

npk/src/manifest.rs
```diff
@@ -523,52 +566,135 @@ log:
...
-    /data: {}
-    /data_rw:
-      flags:
-          - rw
-    /here/we/go:
-        resource: bla:1.0.0/bin/foo
-    /tmpfs:
-        size: 42
-    /big_tmpfs:
-        size: 42G
```
alfonsoros88, line 551: This last 2 lines should be reincorporated. The current test would break with this.

flxo: Check for a later reformat commit of all the tests. The manifest strings are all aligned to two space indentation.
Tests are green ;-)

npk/src/manifest.rs
alfonsoros88: but why not simply let mount handle the format instead of adding the extra logic?

npk/src/manifest.rs
alfonsoros88: I'm referring to the `tmpfs` with the unit letter. The test `serialize_back_and_forth` will break if this lines are kept, which is the reason why you removed them. I'm just not sure if this is a good idea.

PR #205, opened 2020-11-18
Abort the runtime upon panics. Set PR_SET_PDEATHSIG to SIGKILL in child applications

### LINKED ISSUE #52: Signal handling: killing the north-runtime process should terminate running container
When the runtime is terminated by sending a SIGKILL does not terminate running containers

flxo:
When running Tokio, a panic in a task will just cause the thread the task is executed to panic. Setting `panic` to `abort` causes the whole process to exit upon panics in any thread.

Setting PR_SET_PDEATHSIG causes the kernel to signal the application process when the parent (runtime) exits. This  should normally be done via the minijails hooks which are C function pointes that can be registered. Enhancing the minijail Rust wrapper to pass this sort of data to minijail is way more invasive regarding future updates than this one prctl.

Fixes #52

### DIFF
--- a/minijail/libminijail.c
+++ b/minijail/libminijail.c
@@ -3121,6 +3121,13 @@ static int minijail_run_internal(struct minijail *j,
 	minijail_free_run_state(state_out);
 
 	/*
+	 * Set a parent death signal. The kernel will send a SIGKILL to this process
+	 * when the parent exits.
+	 */
+	if(prctl(PR_SET_PDEATHSIG, SIGKILL))
+		pdie("failed to set parent death signal to SIGKILL");
+
+	/*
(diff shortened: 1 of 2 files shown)


minijail/libminijail.c
```diff
@@ -3120,6 +3120,13 @@ static int minijail_run_internal(struct minijail *j,
 	state_out->child_env = NULL;
 	minijail_free_run_state(state_out);
 
+	/*
+	 * Set a parent death signal. The kernel will a SIGKILL to this process
```
alfonsoros88, line 3124: A verb is missing "The kernel will send a SIGKILL..."

flxo: Thanks. done.

PR #213, opened 2020-11-25
Use runtime as library in integration tests

alfonsoros88:
Changes in north:
- Rebased on top of https://github.com/esrlabs/northstar/pull/147.
  Needed to terminatet the minijail log task on drop, otherwise the runtime in the tests hangs.
- Adds a `Error::AsyncRuntime` error type for errors coming from tokio. Needed to add this error to handle some tokio tasks join 
  and some channel send/recv errors.
- The tests use a `test!` macro so that the `tokio` runtime is created after we enter the mount namespace.
- Refactors the check on installed containers so that instances are taken into account when searching for installed containers.
  This is used both, when installing and uninstalling containers.
- `State::install` no longer removes the input npk file. The deletion occurs now on the `console` side when deleting the temporal
  npk file received from the client connection.

Changes in the tests:
- `north_tests` no longer requires the `north` and `nstar` binaries.
- It is possible to access north runtime errors from the tests.
- The logging has colors in the CI now :tada:


flxo:
Configuration the console IO should be done via configuration if needed - could be a uri. Pulling out the `console` module and exposing very internal things feels a little bit odd. Let's talk about a less invasive approach here.

north/src/runtime/mod.rs
```diff
@@ -108,9 +108,13 @@ impl Runtime {
...
-                stopped_tx
-                    .send(runtime_task(config, event_tx, event_rx, stop_rx).await)
-                    .ok(); // Ignore error if calle dropped the handle
+                let result = runtime_task(config, event_tx, event_rx, stop_rx).await;
+                if result.is_err() {
+                    log::error!("Error in runtime task: {:?}", result);
+                } else {
+                    log::debug!("Runtime task exited");
+                }
+                stopped_tx.send(result).ok(); // Ignore error if calle dropped the handle
```
marcmo, line 117: always nicer as a match expression. If you do not use the error, you can be explicit about that with:
```
match runtime_task(...).await {
   Err(_) => ...
   Ok(_) => ...
}
```

north/src/runtime/state.rs
```diff
@@ -407,74 +397,86 @@ impl State {
...
+            .filter(|i| *i > 1)
+            .map(|i| {
+                (0..i)
+                    .into_iter()
+                    .map(|i| format!("{}-{:03}", manifest.name, i))
+                    .collect()
+            })
+            .unwrap_or_else(|| vec![manifest.name.to_string()]);
+
+        for app in instances.iter().filter_map(|name| {
```
marcmo, line 449: That's a little misleading...if there are any running instances, we cannot uninstall the container.
So why not just check if any of the application map keys matches the name + any version and then just return the error?

north/src/runtime/state.rs
```diff
@@ -407,74 +397,86 @@ impl State {
...
 
-            Self::notification(
-                &self.events_tx,
-                Notification::Uninstalled(name.to_owned(), version.clone()),
-            )
-            .await;
+        for instance in instances {
+            let key = (instance, version.clone());
+            self.applications.remove(&key.0);
+            self.resources.remove(&key);
```
marcmo, line 466: Not necessary. If there were any running instances, we would already have bailed.

north_tests/src/runtime.rs
```diff
@@ -14,184 +14,98 @@
...
-            .await
-            .wrap_err_with(|| format!("Failed to stop {}", container_name))
-            .and_then(|result| result)
+    pub async fn stop(&mut self, name: &str) -> Result<()> {
+        timeout(
+            TIMEOUT,
+            self.0.request(api::Request::Stop(name.to_string())),
+        )
+        .await
+        .wrap_err("Stopping contaier timed out")
```
marcmo, line 72: container

north/src/runtime/state.rs
alfonsoros88: Do you mean that we should do the check for running instances separately before we unmount anything? that makes sense.

north/src/runtime/state.rs
alfonsoros88: What is not necessary? we are uninstalling the container, we have to remove it from the state correct?

marcmo: We only have running instances if the app is running...and in this case we won't uninstall anyway

alfonsoros88: I'm not sure if I'm understanding correctly. If there is any running instance, this par of the code wont get executed. Maybe the name "instances" is confusing.

marcmo: you are right... running != installed => so we need to remove the mounted apps that are not running

north/src/runtime/state.rs
```diff
@@ -407,74 +397,91 @@ impl State {
...
-                        fs::remove_file(&entry.path()).await.map_err(|e| {
-                            Error::Io(format!("Failed to remove {}", entry.path().display()), e)
-                        })?;
-                    }
-                }
-            }
+        for app in stopped {
+            info!("Unmounting {}", app);
+            umount_npk(app.container()).await.map_err(Error::Mount)?;
+        }
```
marcmo, line 466: Very nice!

north/src/runtime/state.rs
```diff
@@ -407,74 +397,91 @@ impl State {
...
-                .resources
-                .get(&(name.to_string(), version.clone())) // TODO get rid of copy
-                .is_some(),
-            Some(app) => app.container.manifest.version == *version,
+    async fn find_installed_npk(
+        &self,
+        name: &str,
+        version: &Version,
+    ) -> Result<Option<(Manifest, PathBuf)>, Error> {
+        // Remove npk from registry
```
marcmo, line 405: comment outdated

marcmo:
looks good to me!

PR #219, opened 2020-11-30
examples: Fix unbound variable warning

pclesr:
In bash/sh, any global variable that is modified by a subshell will
lose it's binding when the subshell exits. For a cleanup 'trap' to
work, the scope must be global.

This also fixes an incorrect use of globbing for function params,
discovered by running 'shellcheck'.


flxo:
Thanks @pclesr!

Issue #112, opened 2020-10-07
Activate capabilities dropping in minijail


flxo:
The current minijail integration has disabled capabilities dropping. The reason is the unavailability of `libcap` for the Android build. Within AOSP `libcap` is built along with `minijail` but it's not part of the NDK.

-> Add `libcap` to the workspace
-> Build `libcap` (includes some nasty header generation like in `minijail`
-> Enable capability dropping in `minijail`

PR #222, opened 2020-12-02
Add skeleton libcap functionality

pclesr:
This fixes https://github.com/esrlabs/northstar/issues/112

This is a workaround; the long term solution is to most likely to import libcap into a crate and reference it from there. The issue is that on android, libcap is not part of the toolchain and must be referenced externally.

To avoid problems with musl, android, and x86 this changes minijail to explicitly reference the capabilitiy.h from the embedded libcap. Only the source for routines that are referenced by minijail are pulled in.

With this update, it is now possible to update the linux capabilities of the container.


flxo:
Merged. Thanks!

PR #224, opened 2020-12-04
Add supplementary group and capability handling

pclesr:
This is based on PR https://github.com/esrlabs/northstar/pull/222 

This adds the ability to specify linux capabilities and supplementary groups in the manifest.  

It adds another file to the incorporated capabilities library (`minijail/libcap`) from  `libcap-2.45` . 

The code for handling the group and capability list is in a new file (`minimail/libmj_perms.c`) to try and make merging the already bloated `minijail/libminijail.c `easier in the future.

### DIFF
--- a/examples/container/hello/manifest.yaml
+++ b/examples/container/hello/manifest.yaml
@@ -41,3 +41,5 @@ seccomp:
   exit: 1
   exit_group: 1
   rt_sigreturn: 1
+capability_str: cap_net_raw
+supplgroups_str: inet
--- a/minijail/libcap/cap_names.h
+++ b/minijail/libcap/cap_names.h
@@ -1,0 +1,56 @@
+/*
+ * DO NOT EDIT: this file is generated automatically from
+ *
+ *     <uapi/linux/capability.h>
+ */
+
+#define __CAP_BITS       41
+#define __CAP_NAME_SIZE  23
+
+#ifdef LIBCAP_PLEASE_INCLUDE_ARRAY
+#define LIBCAP_CAP_NAMES { \
+      /* 0 */	"cap_chown", \
+      /* 1 */	"cap_dac_override", \
+      /* 2 */	"cap_dac_read_search", \
+      /* 3 */	"cap_fowner", \
+      /* 4 */	"cap_fsetid", \
+      /* 5 */	"cap_kill", \
+      /* 6 */	"cap_setgid", \
+      /* 7 */	"cap_setuid", \
+      /* 8 */	"cap_setpcap", \
+      /* 9 */	"cap_linux_immutable", \
+      /* 10 */	"cap_net_bind_service", \
+      /* 11 */	"cap_net_broadcast", \
+      /* 12 */	"cap_net_admin", \
+      /* 13 */	"cap_net_raw", \
+      /* 14 */	"cap_ipc_lock", \
+      /* 15 */	"cap_ipc_owner", \
+      /* 16 */	"cap_sys_module", \
+      /* 17 */	"cap_sys_rawio", \
+      /* 18 */	"cap_sys_chroot", \
+      /* 19 */	"cap_sys_ptrace", \
+      /* 20 */	"cap_sys_pacct", \
+      /* 21 */	"cap_sys_admin", \
+      /* 22 */	"cap_sys_boot", \
+      /* 23 */	"cap_sys_nice", \
+      /* 24 */	"cap_sys_resource", \
+      /* 25 */	"cap_sys_time", \
+      /* 26 */	"cap_sys_tty_config", \
+      /* 27 */	"cap_mknod", \
+      /* 28 */	"cap_lease", \
+      /* 29 */	"cap_audit_write", \
+      /* 30 */	"cap_audit_control", \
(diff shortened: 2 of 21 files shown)


marcmo:
I see the need that we specify capabilities but it is now really changing minijail. Let's discuss this with @flxo when he is back.

pclesr:
> I see the need that we specify capabilities but it is now really changing minijail. Let's discuss this with @flxo when he is back.

It is changing minijail, because the interfaces that we require to minijail do not exist. Therefore, new functionality is added in separate files to minimise the challenges of pulling in new versions of minijail. 

@marcmo what is the specific concern?

marcmo:
@pclesr just that it becomes harder to maintain. Yes most of the code is in a separate folder but some merging is still needed. Is that something that we could contribute to the minijail project?

pclesr:
> @pclesr just that it becomes harder to maintain. Yes most of the code is in a separate folder but some merging is still needed. Is that something that we could contribute to the minijail project?

I do not think so; the functionality being used is already present if one uses the minijail command line (which northstar does not). There is no internal API for the approach we need, therefore I had to augment the existing minijail. 

Luckily minijail does not change often. However, the alternative is to write yet another jailer from scratch.

minijail/rust/minijail-sys/build.rs
```diff
@@ -113,6 +113,11 @@ fn main() -> io::Result<()> {
         "../../syscall_wrapper.c",
         "../../system.c",
         "../../util.c",
+        "../../libmj_perms.c",
```
flxo, line 116: We can keep `libcap` in minijail in the first step and separate it into it's own lib later if needed. Since all the cap files are tightly coupled to minijail this is fine for now.

minijail/rust/minijail-sys/libminijail.rs
```diff
@@ -387,3 +387,15 @@ extern "C" {
 extern "C" {
     pub fn minijail_log_to_fd(fd: ::std::os::raw::c_int, min_priority: ::std::os::raw::c_int);
 }
+extern "C" {
```
flxo, line 390: We should get `bindgen` to work in our minijail "fork". Hm. Out of scope for this PR but the correct way. @pclesr can you create a ticket for that: `Use bindgen for libminijail.rs generation`. Maybe there's already a suitable bindgen call in the minijail crate.

npk/src/manifest.rs
```diff
@@ -204,6 +204,13 @@ pub struct Manifest {
     )]
     #[serde(default)]
     pub mounts: HashMap<PathBuf, Mount>,
+    /// String containing capability names to give to
+    /// new container
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub capability_str: Option<String>,
```
flxo, line 210: This should be `Option<Vec<String>>`

npk/src/manifest.rs
```diff
@@ -204,6 +204,13 @@ pub struct Manifest {
...
+    /// String containing group names to give to new container
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub supplgroups_str: Option<String>,
```
flxo, line 213: This should be a `Option<Vec<String>>`

minijail/libmj_perms.c
```diff
@@ -0,0 +1,140 @@
+/*
```
flxo, line 1: I doubt I can review this....

npk/src/manifest.rs
pclesr: Can you (or someone else), then provide a fix ?

I do not have the Rust knowledge to do what you want. I am just following the examples of what is already there.

npk/src/manifest.rs
pclesr: Ditto with previous comment

minijail/rust/minijail-sys/libminijail.rs
pclesr: Will do

minijail/libcap/cap_names.h
```diff
@@ -0,0 +1,56 @@
+/*
```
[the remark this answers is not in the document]
pclesr: Yes, I don't think it is necessary to review anything in the libcap directory. It is taken directly from upstream as stated in the commit, version 2-45. 

I don't think we want to do anything with auto-generating the names; this comes directly from the library.

examples/container/hello/manifest.yaml
```diff
@@ -41,3 +41,5 @@ seccomp:
   exit: 1
   exit_group: 1
   rt_sigreturn: 1
+capability_str: cap_net_raw
```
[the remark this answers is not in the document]
pclesr: I have an example that I use for network testing that eventually will find it's way into the repo.

Agreed, abusing 'hello' is not the way to go.

npk/src/manifest.rs
flxo: Sure. I'm currently a little bit short on time but I think that's just a finger exercise for @marcmo or @alfonsoros88. We can ask in the SU tomorrow.

flxo:
Merged. Thank!

PR #226, opened 2020-12-07
Move boilerplate out of integration tests

alfonsoros88:
Leaves only tests inside the `integration_tests.rs`


marcmo:
looks good

PR #227, opened 2020-12-07
Check API Erros in integration tests

alfonsoros88:
- Wraps `Result<api::Response>` in `must_use` type to check for `API` errors.
- Adds some tests that make assertions for expected `API` errors

### DIFF
--- a/north_tests/src/runtime.rs
+++ b/north_tests/src/runtime.rs
@@ -15,7 +15,7 @@
 //! Controls North runtime instances
 
 use color_eyre::eyre::{eyre, Result, WrapErr};
-use north::{api, api::Response, runtime};
+use north::{api, runtime};
 use std::path::Path;
 use tokio::{time, time::timeout};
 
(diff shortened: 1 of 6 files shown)


north_tests/src/runtime.rs
```diff
@@ -14,98 +14,100 @@
...
+            Ok(api::Response::Err(e)) if err == e => Ok(()),
+            _ => Err(eyre!("Response is not an error")),
+        }
+    }
+
+    pub fn could_fail(self) {}
+}
+
+pub fn timeout_on<R>(f: impl Future<Output = R>) -> Result<R> {
+    futures::executor::block_on(timeout(TIMEOUT, f)).wrap_err("Future timed out")
```
flxo, line 49: Why does this use the `futures` executor?

alfonsoros88: I just wanted to get rid of all the  `await`s on the tests because I think there is no need for them to be asynchronous. Then I could not find a wait to do a `block_on` with `tokio`.

flxo:
Rebased and merged.

PR #228, opened 2020-12-08
Add version field to Manifest struct

inorick:
North runtime now verifies the new optional 'manifest_version' field.
If this field is missing, version '1.0.0' is assumed.
Manifest version '1.0.0' is written to all new manifests created by sextant.

### DIFF
--- a/north/src/runtime/mount.rs
+++ b/north/src/runtime/mount.rs
@@ -25,7 +25,7 @@ use nix::sys::inotify::{AddWatchFlags, InitFlags, Inotify};
 use npk::{
     archive::{ArchiveReader, Container},
     dm_verity::VerityHeader,
-    manifest::{Manifest, Version},
+    manifest::{Manifest, Version, VERSION_1_0_0},
 };
 use std::{
     collections::HashMap,
@@ -437,16 +437,9 @@ async fn wait_for_file_deleted(path: &Path, timeout: time::Duration) -> Result<(
 }
 
 fn is_version_supported(manifest: &Manifest) -> bool {
-    const V1: Version = Version(semver::Version {
-        major: 1,
-        minor: 0,
-        patch: 0,
-        pre: vec![],
-        build: vec![],
-    });
-    const SUPPORTED_MIN_VERSION: &Version = &V1;
-    const KNOWN_MAX_VERSION: &Version = &V1;
-    const DEFAULT_VERSION: &Version = &V1; // used if manifest has no 'manifest_version' entry
+    const SUPPORTED_MIN_VERSION: &Version = &VERSION_1_0_0;
+    const KNOWN_MAX_VERSION: &Version = &VERSION_1_0_0;
+    const DEFAULT_VERSION: &Version = &VERSION_1_0_0;
     let manifest_version = manifest
         .manifest_version
         .as_ref()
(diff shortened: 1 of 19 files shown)


north/src/runtime/mount.rs
```diff
@@ -431,3 +437,32 @@ async fn wait_for_file_deleted(path: &Path, timeout: time::Duration) -> Result<(
...
+    // compare major versions
+    if manifest_version.0.major < SUPPORTED_MIN_VERSION.0.major {
+        return false;
+    }
+    if manifest_version.0.major > KNOWN_MAX_VERSION.0.major {
+        debug!(
+            "Warning: unknown manifest major version {}",
+            manifest_version
+        );
+    }
```
marcmo, line 466: 1.case: npk is too old => not supported by newer runtime
2.case: runtime is too old => npk is not backward compatible (should not happen) but still in this case we should not continue

north/src/runtime/mount.rs
```diff
@@ -431,3 +437,32 @@ async fn wait_for_file_deleted(path: &Path, timeout: time::Duration) -> Result<(
...
+        pre: vec![],
+        build: vec![],
+    });
+    const SUPPORTED_MIN_VERSION: &Version = &V1;
+    const KNOWN_MAX_VERSION: &Version = &V1;
+    const DEFAULT_VERSION: &Version = &V1; // used if manifest has no 'manifest_version' entry
+    let manifest_version = manifest
+        .manifest_version
+        .as_ref()
+        .unwrap_or(DEFAULT_VERSION);
```
marcmo, line 455: Is that safe? In case the manifest does not have a version, the npk might be rather recent or rather old...
OTOH if we disallow npks without a manifest version we will force everyone to rebuild their NPKs...
what do you think?

npk/src/archive.rs
```diff
@@ -194,79 +192,57 @@ const MANIFEST: &str = "manifest.yaml";
...
+    archive_reader.extract_manifest()
 }
 
-impl<'a> ArchiveReader<'a> {
-    pub fn new(
-        npk: &Path,
-        signing_keys: &'a HashMap<String, ed25519_dalek::PublicKey>,
-    ) -> Result<Self, Error> {
+impl ArchiveReader {
+    pub fn new(npk: &Path) -> Result<Self, Error> {
```
marcmo, line 205: much nicer! signing-keys are not used for most AchriveReader operations

marcmo:
Is it. possible to split this PR into 2, one for the `manifest_version` and one for the other stuff?

north/src/runtime/mount.rs
inorick: If we always add manifest version from now on (I'll add it to the pack command), we can assume that all NPKs that are missing it are 'old' NPKs. In that case, we can assume version 1.0.0 for them.
In other words, the 'the npk might be rather recent' case should never occur, right?

north/src/runtime/mount.rs
inorick: Agreed. We should return false in the second case.

inorick:
> Is it. possible to split this PR into 2, one for the `manifest_version` and one for the other stuff?

I moved the archive reader and sextant commit to a new PR (#233). That PR depends on this one.

flxo:
Can we have a chat about the manifest versioning?

marcmo:
@inorick as @flxo pointed out yesterday we did forget something important:
in order to parse the version field, the whole manifest has to be parsed correctly. If we have an older runtime that cannot parse a newer manifest anymore, we also cannot find out the version anymore!
A workaround might be to add a dedicated version file in our manifest. Let's talk about this later

bzld:
We could put the "NPK version" into the ZIP archive comment field...?

inorick:
Superseded by #233

PR #233, opened 2020-12-10
Add manifest version to ZIP comment

inorick:
Read and write manifest version from/to ZIP comments. The initial version is 0.0.1
and should be incremented on manifest changes. NPKs are only accepted if the
manifest version matches the runtime version exactly.

Other changes:
Move signing_keys out of ArchiveReader
Use ArchiveReader in sextant inspect short
Remove hash extraction from ArchiveReader's manifest extraction

### DIFF
--- a/north/src/runtime/mount.rs
+++ b/north/src/runtime/mount.rs
@@ -159,9 +159,11 @@ async fn mount_internal(
     if let Ok(meta) = metadata(&npk).await {
         debug!("Mounting NPK with size {}", meta.len());
     }
+
     let mut archive_reader = ArchiveReader::new(&npk, signing_keys).map_err(Error::Npk)?;
     let hashes = archive_reader.extract_hashes().map_err(Error::Npk)?;
     let manifest = archive_reader.extract_manifest().map_err(Error::Npk)?;
+
     if !is_version_supported(&manifest) {
         return Err(Error::Npk(npk::archive::Error::MalformedManifest(
             "Invalid manifest version".into(),
--- a/north/src/runtime/mount.rs
+++ b/north/src/runtime/mount.rs
@@ -26,7 +26,7 @@ use nix::sys::inotify::{AddWatchFlags, InitFlags, Inotify};
 use npk::{
     archive::ArchiveReader,
     dm_verity::VerityHeader,
-    manifest::{Manifest, Version},
+    manifest::{Manifest, Version, VERSION_1_0_0},
 };
 use std::{
     collections::HashMap,
@@ -438,16 +438,9 @@ async fn wait_for_file_deleted(path: &Path, timeout: time::Duration) -> Result<(
 }
 
 fn is_version_supported(manifest: &Manifest) -> bool {
-    const V1: Version = Version(semver::Version {
-        major: 1,
-        minor: 0,
-        patch: 0,
-        pre: vec![],
-        build: vec![],
-    });
-    const SUPPORTED_MIN_VERSION: &Version = &V1;
-    const KNOWN_MAX_VERSION: &Version = &V1;
-    const DEFAULT_VERSION: &Version = &V1; // used if manifest has no 'manifest_version' entry
+    const SUPPORTED_MIN_VERSION: &Version = &VERSION_1_0_0;
+    const KNOWN_MAX_VERSION: &Version = &VERSION_1_0_0;
+    const DEFAULT_VERSION: &Version = &VERSION_1_0_0;
     let manifest_version = manifest
         .manifest_version
         .as_ref()
@@ -455,13 +448,12 @@ fn is_version_supported(manifest: &Manifest) -> bool {
 
     // compare major versions
     if manifest_version.0.major < SUPPORTED_MIN_VERSION.0.major {
+        debug!("Unsupported manifest major version {}", manifest_version);
         return false;
     }
     if manifest_version.0.major > KNOWN_MAX_VERSION.0.major {
-        debug!(
-            "Warning: unknown manifest major version {}",
-            manifest_version
-        );
+        debug!("Unknown future manifest major version {}", manifest_version);
+        return false;
     }
     true
 }
--- a/north/src/runtime/mount.rs
+++ b/north/src/runtime/mount.rs
@@ -23,11 +23,7 @@ use floating_duration::TimeAsFloat;
 use log::{debug, info};
 pub use nix::mount::MsFlags as MountFlags;
 use nix::sys::inotify::{AddWatchFlags, InitFlags, Inotify};
-use npk::{
-    archive::ArchiveReader,
-    dm_verity::VerityHeader,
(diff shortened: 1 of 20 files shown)


north/src/runtime/mount.rs
```diff
@@ -431,3 +437,24 @@ async fn wait_for_file_deleted(path: &Path, timeout: time::Duration) -> Result<(
...
+    const SUPPORTED_MIN_VERSION: &Version = &VERSION_1_0_0;
+    const KNOWN_MAX_VERSION: &Version = &VERSION_1_0_0;
+    const DEFAULT_VERSION: &Version = &VERSION_1_0_0;
+    let manifest_version = manifest
+        .manifest_version
+        .as_ref()
+        .unwrap_or(DEFAULT_VERSION);
+
+    // compare major versions
+    if manifest_version.0.major < SUPPORTED_MIN_VERSION.0.major {
```
marcmo, line 451: The manifest version change should not be part of this PR. We should keep it separate so that we can further refine the concept since we already found some problems with it yesterday.

npk/src/archive.rs
```diff
@@ -190,83 +192,52 @@ impl TryFrom<SerdeHashes> for Hashes {
...
+        Ok((zip_file.data_start(), zip_file.size()))
     }
 
-    pub fn extract_hashes(&mut self) -> Result<Hashes, Error> {
-        let mut signature_file = self
-            .archive
-            .by_name(SIGNATURE)
-            // .with_context(|| "Failed to read signature".to_string())
-            .map_err(Error::Zip)?;
+    pub fn extract_hashes(
```
marcmo, line 218: We should make the `ArchiveReader` functions async. To do that, take a look at https://docs.rs/tokio/0.3.5/tokio/task/index.html#blocking-and-yielding
this allows the code in an async function to call a blocking API

npk/src/archive.rs
```diff
@@ -190,83 +192,52 @@ impl TryFrom<SerdeHashes> for Hashes {
...
     }
 
-    pub fn extract_manifest_from_archive(&mut self) -> Result<Manifest, Error> {
-        let hashes = self.extract_hashes()?;
-        let mut manifest_file = self
-            .archive
-            .by_name(MANIFEST)
-            .map_err(|e| Error::ArchiveError(format!("Failed to read manifest ({})", e)))?;
-
+    pub fn extract_manifest(&mut self) -> Result<Manifest, Error> {
```
marcmo, line 232: Should also be converted to an async function

npk/src/archive.rs
flxo: Can't we make the npk crate just operate on a `std::io::Read`? This avoid the dependency on `tokio` in the `npk` crate. The wrapping in a `spawn_blocking` etc call is up to the user and the usage of tokio should not be dictated but this crate.

```rust
pub struct Npk<'a, R: io::Read + io::Seek> {
    archive: zip::ZipArchive<io::BufReader<R>>,
    signing_keys: &'a HashMap<String, ed25519_dalek::PublicKey>,
}

impl<'a, R: io::Read + io::Seek> Npk<'a, R> {
    pub fn new(
        npk: R,
        signing_keys: &'a HashMap<String, ed25519_dalek::PublicKey>,
    ) -> Result<Self, Error> {
        let reader = io::BufReader::new(npk);
        let archive = zip::ZipArchive::new(reader).map_err(Error::Zip)?;
        Ok(Self {
            archive,
            signing_keys,
        })
    }
}
```

marcmo: yes that of course is true...keep forgetting that we live in the tokio universe and not in the rust universe here :(

flxo: Why?

north/src/runtime/mount.rs
```diff
@@ -157,13 +158,22 @@ async fn mount_internal(
...
-
-    let manifest = archive_reader
-        .extract_manifest_from_archive()
+    let mut archive_reader = ArchiveReader::new(&npk).map_err(Error::Npk)?;
+    let hashes = archive_reader
+        .extract_hashes(&signing_keys)
+        .map_err(Error::Npk)?;
+    let manifest = archive_reader.extract_manifest().map_err(Error::Npk)?;
+    let manifest_version = archive_reader
+        .extract_manifest_version()
```
flxo, line 168: `archive_reader.npk_version()`

north/src/runtime/mount.rs
```diff
@@ -157,13 +158,22 @@ async fn mount_internal(
...
         .map_err(Error::Npk)?;
+    if manifest_version != Manifest::VERSION {
+        return Err(Error::Npk(npk::archive::Error::MalformedManifest(format!(
+            "Invalid manifest version (detected: {}, supported: {})",
```
flxo, line 172: Invalid NPK version

npk/src/archive.rs
```diff
@@ -190,83 +195,66 @@ impl TryFrom<SerdeHashes> for Hashes {
...
-            archive,
-            signing_keys,
-        })
+        Ok(Self { archive })
+    }
+
+    pub fn extract_manifest_version(&mut self) -> Result<Version, Error> {
+        let zip_comment = String::from_utf8(self.archive.comment().to_vec())
+            .map_err(|e| Error::ArchiveError(format!("Failed to read manifest version {}", e)))?;
+        let split_comment: Vec<_> = zip_comment.split(' ').collect();
```
flxo, line 215: This could be done with this pattern:
```rust
let mut split = zip_comment.split(' ');
while let Some(key) = split.next() {
  if let Some(value) = split.next() {
   // todo
 } else {
   return Err("Missing value for key);
 }
}
```

sextant/src/inspect.rs
```diff
@@ -21,7 +22,35 @@ use std::{
...
+        .context("Failed to find manifest in NPK")?;
+    let name = manifest.name.to_string();
+    let version = manifest.version.to_string();
+    let manifest_version = reader
+        .extract_manifest_version()
+        .map_or("unknown".to_string(), |v| v.to_string());
+    let is_resource_container = manifest.init.map_or("yes", |_| "no");
+    let instances = manifest.instances.unwrap_or(1).to_string();
+    println!(
+        "name: {}, version: {}, manifest version: {}, resource container: {}, instances: {}",
```
flxo, line 46: NPK version

flxo (changes requested):

npk/src/archive.rs
marcmo: let's do that in another pr

marcmo:
Nicer now with the version in the zip comments!

marcmo:
@flxo you requested changes but I think that has been resolved...can we merge this?

inorick:
> @flxo you requested changes but I think that has been resolved...can we merge this?

Yes, this PR can be merged. It was my understanding that the original reviewer is required for a merge. Merge away!

PR #236, opened 2020-12-11
Ignore send error in waitpid

flxo:
If a process is SIGKILLED it's immediatelly removed from the
list of known processes. The rx part of the exit handle is dropped
but the waitpid task is not stopped yet (not possible because of
the blocking waitpid call) and tries to send to a dead end channel.
Ignore the send error in this case.


flxo:
Fixes #235

Issue #174, opened 2020-11-10
Refactor the container_dir settings. Provide a way to configure multiple individual repositories


flxo:
Currently the keys are referenced in the configuration with a directory. The runtime searches for files within that directory which is not intuitive. 

The configuration should look like this:

```toml
debug = true
console_address = "localhost:4200"
key = "examples/keys/north.pub"
container_uid = 1000
container_gid = 1000

[directories]
container_dirs = [ "target/north/registry" ]
run_dir = "target/north/run"
data_dir = "target/north/data"
```

bzld:
The `signature.yaml` contains a `key:` field to support having different signing keys between NPKs (owned by different parties for example).
That's why the folder approach was used to search for the right one to use.

Maybe we don't need this flexbility.

In case we _do_ want to keep it, we should rather have a key name -> key file mapping in configuration? Like:
```toml
[keys]
north = "examples/keys/north.pub"
external = "examples/keys/external.pub"
```

flxo:
Or even tie it up to the `container_dir`:

```
[registries] // Not sure if that's a good name at all
foo:
   key: /system/etc/north/blah...
bar:
```
This would also allow to drive a "registry" dir without verity.

bzld:
Yes, interesting middle ground 😉 
I think the right name for that would be `[repositories]` instead of `[registries]`?

flxo:
I tried to sketch this in `config.rs`:

* The key is optional for a repository and indicated that the npk shouldn't be mounted via verity
* `runtime::state::State::applications` gets a reference to a `Repository`
* `runtime::state::State::signing_keys` is changed and holds a key for each repository.
* The `writeable` flag indicated whether npk can be added/removed at runtime
* The `install`/`uninstall` api commands must get a repository id
* Npks with the same name in the manifest are not allowed
* `api::Container` gets a repository id
* `api::Request::Repositories` is added and returns a `Map<api::RepositoryId, api::Repository>` (note the `api::`)
 

```rust
/// Identification of a repository. This is used during add and remove operations
/// to identify the location.
type RepositoryId = String;

#[derive(Clone, Debug, Deserialize)]
pub struct Repository {
    /// Repository directory
    pub dir: PathBuf,
    /// Writeable
    pub writeable: bool,
    /// Signing key. If this is None the containers are mounted without verity
    pub key: Option<PathBuf>,
}

#[derive(Clone, Debug, Deserialize)]
pub struct Config {
    /// Print debug logs
[...]

PR #237, opened 2020-12-11
Refactor container dir setting

alfonsoros88:
solves #174

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+* The `[directories]` section in the `north.toml` configuration file is replaced
+  by multiple `[repositories.<name>]` sections. Where each corresponds to a
+  directory previously specified in the `container_dirs` parameter. The
+  specified `<name>` corresponds to the __repository identifier__.
+* Each `[repositories.<name>]` section has a `dir`, `writable` and an optional
+  `key` parameter. Where `dir` is the directory containing the `npk` files,
+  `writable` indicates whether cotainers can be installed or uninstalled to this
+  repository and `key` is the path to the public signature file.
+* To install a container, a __repository identifier__ is required to indicate
+  the destination where the container will be installed.
+* If the `key` parameter is unspecified, the containers from the repository are
+  mounted without being verified using `verity`.
 
 ## [0.6.0] - 2020-11-22
 ### Added
--- a/README.md
+++ b/README.md
@@ -148,9 +148,9 @@ The script in `doc/tools/check_conf.sh` can be used to check your running kernel
 Open a terminal, clone the repository and build the solution and start the runtime:
 
     $ git clone https://github.com/esrlabs/northstar.git
-    $ cargo build --release --bin north 
-    $ cargo build --release --bin nstar 
-    $ ./examples/build_examples.sh 
+    $ cargo build --release --bin north
+    $ cargo build --release --bin nstar
+    $ ./examples/build_examples.sh
     $ sudo ./target/release/north
 
 ### Run sample code
@@ -158,23 +158,23 @@ Open a second terminal navigate to the directory where the northstar source is l
 Execute the following commands:
 
 1. `containers` to list all registered containers
-2. `start crashing` to start an example container which will crash within the next 10 seconds 
+2. `start crashing` to start an example container which will crash within the next 10 seconds
 
-[//]: # 
- 
+[//]: #
+
 
     $ ./target/release/nstar
     $ >> containers
-    Name              | Version | Type     | PID | Uptime 
+    Name              | Version | Type     | PID | Uptime
     -------------------+---------+----------+-----+--------
-    cpueater          | 0.0.1   | App      |     |  
-    crashing          | 0.0.1   | App      |     |  
-    datarw            | 0.0.1   | App      |     |  
-    ferris_says_hello | 0.0.3   | App      |     |  
-    hello             | 0.0.2   | App      |     |  
-    memeater          | 0.0.1   | App      |     |  
-    ferris            | 0.0.2   | Resource |     |  
-    hello_message     | 0.1.2   | Resource |     |  
+    cpueater          | 0.0.1   | App      |     |
+    crashing          | 0.0.1   | App      |     |
+    datarw            | 0.0.1   | App      |     |
+    ferris_says_hello | 0.0.3   | App      |     |
+    hello             | 0.0.2   | App      |     |
+    memeater          | 0.0.1   | App      |     |
+    ferris            | 0.0.2   | Resource |     |
+    hello_message     | 0.1.2   | Resource |     |
     >> start crashing
     crashing-0.0.1 was started
     start succeeded
@@ -210,12 +210,14 @@ debug = true
 console_address = "localhost:4200"
 container_uid = 1000
 container_gid = 1000
-
-[directories]
-container_dirs = [ "target/north/registry" ]
 run_dir = "target/north/run"
 data_dir = "target/north/data"
-key_dir = "examples/keys"
+
+[repositories.examples]
+dir = "target/north/registry"
+writable = true
+key = "examples/keys/north.pub"
+
 
 [cgroups]
 memory = "north"
@@ -230,12 +232,16 @@ device_mapper = "/dev/mapper/control"
 device_mapper_dev = "/dev/dm-"
 ```
 
-The `[directories]` section just tells north what directories to use.
+The `[repositories.examples]` describes a container repository named `examples`.
+Within it, the following options can be specified:
 
-* **`container_dir`** -- list of directories where to find the `*.npk` packages for the correct architecture
-  are to be found
-* **`run_dir`** -- where the container content will be mounted
-* **`data_dir`** -- In data_dir a directory for each container is created if a mount of type data is used in the manifest
+* **`dir`** -- The directory where to find `*.npk` packages for the correct architecture.
+* **`writable`** -- That indicates if containers can be installed/uninstalled
+    from this repository.
+* **`key`** -- The path to the public signing key used to sign the containers.
+
+Multiple `[repositories.<name>]` sections can be specified for separate
+repositories, where `<name>` is the __repository identifier__.
 
 The [`cgroups`] optionally configures northstar applications CGroups settings.
 Both `memory` and `cpu` will tell northstar where to mount the cgroup hierarchies.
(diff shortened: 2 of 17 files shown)


north/src/runtime/state.rs
```diff
@@ -127,20 +136,54 @@ impl fmt::Display for Application {
...
+            repositories,
             applications: HashMap::new(),
             resources: HashMap::new(),
             config: config.clone(),
         })
     }
 
+    /// Mount container repositories
+    pub(crate) async fn mount_repositories(&mut self) -> Result<(), Error> {
+        let repos: Vec<Repository> = self.repositories.values().cloned().collect();
```
flxo, line 172: This is a deep copy. Probably not needed.

north/src/runtime/state.rs
```diff
@@ -332,40 +380,58 @@ impl State {
...
     }
 
     /// Install a npk
-    pub async fn install(&mut self, npk: &Path) -> Result<(), Error> {
-        let manifest = read_manifest(npk, &self.signing_keys).map_err(Error::Npk)?;
+    pub async fn install(&mut self, id: RepositoryId, npk: &Path) -> Result<(), Error> {
+        let repository = self
+            .repositories
+            .get(&id)
+            .ok_or_else(|| Error::Repository(format!("Failed to find repository ID: {}", id)))?;
```
flxo, line 412: "Failed to find repository with id {}"

north/src/runtime/state.rs
```diff
@@ -468,17 +527,18 @@ impl State {
...
         }
 
-        let application_containers = instances
-            .iter()
-            .filter_map(|name| self.applications.get(name).map(|a| a.container()));
-        let resource_containers = instances
-            .iter()
-            .filter_map(|name| self.resources.get(&(name.clone(), version.clone())));
-        let containers = application_containers.chain(resource_containers);
+        let containers = instances.iter().filter_map(|name| {
```
flxo, line 530: Nice!

north_tests/src/runtime.rs
```diff
@@ -74,8 +74,8 @@ impl Runtime {
         Ok(())
     }
 
-    pub async fn install(&mut self, npk: &Path) -> Result<()> {
-        let response = timeout(TIMEOUT, self.0.install(npk))
+    pub async fn install(&mut self, repo: api::RepositoryId, npk: &Path) -> Result<()> {
```
flxo, line 77: Take a ref for `repo`

nstar/src/main.rs
```diff
@@ -35,11 +35,12 @@ mod codec;
...
-shutdown:                   Stop the northstar runtime
-start <name>:               Start application
-stop <name>:                Stop application
-install <file>:             Install/Update npk
-uninstall <name> <version>: Unstall npk";
+repositories:                 List available repositories
+shutdown:                     Stop the northstar runtime
+start <name>:                 Start application
+stop <name>:                  Stop application
+install <repository> <file>:  Install/Update npk into repository
```
flxo, line 42: The comment is wrong. We cannot update just uninstall and install.

nstar/src/main.rs
```diff
@@ -229,9 +230,26 @@ async fn main() -> Result<()> {
                                 }
                             };
                         }
+                        Some("repositories") => {
+                            match request_response(&mut framed_write, &mut response_rx, Request::Repositories).await {
+                                Ok(Some(Response::Repositories(r))) => {
+                                    if opt.json {
```
flxo, line 236: Oh. I really have to finish the refactoring of nstar.... This hurts...

flxo (changes requested):
@alfonsoros88 Left you a couple of nitpicks. Think we can later on optimise the mounting stuff and combine things as you suggested.
Good work!

north/src/api.rs
```diff
@@ -13,12 +13,13 @@
...
 use derive_new::new;
-use serde::{Deserialize, Serialize};
-
 use npk::manifest::{Manifest, Version};
+use serde::{Deserialize, Serialize};
+use std::{collections::HashMap, path::PathBuf};
 
 pub type Name = String;
 pub type MessageId = String; // UUID
+pub use npk::archive::RepositoryId;
```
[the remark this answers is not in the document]
alfonsoros88: Actually it's not necessary, good catch. I thought it was going to be part of the `api` instead of `npk`.

north/src/runtime/console.rs
```diff
@@ -138,8 +144,8 @@ impl Console {
                     warn!("Received message is not a request");
                 }
             }
-            Request::Install(message, path) => {
-                let payload = match state.install(&path).await {
+            Request::Install(message, repo, path) => {
+                let payload = match state.install(repo.clone(), &path).await {
```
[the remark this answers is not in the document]
alfonsoros88: Yes, I will change it to `&RepositoryId`.

north/src/runtime/error.rs
```diff
@@ -38,10 +40,12 @@ pub enum Error {
     Console(super::console::Error),
     #[error("Cgroups: {0}")]
     Cgroups(#[from] super::cgroups::Error),
-    #[error("Mount: {0:?}")]
+    #[error("Mount: {0}")]
```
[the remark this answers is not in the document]
alfonsoros88: The reason why this was changed is that with the debug print you lose the error message and you can only see the structure, is it still preferable to use the debug print?

north/src/runtime/mod.rs
```diff
@@ -158,14 +158,14 @@ impl Runtime {
     }
 
     /// Installs a container in the repository
-    pub async fn install(&self, npk: &Path) -> Result<api::Response, Error> {
+    pub async fn install(&self, repo: RepositoryId, npk: &Path) -> Result<api::Response, Error> {
```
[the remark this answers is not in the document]
alfonsoros88: this was exactly the reason. I would not like to use `&str` since that assumes `&RepositoryId` can coerce to a string reference but I guess that's probably a detail we can ignore for now.

north/src/runtime/mount.rs
```diff
@@ -62,67 +64,50 @@ pub enum Error {
...
+        .map_err(|e| Error::Io(format!("Failed to read {}", &repo.dir.display()), e))?
         .filter_map(move |d| d.ok())
         .map(|d| d.path());
 
-    let dm = Dm::new(&device_mapper).map_err(Error::DeviceMapper)?;
-    let lc = LoopControl::open(loop_control, loop_dev)
-        .await
-        .map_err(Error::LoopDevice)?;
-
+    let mut containers: Vec<Container> = vec![];
```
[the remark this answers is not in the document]
alfonsoros88: Unfortunately, there is no `flatten` method implemented for `StreamExt`. I think this is as good as it get's for now.

north/src/api.rs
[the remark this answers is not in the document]
flxo: ok

north/src/api.rs
```diff
@@ -19,7 +19,7 @@ use std::{collections::HashMap, path::PathBuf};
 
 pub type Name = String;
 pub type MessageId = String; // UUID
-pub use npk::archive::RepositoryId;
+use npk::archive::RepositoryId;
```
flxo, line 22: Rustfmt should move this to the above use block.

flxo (approved):

north/src/api.rs
alfonsoros88: it does not, but i can move it :)

CHANGELOG.md
```diff
@@ -6,6 +6,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Changed
+* The `[directories]` section in the `north.toml` configuration file is replaced
+  by multiple `[repositories.<name>]` sections. Where each corresponds to a
+  directory previously specified in the `container_dirs` parameter. The
+  specified `<name>` corresponds to the __repository identifiet__.
```
marcmo, line 13: identifier

CHANGELOG.md
```diff
@@ -6,6 +6,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+* Each `[repositories.<name>]` section has a `dir`, `writable` and an optional
+  `key` parameter. Where `dir` is the directory containing the `npk` files,
+  `writable` indicates whether cotainers can be installed or uninstalled to this
+  repository and `key` is the path to the public signature file.
```
marcmo, line 17: the writeable description does not make sense. A container can be installed into this section, but is itself not writable...at least that's how I understand it.

CHANGELOG.md
```diff
@@ -6,6 +6,19 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
+* To install a container, a __repository identifier__ is required to indicate
+  the destination where the container will be installed.
+* If the `key` parameter is unspecified, the containers from the repository are
+  mounted without being verified using `verity`.
```
marcmo, line 21: Aren't npks int a writable section always required to provide a key parameter?

README.md
```diff
@@ -230,12 +232,16 @@ device_mapper = "/dev/mapper/control"
...
-The `[directories]` section just tells north what directories to use.
+The `[repositories.examples]` describes a container repository named `examples`.
+Within it, the following options can be specified:
 
-* **`container_dir`** -- list of directories where to find the `*.npk` packages for the correct architecture
-  are to be found
-* **`run_dir`** -- where the container content will be mounted
-* **`data_dir`** -- In data_dir a directory for each container is created if a mount of type data is used in the manifest
+* **`dir`** -- The directory where to find `*.npk` packages for the correct architecture.
+* **`writable`** -- That indicates if containers can be installed/uninstalled
```
marcmo, line 239: same thing as above

north/src/runtime/error.rs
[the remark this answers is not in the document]
marcmo: the problem is that `thiserror` does not include underlying errors when printing with `{0}` ...do you really loose the error message with this?

north/src/runtime/error.rs
```diff
@@ -29,6 +29,8 @@ pub enum Error {
     MissingResource(String),
     #[error("Container {0} already installed")]
     ContainerAlreadyInstalled(String),
+    #[error("Repository: {0}")]
+    Repository(String),
```
marcmo, line 33: There are only 2 error conditions with a repository:
- not found
- not writable
I would make those 2 distinct errors without a `String`

north/src/runtime/state.rs
```diff
@@ -332,19 +367,44 @@ impl State {
         }
 
         for (_name, container) in self.applications.values().map(|a| (a.name(), &a.container)) {
-            umount_npk(container).await.map_err(Error::Mount)?;
+            let wait_for_dm = self
+                .repositories
+                .get(&container.repository)
+                .map(|r| r.key.is_some())
+                .unwrap_or(false);
```
marcmo, line 374: Why is it relevant if a key exists to know if we want to wait for the device mapping?

alfonsoros88: yes, this is pretty much a hack. Checking for the key here indicates whether the container is mounted through verity or directly in the loopback device. In the verity case, we have to wait for the device mapper device to  be removed indirectly after removing the loop device.

marcmo: ok..then I'd call it rather something like `verity_dm_used`

PR #239, opened 2020-12-16
Fix uninstall function

alfonsoros88:


flxo:
Thanks @alfonsoros88!

PR #248, opened 2021-01-11
Replace ArchiveReader with Npk struct

inorick:
Move functionality from ArchiveReader struct into Npk struct to unify sextant and runtime npk functionality. Remove archive.rs.


northstar/src/runtime/mount.rs
```diff
@@ -52,6 +59,8 @@ pub enum Error {
     JoinError(tokio::task::JoinError),
     #[error("Os error: {0:?}")]
     Os(nix::Error),
+    #[error("Repository error: {0:?}")]
+    Repository(String),
```
flxo, line 63: What is a repository error? Maybe this is more a "InvalidRepository" or "UnknownRepository"

northstar/src/runtime/mount.rs
```diff
@@ -72,23 +81,36 @@ pub(super) async fn mount_npk_repository(
 
 pub(super) async fn mount_npk(
     config: &Config,
-    npk: &Path,
+    npk_path: &Path,
```
flxo, line 84: the `_path` is redundant with the type `Path`

northstar/src/runtime/mount.rs
```diff
@@ -72,23 +81,36 @@ pub(super) async fn mount_npk_repository(
...
     }
 
-    let archive_reader = ArchiveReader::new(&npk, repo.key.as_ref()).map_err(Error::Npk)?;
-    let manifest = archive_reader.manifest();
-    let npk_version = archive_reader.npk_version();
-    if *npk_version != Manifest::VERSION {
-        return Err(Error::Npk(npk::archive::Error::MalformedManifest(format!(
+    let file = std::fs::File::open(&npk_path)
+        .map_err(|e| Error::Io(format!("Failed to open NPK at {}", npk_path.display()), e))?;
+    let mut npk = Npk::new(file).map_err(Error::Npk)?;
```
flxo, line 98: This needs to go into a `block_in_place`

northstar/src/runtime/mount.rs
```diff
@@ -72,23 +81,36 @@ pub(super) async fn mount_npk_repository(
...
+    if let Some(pub_key) = repo.key {
+        npk.verify(&pub_key).map_err(Error::Npk)?;
+    } else {
+        return Err(Error::Repository(
+            "No public key in repository to verify NPK signature".to_string(),
+        ));
+    }
+
+    let manifest = npk.manifest().map_err(Error::Npk)?;
+    let npk_version = npk.version().map_err(Error::Npk)?;
```
flxo, line 110: This needs to be done in `block_in_place`

northstar/src/runtime/mount.rs
```diff
@@ -238,23 +260,33 @@ pub async fn verity_setup(
...
+        Ok((
+            npk.verity_header()?,
+            npk.hashes()?,
+            npk.fsimg_offset()?,
+            npk.fsimg_size()?,
+        ))
+    })
+    .map_err(Error::Npk)?;
+
+    let mut fs = fs::File::open(npk_path)
```
flxo, line 281: This needs a `block_in_place`

northstar/src/runtime/mount.rs
```diff
@@ -263,7 +295,7 @@ async fn setup_and_mount(
 
     if !use_verity {
         let device = loop_device.path().await.unwrap();
-        mount(&device, root, fs_type, MountFlags::MS_RDONLY, None).await?;
+        mount(&device, root, &FS_TYPE, MountFlags::MS_RDONLY, None).await?;
```
flxo, line 298: If we assume to have squashfs in the npk it should be validated in `Npk::new`. On the other hand a npk is not assembled manually and therefore is just contains what is done in sextant. Never mind.

northstar/src/runtime/state.rs
```diff
@@ -388,17 +390,17 @@ impl State {
         Ok(())
     }
 
-    /// Install a npk
-    pub async fn install(&mut self, id: &str, npk: &Path) -> Result<(), Error> {
+    /// Install an NPK
+    pub async fn install(&mut self, id: &str, npk_path: &Path) -> Result<(), Error> {
```
flxo, line 394: the `_path` is redundant with the type `Path`

northstar/src/runtime/state.rs
```diff
@@ -388,17 +390,17 @@ impl State {
...
             .ok_or_else(|| Error::RepositoryNotFound(id.to_owned()))?;
 
-        let manifest = ArchiveReader::new(npk, repository.key.as_ref())
-            .map_err(Error::Npk)?
-            .manifest()
-            .clone();
+        let file = std::fs::File::open(&npk_path)
+            .map_err(|e| Error::Io(format!("Failed to open NPK at {}", npk_path.display()), e))?;
+        let mut npk = Npk::new(file).map_err(Error::Npk)?;
+        let manifest = npk.manifest().map_err(Error::Npk)?;
```
flxo, line 403: This should be done in a `block_in_place`

northstar/src/runtime/state.rs
```diff
@@ -457,10 +459,13 @@ impl State {
                 Error::Io(format!("Failed to read {}", repository.dir.display()), e)
             })?;
             while let Ok(Some(entry)) = dir.next_entry().await {
-                let manifest = ArchiveReader::new(&entry.path().as_path(), None)
-                    .map_err(Error::Npk)?
-                    .manifest()
-                    .clone();
+                let manifest = tokio::task::block_in_place(|| {
+                    let file = File::open(entry.path().as_path()).map_err(|e| {
+                        npk::npk::Error::FileOperation(format!("Failed to open NPK: {}", e))
```
flxo, line 464: ` npk::npk::Error::FileOperation` is a `Io` error. In all the `Error` enums we have a `Io` that wraps a `io::Error`. Think we should rename this to `Io`.

northstar_tests/src/test_container.rs
```diff
@@ -53,7 +53,7 @@ lazy_static! {
         npk::pack(
             package_dir.path(),
             REPOSITORIES_DIR.path(),
-            Path::new("examples/keys/northstar.key"),
+            Option::from(Path::new("examples/keys/northstar.key")),
```
flxo, line 56: Very uncommon way of producing a `Some` ;-)

northstar_tests/src/test_container.rs
```diff
@@ -62,7 +62,7 @@ lazy_static! {
         npk::pack(
             Path::new("northstar_tests/test_resource"),
             REPOSITORIES_DIR.path(),
-            Path::new("examples/keys/northstar.key"),
+            Option::from(Path::new("examples/keys/northstar.key")),
```
flxo, line 65: Very uncommon way of producing a `Some` ;-)

npk/Cargo.toml
```diff
@@ -5,6 +5,7 @@ authors = ["ESRLabs"]
 edition = "2018"
 
 [dependencies]
+anyhow = "1.0.34"
```
flxo, line 8: Why is anyhow needed in the `npk` crate? It has it's own error type and there's no need for anyhow

If it's just used in the tests it should be reference in `dev.dependencies`.

npk/src/dm_verity.rs
```diff
@@ -146,15 +146,14 @@ impl VerityHeader {
 
     pub fn check(&self) -> Result<(), Error> {
         if !self.header.starts_with(VerityHeader::HEADER) {
-            return Err(Error::InvalidHeader);
+            Err(Error::InvalidHeader)
+        } else if self.version != VerityHeader::VERITY_VERSION {
+            Err(Error::UnsupportedVersion(self.version))
+        } else if !self.algorithm.starts_with(VerityHeader::ALGORITHM) {
+            Err(Error::UnsupportedAlgorithm())
```
flxo, line 153: Think the brackets can be removed

npk/src/manifest.rs
```diff
@@ -12,6 +12,7 @@
 //   See the License for the specific language governing permissions and
 //   limitations under the License.
 
+use anyhow::Result;
```
flxo, line 15: `anyhow` should no be necessary in a lib crate.

npk/src/manifest.rs
```diff
@@ -447,10 +448,18 @@ impl Manifest {
...
     });
 
-    fn verify(&self) -> Result<(), ManifestError> {
+    pub fn from_yaml<R: io::Read>(reader: R) -> Result<Self, Error> {
+        serde_yaml::from_reader(reader).map_err(Error::SerdeYaml)
+    }
+
+    pub fn to_yaml<W: io::Write>(&self, writer: W) -> Result<(), Error> {
+        serde_yaml::to_writer(writer, self).map_err(Error::SerdeYaml)
+    }
```
flxo, line 457: I'm struggling with the names. `from_yaml` seems to me more like a `from_reader`.

npk/tests/npk.rs
```diff
@@ -39,7 +39,7 @@ env:
         let key_dir = create_tmp_dir();
         create_test_manifest(&src);
         let (_pub_key, prv_key) = gen_test_key(&key_dir);
-        pack(&src, &dest, &prv_key).expect("Pack NPK");
+        pack(&src, &dest, Option::from(prv_key.as_path())).expect("Pack NPK");
```
flxo, line 42: `Option::from` is a `Some`

flxo (changes requested):

northstar/src/runtime/mount.rs
inorick: We only use it in case there is "No public key in repository to verify NPK signature". Better rename it to 'MissingKey' error or something like that.

northstar/src/runtime/mount.rs
inorick: It is used to distinguish it from the `npk: Npk<File>` instance used below.
Shadowing `npk_path` is also not an optiona as is needed to call setup_and_mount in lin 140.

northstar/src/runtime/mount.rs
inorick: Do you mean this call? Isn't it async already?
let mut npk_file = fs::File::open(npk_path).await ...

npk/src/dm_verity.rs
inorick: Doesn't look like it:
```error[E0308]: mismatched types
   --> npk/src/dm_verity.rs:153:17
    |
40  |     UnsupportedAlgorithm(),
    |     ---------------------- fn() -> dm_verity::Error {dm_verity::Error::UnsupportedAlgorithm} defined here
...
153 |             Err(Error::UnsupportedAlgorithm)
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected enum `dm_verity::Error`, found fn item
    |
    = note: expected enum `dm_verity::Error`
            found fn item `fn() -> dm_verity::Error {dm_verity::Error::UnsupportedAlgorithm}`
help: use parentheses to instantiate this tuple variant
    |
153 |             Err(Error::UnsupportedAlgorithm())
    |          ```

PR #249, opened 2021-01-11
Move api codec into api module. Add a client implementation

flxo:


northstar/src/api/codec.rs
```diff
@@ -0,0 +1,144 @@
...
+    type Error = io::Error;
+
+    fn decode(&mut self, src: &mut bytes::BytesMut) -> Result<Option<Self::Item>, Self::Error> {
+        if let Some(position) = src.iter().position(|b| *b == b'\n') {
+            let buf = src.split_to(position);
+            // Consume the newline
+            src.advance(1);
+            match serde_json::from_slice::<model::Message>(&buf) {
+                Ok(message) => Ok(Some(message)),
+                Err(e) => Err(io::Error::new(ErrorKind::Other, e)),
```
marcmo, line 53: better use `ErrorKind::InvalidData`

northstar/src/api/codec.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    fn poll_flush(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Pin::new(&mut self.framed).poll_flush(cx)
+    }
+
+    fn poll_close(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
+        Pin::new(&mut self.framed).poll_close(cx)
+    }
+}
```
marcmo, line 144: The `AsyncWrite`, `AsyncRead`, `Sink` and `Stream` implementation looks a lot like boilerplate code. My understanding of the codec package was that you can use `FramedRead` and `FramedWrite` to achieve this.

flxo: `FramedWrite` and `AsyncWrite` work different. If we want to use `FramedWrite` (mean `Sink`) for sending the install stream we need to introduce another enum variant for `Framed::Item` and chunk manually. This is safe but probably slower than this direct access to the wrapped `AsyncWrite`. The intention to newtype `Framed` is to use it as `Stream`, `Sink` *and* `AsyncWrite`.

northstar/src/api/codec.rs
flxo: With `ErrorKind::Invalid` we'd discard the error information from `serde`.

marcmo: why is this?
`Err(e) => Err(io::Error::new(ErrorKind::InvalidData, e)),`
what information is lost here?

flxo: Ah. Its a `ErrorKind`. My bad. This makes sense. I'll change that.

marcmo:
would be nice to add some info about this change to our changelog

flxo:
> would be nice to add some info about this change to our changelog

Done.

PR #251, opened 2021-01-12
Add comp alg and block size to sextant pack

### LINKED ISSUE #250: Add sextant parameters for mksquashfs's compression algorithm and block size
The supported compression algorithms by squashfs depend on the kernel configuration. Different targets might support different compression algorithms. For this reason, it makes sense to add a parameter to sextant to specify the compression algorithm used to pack the container rootfs. 

An other important parameter is the size of the compressed blocks of data. It has a influence both in the access speed as well as the size of the resulting file.

alfonsoros88:
closes #250

### DIFF
--- a/examples/build_examples.sh
+++ b/examples/build_examples.sh
@@ -17,6 +17,13 @@ else
   PLATFORM=${1}
 fi
 
+if [ -z ${2+x} ]
+then
+  COMPRESSION_ALGORITHM="gzip"
+else
+  COMPRESSION_ALGORITHM=${2}
+fi
+
 exe() { echo " + $*" ; $* ; }
 
 log_err() {
@@ -97,7 +104,7 @@ build_example() {
     provision_artifact "${NAME}" "${ROOT_DIR}"
   fi
 
-  exe cargo run --bin sextant -- pack --dir "${TMP_DIR}" --out "${OUTPUT_DIR}" --key "./examples/keys/northstar.key"
+  exe cargo run --bin sextant -- pack --dir "${TMP_DIR}" --out "${OUTPUT_DIR}" --key "./examples/keys/northstar.key" --comp "${COMPRESSION_ALGORITHM}"
 }
 
 main() {
--- a/examples/build_examples.sh
+++ b/examples/build_examples.sh
@@ -10,19 +10,48 @@ this_script=$(basename $0)
 bold=$(tput bold)
 normal=$(tput sgr0)
 
-if [ -z ${1+x} ]
-then
-  PLATFORM="host"
-else
-  PLATFORM=${1}
-fi
+usage() {
+    echo "USAGE:"
+    echo "    build_examples.sh [OPTIONS]"
+    echo ""
+    echo "OPTIONS:"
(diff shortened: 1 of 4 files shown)


examples/build_examples.sh
```diff
@@ -17,6 +17,13 @@ else
   PLATFORM=${1}
 fi
 
+if [ -z ${2+x} ]
+then
+  COMPRESSION_ALGORITHM="gzip"
+else
+  COMPRESSION_ALGORITHM=${2}
+fi
+
```
flxo, line 26: I't probably time to document the two optional arguments for this script in the README.md

The enum can have a `Display` impl to be used when invoking `mksquashfs`

npk/src/npk.rs
```diff
@@ -84,10 +84,32 @@ pub enum Error {
     },
 }
 
+/// Squashfs Options
+pub struct SquashfsOpts<'a> {
+    /// The compression algorithm used (default gzip)
+    pub comp: &'a str,
```
flxo, line 90: Would be great to introduce a enum for the compression algorithm. This allows `clap` to hint if the argument is unknown.

sextant/src/main.rs
```diff
@@ -34,6 +34,12 @@ enum Opt {
         /// Output directory
         #[structopt(short, long)]
         out: PathBuf,
+        /// Compression algorithm to use in squashfs (default gzip)
+        #[structopt(short, long)]
+        comp: Option<String>,
```
flxo, line 39: Use a enum here instead of `String` and declare a `default_value`

flxo (changes requested):
Think we can improve the CLI experience by typing the compression algorithm.

npk/src/npk.rs
```diff
@@ -350,10 +359,64 @@ impl Builder {
...
+    pub comp: CompAlg,
+    /// Size of the blocks of data compressed separately
+    pub block_size: u32,
+}
+
+impl Default for SquashfsOpts {
+    fn default() -> Self {
+        SquashfsOpts {
+            comp: CompAlg::Gzip,
+            block_size: 131072,
```
flxo, line 406: Where does this number come from?

npk/src/npk.rs
```diff
@@ -350,10 +359,64 @@ impl Builder {
...
+        }
+    }
+}
+
+/// Squashfs Options
+pub struct SquashfsOpts {
+    /// The compression algorithm used (default gzip)
+    pub comp: CompAlg,
+    /// Size of the blocks of data compressed separately
+    pub block_size: u32,
```
flxo, line 399: Could this be optional? If it's None take the default of `mksquashfs`

npk/src/npk.rs
```diff
@@ -350,10 +359,64 @@ impl Builder {
     }
 }
 
+#[derive(Debug)]
+pub enum CompAlg {
```
flxo, line 363: Rename this to `CompressionAlgorithm`

npk/src/npk.rs
alfonsoros88: That's the default value used by `mksquashfs`

PR #253, opened 2021-01-13
northcracker_jan13

pclesr:
This includes both network namespace and Northcracker support.


flxo:
Renamed the branch to `ex-firecracker` and set it to protected.

PR #256, opened 2021-01-19
Make NPK functionality async

inorick:
Move all NPK functionality to tokio async.

### DIFF
--- a/northstar/src/runtime/mount.rs
+++ b/northstar/src/runtime/mount.rs
@@ -29,13 +29,16 @@ use npk::{
     npk::{Error::MalformedManifest, Npk},
 };
 use std::{
-    fs::File,
     io,
     path::{Path, PathBuf},
     process,
 };
 use thiserror::Error;
-use tokio::{fs, fs::metadata, task, time};
+use tokio::{
+    fs,
+    fs::{metadata, File},
+    task, time,
+};
 
 const FS_TYPE: &str = "squashfs";
 
@@ -99,19 +102,17 @@ impl MountControl {
             debug!("Mounting NPK with size {}", meta.len());
         }
 
-        let (manifest, npk_version) = tokio::task::block_in_place(|| {
-            let file = std::fs::File::open(&npk)
-                .map_err(|e| Error::Io(format!("Failed to open NPK at {}", npk.display()), e))?;
-            let mut npk_archive = Npk::new(file).map_err(Error::Npk)?;
+        let file = File::open(&npk)
+            .await
+            .map_err(|e| Error::Io(format!("Failed to open NPK at {}", npk.display()), e))?;
+        let mut npk_archive = Npk::new(file).await.map_err(Error::Npk)?;
 
-            // verify signature only if a key is present
-            if let Some(pub_key) = repo.key {
-                npk_archive.verify(&pub_key).map_err(Error::Npk)?;
-            }
-
-            let npk_version = npk_archive.version().map_err(Error::Npk)?;
-            Ok((npk_archive.manifest().map_err(Error::Npk)?, npk_version))
-        })?;
+        // verify signature only if a key is present
+        if let Some(pub_key) = repo.key {
+            npk_archive.verify(&pub_key).await.map_err(Error::Npk)?;
+        }
+        let manifest = npk_archive.manifest().await.map_err(Error::Npk)?;
+        let npk_version = npk_archive.version().await.map_err(Error::Npk)?;
 
         if npk_version != Manifest::VERSION {
             return Err(Error::Npk(MalformedManifest(format!(
@@ -268,19 +269,14 @@ impl MountControl {
         root: &Path,
         use_verity: bool,
     ) -> Result<PathBuf, Error> {
-        // open NPK synchronously
-        let (verity, hashes, fsimg_offset, fsimg_size) = tokio::task::block_in_place(|| {
-            let file = File::open(npk)
-                .map_err(|e| npk::npk::Error::Io(format!("Failed to open NPK: {}", e)))?;
-            let mut npk = Npk::new(file)?;
-            Ok((
-                npk.verity_header()?,
-                npk.hashes()?,
-                npk.fsimg_offset()?,
-                npk.fsimg_size()?,
-            ))
-        })
-        .map_err(Error::Npk)?;
+        let npk_file = fs::File::open(npk)
+            .await
+            .map_err(|e| Error::Io("Failed to open NPK".to_string(), e))?;
+        let mut npk_archive = Npk::new(npk_file).await.map_err(Error::Npk)?;
+        let verity = npk_archive.verity_header().await.map_err(Error::Npk)?;
+        let hashes = npk_archive.hashes().await.map_err(Error::Npk)?;
+        let fsimg_offset = npk_archive.fsimg_offset().await.map_err(Error::Npk)?;
+        let fsimg_size = npk_archive.fsimg_size().await.map_err(Error::Npk)?;
 
         let mut npk_file = fs::File::open(npk)
             .await
--- a/npk/src/npk.rs
+++ b/npk/src/npk.rs
@@ -18,7 +18,6 @@ use crate::{
     dm_verity::{append_dm_verity_block, Error as VerityError, VerityHeader, BLOCK_SIZE},
     manifest::{Manifest, Mount, MountFlag, Version},
 };
-use const_format::concatcp;
 use ed25519_dalek::{
     ed25519::signature::Signature, Keypair, PublicKey, SecretKey, SignatureError, Signer,
     SECRET_KEY_LENGTH,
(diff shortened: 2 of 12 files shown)


sextant/Cargo.toml
```diff
@@ -15,4 +15,5 @@ northstar = {path = "../northstar"}
 npk = { path = "../npk" }
 structopt = "0.3.21"
 tempfile = "3.1.0"
+tokio = { version = "1.0.1", features = ["full"] }
```
flxo, line 18: Here we can identify the features subset of tokio instead of using `full`. Same for the runtime

npk/src/npk.rs
```diff
@@ -152,15 +154,15 @@ pub struct Npk<R: io::Read + io::Seek> {
 }
 
 impl<R: io::Read + io::Seek> Npk<R> {
```
flxo, line 156: This is key: The signature requires `std` types but is async. Thats not ok from API perspective.
The `zip` crates needs a `std::io::Read`. We want to provide an api with `tokio::io::*`. Two things must be done:
When creating the `ZipArchive` the `tokio` types must be wrapped or converted into std types *and* every call to `ArchiveReader` must be wrapped in a `block_in_place` or `spawn_blocking` call.

npk/src/npk.rs
```diff
@@ -477,36 +479,42 @@ pub fn unpack(npk: &Path, out: &Path) -> Result<(), Error> {
...
+    write(&key_pair.secret.to_bytes(), &prv_key).await?;
     Ok(())
 }
 
-pub fn open_zipped_npk(npk: &Path) -> Result<ZipArchive<File>, Error> {
-    let zip = zip::ZipArchive::new(File::open(&npk).map_err(|e| Error::Os {
-        context: format!("Failed to open NPK at '{}'", &npk.display()),
-        error: e,
-    })?)
+pub async fn open_zipped_npk(npk: &Path) -> Result<ZipArchive<std::fs::File>, Error> {
```
flxo, line 507: npks are always zipped ;-)

npk/src/npk.rs
```diff
@@ -152,15 +154,15 @@ pub struct Npk<R: io::Read + io::Seek> {
 }
 
 impl<R: io::Read + io::Seek> Npk<R> {
-    pub fn new(npk: R) -> Result<Self, Error> {
+    pub async fn new(npk: R) -> Result<Self, Error> {
         let zip = zip::ZipArchive::new(npk).map_err(|e| Error::Zip {
```
flxo, line 158: Probably we cannot construct a `ArchiveReader` from a `tokio::io::Read/Seek`.

npk/src/npk.rs
inorick: I am not sure how this can be done:
> When creating the ZipArchive the tokio types must be wrapped or converted into std types
Wouldn't this require some kind of adapter that wraps the async object and implements the synchronous `Read` and `Seek` traits?
```struct AsyncReadAdapter<R: AsyncRead + AsyncSeek> {
    inner: R,
}

impl<R> io::Read for AsyncReadAdapter<R>
where
    R: AsyncRead + AsyncSeek,
{
    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
        unimplemented!()
    }
}

impl<R> io::Seek for AsyncReadAdapter<R>
where
    R: AsyncRead + AsyncSeek,
{
    fn seek(&mut self, pos: SeekFrom) -> io::Result<u64> {
        unimplemented!()
    }
}```
If so, the required `read()` and `seek()` methods would be synchronous themselves but have to call async code at the same time.

File: npk/src/npk.rs (hunk 4 above)
Comment (inorick) [reply npk/src/npk.rs:158]: I think this is true. The `Npk` struct is a wrapper around a `ZipArchive` which does not support async. Most of our IO heavy operations go through the `zip` crate and are therefore synchronous. Slapping an async label on our struct will not change that until we find an async zip implementation.

File: npk/src/npk.rs [hunk 5]
```diff
@@ -459,66 +461,72 @@ pub fn pack_with(
...
+    let zip = zip::ZipArchive::new(
+        File::open(&file)
+            .await
+            .map_err(|e| Error::Os {
+                context: format!("Failed to open file: '{}'", &file.display()),
+                error: e,
+            })?
+            .into_std()
+            .await,
+    )
```
alfonsoros88, line 517: I think `zip::ZipArchive::new` is still a synchronous function and it should be wrapped in a `task::block_in_place` or a `task::spawn_blocking`. Maybe consider:

```rust
pub async fn open_zip(file: &Path) -> Result<ZipArchive<std::fs::File>, Error> {
    let path = file.to_owned();
    task::block_in_place(move || {
        let file = std::fs::File::open(&path).map_err(|e| Error::Os {
            context: format!("Failed to open file: '{}'", &file.display()),
            error: e,
        })?;
        zip::ZipArchive::new(file).map_err(|e| Error::Zip {
            context: format!("Failed to parse ZIP format: '{}'", path.display()),
            error: e,
        })
    })
}
```

npk/src/npk.rs
```diff
@@ -692,7 +707,7 @@ fn sign_hashes(key_pair: &Keypair, hashes_yaml: &str) -> String {
     signature_yaml
 }
 
-fn copy_src_root_to_tmp(src: &Path, tmp: &TempDir) -> Result<PathBuf, Error> {
+async fn copy_src_root_to_tmp(src: &Path, tmp: &TempDir) -> Result<PathBuf, Error> {
```
marcmo, line 710: Just putting the `async` keyword before a rust function does not make the function asynchronous.
I this case it this function still blocks unless the call to `fs_extra::dir::copy` is dealt with correctly.
The same logic applies to all the other functions where just the `async` keyword was added.

npk/src/npk.rs
```diff
@@ -848,19 +863,19 @@ fn write_npk<W: Write + Seek>(
...
             error: e,
         })?;
-    let mut fsimg = File::open(&fsimg).map_err(|e| Error::Os {
+    let fsimg = File::open(&fsimg).await.map_err(|e| Error::Os {
         context: format!("Failed to open '{}'", &fsimg.display()),
         error: e,
     })?;
 
-    io::copy(&mut fsimg, &mut zip).map_err(|e| Error::Os {
+    io::copy(&mut fsimg.into_std().await, &mut zip).map_err(|e| Error::Os {
```
alfonsoros88, line 871: probably this should be `tokio::io::copy` instead

northstar/src/runtime/mount.rs
```diff
@@ -267,18 +270,14 @@ async fn setup_and_mount(
...
-    })
-    .map_err(Error::Npk)?;
+    let file = File::open(npk)
+        .await
+        .map_err(|e| Error::Io("Failed to open NPK".to_string(), e))?;
+    let mut npk_archive = Npk::new(file).await.map_err(Error::Npk)?;
+    let verity = npk_archive.verity_header().await.map_err(Error::Npk)?;
+    let hashes = npk_archive.hashes().await.map_err(Error::Npk)?;
+    let fsimg_offset = npk_archive.fsimg_offset().await.map_err(Error::Npk)?;
+    let fsimg_size = npk_archive.fsimg_size().await.map_err(Error::Npk)?;
```
flxo, line 280: nice!

npk/src/npk.rs
```diff
@@ -18,6 +18,7 @@ use crate::{
     dm_verity::{append_dm_verity_block, Error as VerityError, VerityHeader, BLOCK_SIZE},
     manifest::{Manifest, Mount, MountFlag, Version},
 };
+use const_format::concatcp;
```
flxo, line 21: Do we need another dependency for just doing constant strings that never change? If I got it right it's just used for `FS_IMG_NAME`, `MANIFEST_NAME` and `SIGNATURE_NAME`

npk/src/npk.rs
```diff
@@ -302,7 +313,7 @@ impl<R: io::Read + io::Seek> Npk<R> {
         }
     }
 
-    pub fn into_inner(self) -> ZipArchive<R> {
+    pub async fn into_inner(self) -> ZipArchive<std::fs::File> {
```
flxo, line 316: Shouldn't this return a `tokio::fs::File` since that's use for the construction of a `Zip`?

npk/src/npk.rs
```diff
@@ -755,113 +796,123 @@ fn create_squashfs_img(
...
+    let squashfs_root = out.join("squashfs-root");
+
+    if task::block_in_place(|| which::which(&UNSQUASHFS_BIN).is_err()) {
         return Err(Error::Squashfs(format!(
             "Failed to locate '{}'",
             &UNSQUASHFS_BIN
         )));
     }
-    if !image.exists() {
+    if !task::block_in_place(|| image.exists()) {
```
flxo, line 826: Think this `block_in_place` is not needed.

npk/src/npk.rs
inorick: The second argument is unfortunately a synchronous zip file.

npk/src/npk.rs
inorick: Isn't checking for file existence considered a slow IO operation?

npk/src/npk.rs
```diff
@@ -674,14 +709,14 @@ fn gen_pseudo_files(manifest: &Manifest) -> Result<NamedTempFile, Error> {
...
         }
     }
 
     Ok(pseudo_file_entries)
 }
 
-fn sign_hashes(key_pair: &Keypair, hashes_yaml: &str) -> String {
+async fn sign_hashes(key_pair: &Keypair, hashes_yaml: &str) -> String {
     let signature = key_pair.sign(hashes_yaml.as_bytes());
     let signature_base64 = base64::encode(signature);
```
marcmo, line 721: missing: `block_in_place`

inorick: This happens in memory only (`hashes_yaml` is a `&str`). Do we want to block on that too or only on disk IO?

PR #257, opened 2021-01-19
Preserve minijail log pipe fd when starting apps

alfonsoros88:


northstar/src/runtime/minijail.rs
```diff
@@ -273,9 +279,12 @@ impl Process {
         let stderr =
             CaptureOutput::new(OutputStream::Stderr, &manifest.name, event_tx.clone()).await?;
 
+        // Preserve the log fd to avoid missing errors between close_open_fds and execve
+        let log_fd = (*MINIJAIL_LOG_FD, *MINIJAIL_LOG_FD);
+
         let pid = jail.run_remap_env_preload(
             &init.as_path(),
-            &[(stdout.0, 1), (stderr.0, 2)],
+            &[(stdout.0, 1), (stderr.0, 2), log_fd],
```
flxo, line 287: ah. This is the important part. :-)

northstar/src/runtime/minijail.rs
```diff
@@ -42,35 +43,40 @@ use tokio::{
     select, task, time,
 };
 
+lazy_static! {
```
flxo, line 46: `lazy_static` is handy but horrible regarding error handling as you can see with the `unwraps`.
This block also assumes that it's accessed in a sync block on  a tokio runtime.

The `fd` represents some kind of state for the minijail module and therefore I think we should maintain it via composition. Initializing the minijail logging lazily also has an impact on the startup time of the *first* container ever started which might fuzz test.

northstar/src/runtime/minijail.rs
flxo: We just need the `fd` as argument for `start`.  `State` can get a `minijail: Minijail` that holds this fd initialised from `state::State::init`

```rust
structd Minijail {
  log_fd: i32,
}

impl Minijail {
  pub fn init() -> Result<Self, Error> {
   ...
  }
```

flxo:
@alfonsoros88 merged.
 I merged the structs `MinijailLog` and `Minijail`. The `start/spawn` function now is a member of this.
Thank!

PR #260, opened 2021-01-19
Add struct with state for mounts that holds lc and dm for reuse

### LINKED ISSUE #141: Optimise LoopControl and DeviceMapper instantiation
The control structures for loopback and the device mapper are instantiated with a `open` upon each installation call - see `runtime::npk::linux::mount`. Those can be instantiated once at startup and be kept in a e.g `Linux` struct in `State`.

alfonsoros88:
closes #141

### DIFF
--- a/northstar/src/runtime/mount.rs
+++ b/northstar/src/runtime/mount.rs
@@ -63,101 +63,250 @@ pub enum Error {
     MissingKey(String),
 }
 
-pub(super) async fn mount_npk_repository(
-    config: &Config,
-    repo: &Repository,
-) -> Result<Vec<Container>, Error> {
-    info!("Mounting NPKs from {}", repo.dir.display());
-    let mut dir = fs::read_dir(&repo.dir)
-        .await
-        .map_err(|e| Error::Io("Failed to read repository".to_string(), e))?;
-    let mut containers = Vec::new();
-    while let Ok(Some(entry)) = dir.next_entry().await {
-        containers.push(mount_npk(&config, &entry.path(), repo).await?);
-    }
-
-    Ok(containers)
+#[derive(Debug)]
+pub struct MountControl {
+    dm: dm::Dm,
+    lc: LoopControl,
+    device_mapper_dev: String,
 }
 
-pub(super) async fn mount_npk(
-    config: &Config,
-    npk: &Path,
-    repo: &Repository,
-) -> Result<Container, Error> {
-    debug!("Mounting {}", npk.display());
-    let use_verity = repo.key.is_some();
+/// Creates the mount point directory for the corresponding container
+async fn create_mount_point(run_dir: &Path, manifest: &Manifest) -> Result<PathBuf, Error> {
+    let mnt = run_dir
+        .join(&manifest.name)
+        .join(manifest.version.to_string());
+    if !mnt.exists() {
+        info!("Creating mount point {}", mnt.display());
+        fs::create_dir_all(&mnt)
+            .await
+            .map_err(|e| Error::Io(format!("Failed to create directory {}", mnt.display()), e))?;
+    }
+    Ok(mnt)
+}
 
-    let start = time::Instant::now();
-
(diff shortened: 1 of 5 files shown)


northstar/src/runtime/mount.rs
```diff
@@ -63,101 +63,271 @@ pub enum Error {
...
+#[derive(Debug)]
+pub struct MountControl {
+    dm: dm::Dm,
+    lc: LoopControl,
+    run_dir: PathBuf,
+    device_mapper_dev: String,
+}
 
-    Ok(containers)
+async fn mount_point(run_dir: &Path, manifest: &Manifest) -> Result<PathBuf, Error> {
```
flxo, line 74: The name is very generic. Probably this is more a e.g `run_dir_mount_point`.

northstar/src/runtime/mount.rs
```diff
@@ -63,101 +63,271 @@ pub enum Error {
...
-        .map_err(|e| Error::Io("Failed to read repository".to_string(), e))?;
-    let mut containers = Vec::new();
-    while let Ok(Some(entry)) = dir.next_entry().await {
-        containers.push(mount_npk(&config, &entry.path(), repo).await?);
-    }
+#[derive(Debug)]
+pub struct MountControl {
+    dm: dm::Dm,
+    lc: LoopControl,
+    run_dir: PathBuf,
```
flxo, line 70: I'd keep the `run_dir` detail out of `MountControl` it's to closely coupled to the `State` impl. `MountControl` should be something generic for mounting. The pathes can be passed as arguments.

northstar/src/runtime/mount.rs
```diff
@@ -63,101 +63,271 @@ pub enum Error {
...
     }
 
-    let (manifest, npk_version) = tokio::task::block_in_place(|| {
-        let file = std::fs::File::open(&npk)
-            .map_err(|e| Error::Io(format!("Failed to open NPK at {}", npk.display()), e))?;
-        let mut npk_archive = Npk::new(file).map_err(Error::Npk)?;
+    pub(super) async fn mount_npk(
+        &self,
+        npk: &Path,
+        repo: &Repository,
```
flxo, line 122: The `repo` shouldn't be needed if `npk` is a `Path`.

northstar/src/runtime/mount.rs
```diff
@@ -63,101 +63,271 @@ pub enum Error {
...
+        &self,
+        repo: &Repository,
+    ) -> Result<Vec<Container>, Error> {
+        info!("Mounting NPKs from {}", repo.dir.display());
+        let mut dir = fs::read_dir(&repo.dir)
+            .await
+            .map_err(|e| Error::Io("Failed to read repository".to_string(), e))?;
+        let mut containers = Vec::new();
+        while let Ok(Some(entry)) = dir.next_entry().await {
+            containers.push(self.mount_npk(&entry.path(), repo).await?);
```
flxo, line 113: I'd move this to `State` and keep `MountControl` dedicated to mounting.

northstar/src/runtime/mount.rs
alfonsoros88: I don't understand :(  `repo` is necessary for the `key` to check with `verity` and to tag the `container` with the repository id.

flxo: Hm. A repository is basically the key, the name and the dir. You're right to pass the repo instead of it's content split into single arguments.

northstar/src/runtime/mount.rs
```diff
@@ -63,101 +63,250 @@ pub enum Error {
...
+        )))
     }
 
-    let name = format!(
-        "north_{}_{}_{}",
-        process::id(),
-        manifest.name,
-        manifest.version
-    );
+    pub async fn verity_setup(
```
flxo, line 186: does this need to be `pub`?

flxo (approved):
Looks good to me!

PR #261, opened 2021-01-22
Pr installation check

marcmo:


northstar_tests/tests/integration_tests.rs
```diff
@@ -90,16 +90,10 @@ test!(missing_resource_container, {
...
-    runtime
-        .start("test_container")
-        .expect_err(api::model::Error::MissingResource(
-            "test_resource".to_owned(),
-        ))?;
-
-    runtime.uninstall("test_container", "0.0.1").expect_ok()?;
+    runtime.install(get_test_container_npk()).expect_err_kind(
+        api::model::Error::MissingResource("test_resource".to_owned()),
+    )?;
```
alfonsoros88, line 96: The only thing that is not that nice with `expect_err_kind` is that the enum's value is ignored. Perhaps we can use something more like:

```rust
    let response = runtime.install(get_test_container_npk());
    assert!(matches!(
        response.0,
        Ok(api::model::Response::Err(api::model::Error::MissingResource(_)))
    ));
```
Maybe we leave that for the next iteration of the integration tests.

alfonsoros88 (commented):
Looks good,  there are only a couple of comments.

northstar/src/runtime/state.rs
```diff
@@ -436,6 +439,20 @@ impl State {
...
+        });
+        let mut required_resources_stream = stream::iter(required_resources);
+        while let Some((name, version)) = required_resources_stream.next().await {
+            if let Ok(false) = self.resource_available(name, version).await {
+                return Err(Error::MissingResource(format!(
+                    "Resource {}.{} unavailable",
+                    name, version
+                )));
+            }
+        }
```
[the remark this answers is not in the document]
marcmo: good point!

northstar/src/runtime/state.rs
```diff
@@ -400,10 +401,12 @@ impl State {
...
     /// Install an NPK
     pub async fn install(&mut self, id: &str, npk: &Path) -> Result<(), Error> {
-        let repository = self
-            .repositories
-            .get(id)
-            .ok_or_else(|| Error::RepositoryNotFound(id.to_owned()))?;
+        let repository = self.repositories.get(id).ok_or_else(|| {
+            Error::RepositoryIdUnknown(
+                id.to_owned(),
+                self.repositories().keys().map(|id| id.into()).collect(),
```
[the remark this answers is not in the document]
marcmo: yes but this is a bit overdone since we will never have a lot of repositories

northstar_tests/tests/integration_tests.rs
marcmo: I don't think you can compile that...you need an expression here, not a wildcard identifier

northstar/src/api/client.rs
```diff
@@ -194,7 +194,7 @@ impl Client {
             Response::Ok(()) => Err(Error::Protocol),
```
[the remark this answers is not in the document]
marcmo: not sure what you mean here...?

northstar_tests/tests/integration_tests.rs
alfonsoros88: That does indeed compile, but the inner field of `ApiResponse` has to be turn public. But we can better leave this for a general refactoring of the tests.

northstar/src/api/client.rs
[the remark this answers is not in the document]
alfonsoros88: I mean that this arm, returns a `Err(Error::Protocol)` just as the wildcard `_` arm. So you could leave this arm to be handled by the wildcard one. In other words, I think this line should be removed.

northstar/src/runtime/state.rs
```diff
@@ -400,10 +400,12 @@ impl State {
...
     /// Install an NPK
     pub async fn install(&mut self, id: &str, npk: &Path) -> Result<(), Error> {
-        let repository = self
-            .repositories
-            .get(id)
-            .ok_or_else(|| Error::RepositoryNotFound(id.to_owned()))?;
+        let repository = self.repositories.get(id).ok_or_else(|| {
+            Error::RepositoryIdUnknown(
+                id.to_owned(),
+                self.repositories.keys().map(|id| id.into()).collect(),
```
alfonsoros88, line 406: is it ok then computing this every time the error is generated?

alfonsoros88: didn't see this was answered somewhere :) my bad

marcmo: I think this is a much simple solution like it is. And I'd say simplicity outweighs the cost (which is negligible here)

northstar_tests/tests/integration_tests.rs
marcmo: nice...the `match!` macro makes this already possible! even better!

PR #263, opened 2021-01-26
Batteries included debug features

flxo:
* Attach `strace` when starting an application
* Configure runtime log level
* Disable mount namespaces
* Redirect application stdout/stderr to somewhere

### DIFF
--- a/northstar.toml
+++ b/northstar.toml
@@ -1,3 +1,5 @@
+# The runtime log level: TRACE, DEBUG, INFO, WARN, ERROR
+log_level = "DEBUG"
 console = "tcp://localhost:4200"
 run_dir = "target/northstar/run"
 data_dir = "target/northstar/data"
@@ -15,8 +17,6 @@ device_mapper = "/dev/mapper/control"
 device_mapper_dev = "/dev/dm-"
 
 [debug.runtime]
-# The runtime log level: TRACE, DEBUG, INFO, WARN, ERROR
-log_level = "DEBUG"
 # Disable the mount namespace feature. If this optional flag is set to
 # true the runtime will not enter a mount namespace upon startup. This
 # allows to examine the mounted roots of the npks in the repositories.
--- a/northstar.toml
+++ b/northstar.toml
@@ -1,11 +1,7 @@
-debug = true
 console = "tcp://localhost:4200"
 run_dir = "target/northstar/run"
 data_dir = "target/northstar/data"
-
-[repositories.default]
-dir = "target/northstar/repository"
-key = "examples/keys/northstar.pub"
+log_dir = "target/northstar/logs"
 
 [cgroups]
 memory = "northstar"
@@ -17,3 +13,38 @@ loop_control = "/dev/loop-control"
 loop_dev = "/dev/loop"
 device_mapper = "/dev/mapper/control"
 device_mapper_dev = "/dev/dm-"
+
+[debug.runtime]
+# The runtime log level: TRACE, DEBUG, INFO, WARN, ERROR
+log_level = "DEBUG"
+# Disable the mount namespace feature. If this optional flag is set to
+# true the runtime will not enter a mount namespace upon startup. This
+# allows to examine the mounted roots of the npks in the repositories.
+# 
+# WARNING: Be aware that this prevents automatic umounts of the npks when
+# the runtime quits abnormally.
+disable_mount_namespace = false
+
+# # Start a `strace -p PID ...` instance after a container is started.
+# # The execution of the application is defered until strace is attached.
+# [debug.strace]
+# # Confgure the output of the strace instance attached to a started
+# # application. "file" for a file named strace-<PID>-name.log or "log"
+# # to forward the strace output to the runtimes log.
+# output = "file"
+# # Optional additional flags passed to `strace`
+# flags = "-v -f -s 256"
+# # Optional path to the strace binary
+# path = /bin/strace
+
+# # Start a `perf record -p PID -o LOG_DIR/perf-PID-NAME.perf FLAGS` instance
+# # after a container is started.
+# [debug.perf]
+# # Optional path to the perf binary
+# path = /bin/perf
+# # Optional additional flags passed to `perf`
+# flags = ""
+
+[repositories.default]
+dir = "target/northstar/repository"
+key = "examples/keys/northstar.pub"
--- a/northstar/src/main.rs
+++ b/northstar/src/main.rs
@@ -37,13 +37,11 @@ fn main() -> Result<(), Error> {
     let config: Config = toml::from_str(&config)
         .with_context(|| format!("Failed to read configuration file {}", opt.config.display()))?;
 
-    let debug = config.debug.as_ref().and_then(|d| d.runtime.as_ref());
-
-    // Log level
-    let level = debug.and_then(|r| r.log_level).unwrap_or(log::Level::Info);
-
     logd_logger::builder()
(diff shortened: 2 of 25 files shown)


northstar.toml
```diff
@@ -20,3 +16,15 @@ loop_control = "/dev/loop-control"
...
 device_mapper = "/dev/mapper/control"
 device_mapper_dev = "/dev/dm-"
+
+[debug.runtime]
+log_level = "DEBUG"
+disable_mount_namespace = false
+
+# [debug.strace]
+# log_to_file = true
+# flags = "-v -f -s 256"
```
marcmo, line 26: Maybe we should not just comment out a section to disable it.
In case of `debug.runtime` it is also misleading since commenting it out will not turn off logging but select a default level.
e.g. for `[debug.strace]` it could be:
```
enabled = false
...
```

northstar/src/main.rs
```diff
@@ -41,36 +37,45 @@ fn main() -> Result<(), Error> {
...
-        nix::mount::MsFlags::MS_PRIVATE,
-        Option::<&'static [u8]>::None,
-    )?;
+    // Skip mount namespace setup in case it's disabled for debugging purposes
+    if !config
+        .debug
+        .as_ref()
+        .and_then(|c| c.runtime.as_ref())
+        .map(|r| r.disable_mount_namespace)
+        .unwrap_or_default()
```
marcmo, line 64: more clear:
`unwrap_or(false)`

northstar.toml
```diff
@@ -20,3 +16,15 @@ loop_control = "/dev/loop-control"
 loop_dev = "/dev/loop"
 device_mapper = "/dev/mapper/control"
 device_mapper_dev = "/dev/dm-"
+
+[debug.runtime]
+log_level = "DEBUG"
+disable_mount_namespace = false
```
marcmo, line 22: very useful!! :)

northstar/src/runtime/config.rs
```diff
@@ -17,6 +17,29 @@ use std::{collections::HashMap, path::PathBuf};
...
+    pub data_dir: PathBuf,
+    /// Directory for logfile
+    pub log_dir: PathBuf,
+
+    pub repositories: HashMap<RepositoryId, Repository>,
+    pub cgroups: CGroups,
+    pub devices: Devices,
+
+    /// Debugging options
+    pub debug: Option<Debug>,
```
marcmo, line 40: why optional? you already can enable/disable individual sections

northstar/src/runtime/minijail.rs
```diff
@@ -13,75 +13,50 @@
...
 #[derive(Debug)]
-pub struct Minijail {
+pub struct Minijail<'a> {
     log_fd: i32,
     event_tx: EventTx,
-    run_dir: PathBuf,
-    data_dir: PathBuf,
-    uid: u32,
-    gid: u32,
+    config: &'a Config,
```
marcmo, line 53: is it worth dragging in lifetimes for the config?
in my experience, this will usually lead to more trouble down the road so I try to use clones where appropriate.

northstar/src/runtime/minijail.rs
```diff
@@ -162,9 +129,9 @@ impl Minijail {
         }
 
         // Configure UID
-        jail.change_uid(self.uid);
+        jail.change_uid(self.config.container_uid);
```
marcmo, line 132: nice

northstar/src/runtime/minijail.rs
```diff
@@ -214,27 +181,54 @@ impl Minijail {
...
+            } else {
+                let io = Io::new(io).await?;
+                fds.push((io.0, 2));
+                Some(io)
+            }
+        } else {
+            None
+        };
+        if let Some(io) = manifest.io.as_ref().and_then(|io| io.stdin.as_ref()) {
+            if io == &IoInput::Pipe {
```
marcmo, line 210: it's probably only my lacking linux knowledge but this is hard to read without any comment

alfonsoros88 (commented):
looks good to me :)

npk/src/manifest.rs
```diff
@@ -153,8 +155,45 @@ pub enum Mount {
...
+
+#[derive(Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
+pub enum IoInput {
+    /// Inherit the runtimes stdin
+    #[serde(rename = "pipe")]
+    Pipe,
+}
+
+#[derive(Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
+pub enum IoOutput {
```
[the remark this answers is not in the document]
flxo: I renamed it to `Output`, since it's used in `Io`.

northstar/src/runtime/minijail.rs
```diff
@@ -477,45 +564,116 @@ impl AsyncRead for AsyncPipe {
...
-        stream: OutputStream,
-        tag: &str,
-        event_tx: EventTx,
-    ) -> Result<CaptureOutput, Error> {
-        let pipe = AsyncPipe::new()?;
-        let writefd = pipe.writefd();
-        let mut lines = io::BufReader::new(pipe).lines();
-        let tag = tag.to_string();
-        let (tx, mut rx) = oneshot::channel();
+struct Io(i32, CancellationToken);
```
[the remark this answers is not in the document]
flxo: I converted the tuple struct but would like to keep the `Io` name.

marcmo:
AFAICT looks very clever although I have to admit that I don't fully understand the fd handling (preserving in j).
I would appreciate it very much if @flxo could explain this to me (and maybe if it is interesting to @alfonsoros88 and @pclesr )?

marcmo:
looks good

northstar/src/runtime/minijail.rs
flxo: In this case it's super easy with `'a` for `State` and all fields.

northstar/src/runtime/config.rs
flxo: I wanted to avoid having a struct filled with `Nones`.

northstar.toml
flxo: For the logs I agree that we can make the default explicit because "something" is done from the default.

For `strace` I'd like to keep it optional since if it's not there - *nothing* is done.

flxo: Move `log_level` to the main section.

northstar/src/runtime/minijail.rs
flxo: In case of `IoOutput::Pipe` just add the fd from the rt (1 or 2) in the child process. This is done by add a (1, 1) or (2, 2) to the list of `preserved_fds`.

Otherwise create a `Io` that provides a fd that is mapped to 1 or 2 in the child.

npk/src/manifest.rs
```diff
@@ -153,8 +155,45 @@ pub enum Mount {
...
 }
 
-pub type CGroup = HashMap<String, String>;
-pub type CGroups = HashMap<String, CGroup>;
+/// IO configuration for stdin, stdout, stderr
+#[derive(Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
+pub struct Io {
+    /// stdin configuration
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub stdin: Option<IoInput>,
```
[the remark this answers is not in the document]
flxo: The default is to discard stdout, stderr. This is done by minijail by opening /dev/null and using this for stdout/stderr.
The `Pipe` is optional.

PR #264, opened 2021-01-26
Configure UID and GID in container's manifest

alfonsoros88:


flxo (approved):
@alfonsoros88 The `cp` is short for checkpoint and just something I used to tell myself: `do not push this is WIP`. Would be great to replace this with a concrete commit msg.

PR #265, opened 2021-01-27
Overlay default repository in integration tests

alfonsoros88:
This change makes the repository used by the runtime in the integration tests to be overlayed on top of the one with the example containers. In other words, the repository with the example containers is treated as read-only, and containers such as the `test_container` that are installed/uninstalled to the repository go away after the test is done. The main advantage being that changes to the repository do not get carry over to subsequent tests.


flxo (commented):
@alfonsoros88 Can you elaborate on  this feature? I don't get the point what it does and why it's needed

flxo (approved):

PR #270, opened 2021-01-29
Wait for `SIGKILL`ed processes

alfonsoros88:
Immediately after sending a `SIGKILL` to a process, the runtime proceeds destroy the process `cgroups`. If the process has not stopped at that point, the deletion of the `cgroups` causes a panic.

This change fixes that by waiting for the process that was `SIGKILL`ed.

### DIFF
--- a/northstar/src/runtime/minijail.rs
+++ b/northstar/src/runtime/minijail.rs
@@ -406,20 +406,19 @@ impl Process {
         signal::kill(unistd::Pid::from_raw(self.pid as i32), Some(sigterm))
             .map_err(|e| Error::Os(format!("Failed to SIGTERM {}", self.pid), e))?;
 
-        let timeout = Box::pin(time::sleep(timeout));
-        let exited = Box::pin(self.exit_handle_wait.recv());
-
-        let pid = self.pid;
-        Ok(select! {
-            s = exited => {
-                s.expect("Internal channel error during process termination")  // This is the happy path...
-            },
-            _ = timeout => {
-                signal::kill(unistd::Pid::from_raw(pid as i32), Some(signal::Signal::SIGKILL))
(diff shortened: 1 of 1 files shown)


northstar/src/runtime/minijail.rs
```diff
@@ -404,20 +404,19 @@ impl Process {
...
-                ExitStatus::Signaled(signal::Signal::SIGKILL)
+        let s = match tokio::time::timeout(timeout, self.exit_handle_wait.recv()).await {
+            Err(_) => {
+                // Send SIGKILL if the process did not terminate before timeout
+                signal::kill(
+                    unistd::Pid::from_raw(self.pid as i32),
+                    Some(signal::Signal::SIGKILL),
+                )
+                .map_err(|e| Error::Os("Failed to kill process".to_string(), e))?;
+                self.exit_handle_wait.recv().await
```
alfonsoros88, line 415: Is it ok to wait for a process that was `SIGKILL`ed? or can this potentially get stuck forever?

flxo: I think so. The `exit_handle` is fed from the `waitpid` task which unblocks regardless of what happens to the pid.
Do you have doubts?

alfonsoros88: I was not sure but @pclesr  did some checks to be sure.

flxo:
I think if the SIGKILL fails we have to admit that we're failed and panic. Otherwise we would leak a zombi. Waiting for the process to terminate on a SIGKILL is good but if it fails we should really `panic`.

alfonsoros88:
But how could the `SIGKILL` fail? assuming that the signal is properly sent to the process, it is guaranteed to be stopped correct?

flxo:
Merged. Thanks!

PR #271, opened 2021-02-01
Handle panics in test runtime

alfonsoros88:


northstar/src/main.rs
```diff
@@ -80,26 +80,26 @@ fn main() -> Result<(), Error> {
...
         _ = sigterm.recv() => {
             info!("Received SIGTERM. Stopping Northstar runtime");
-            runtime.stop_wait().await
+            runtime.stop()
         }
-        status = &mut runtime => status,
     };
-    match status {
+
+    match runtime.stop_wait().await {
```
flxo, line 102: Why is this change needed? A `stop_wait` does the same as a `stop` and `stop_wait.await`.

northstar_tests/src/runtime.rs
```diff
@@ -65,7 +65,16 @@ pub fn timeout_on<R>(f: impl Future<Output = R>) -> Result<R> {
 }
 
 /// A running instance of northstar.
-pub struct Runtime(runtime::Runtime);
+pub struct Runtime(Option<runtime::Runtime>);
+
+impl Drop for Runtime {
```
flxo, line 70: drop takes a &mut self. Why is this needed to be an `Option`? Because you want to move it into the `block_on`?

northstar/src/runtime/mod.rs
```diff
@@ -94,7 +89,7 @@ pub struct Runtime {
 
 impl Runtime {
     pub async fn start(config: Config) -> Result<Runtime, Error> {
-        let (stop_tx, stop_rx) = oneshot::channel();
+        let token = CancellationToken::new();
```
flxo, line 92: Nice.

northstar_tests/src/runtime.rs
alfonsoros88: Yes, exactly this :arrow_up:

northstar/src/main.rs
alfonsoros88: Hmmm yes, I think this is wrong :thinking:

flxo (approved):

flxo:
@alfonsoros88 Let's talk about `Runtime::stop` and `Runtime::wait`. I'd merge them again. because stopping without waiting for a result doesn't make sense and ignore the result is always an option.

northstar/src/runtime/mod.rs
```diff
@@ -118,23 +121,20 @@ impl Runtime {
...
+            token,
+        ))
     }
 
-    /// Stop the runtime and wait for the termination
-    pub fn stop_wait(mut self) -> impl Future<Output = RuntimeResult> {
-        self.stop.take();
-        self
+    /// Wait for the termination
+    pub async fn wait(self) -> RuntimeResult {
```
flxo, line 134: Should me merged into `Runtime::stop`

northstar/src/runtime/mod.rs
```diff
@@ -93,8 +86,8 @@ pub struct Runtime {
 }
 
 impl Runtime {
-    pub async fn start(config: Config) -> Result<Runtime, Error> {
-        let (stop_tx, stop_rx) = oneshot::channel();
+    pub async fn start(config: Config) -> Result<(Runtime, CancellationToken), Error> {
```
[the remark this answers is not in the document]
alfonsoros88: One problem with `stop_wait` is that the runtime has to be moved to the tasks that handles the `SIGINT` and `SIGTERM`. In general, I think most of the time we just want to signal the shutdown without waiting for it. Separating the token from the runtime allows that. An other alternative that would be to send a `Request::Shutdown` instead of using the token.

northstar/src/runtime/mod.rs
alfonsoros88: hmm I don't understand why `wait` has to be combined with `stop`.  Sometimes we want to stop without wait and sometimes we want to wait without stop.

northstar/src/runtime/mod.rs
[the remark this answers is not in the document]
flxo: I think there's no need for a separate task at all as it's done in the current implementation. That a `Request::Shutdown` is sent internally is a impl detail and should not be visible on the pub API.

northstar/src/runtime/mod.rs
flxo: What's the use case for stopping without waiting for a proper shutdown?

alfonsoros88: For the signal handling for example. They just trigger the stop without waiting.

flxo: Today there's a wait ;-). I just don't get the point why this is bad.

alfonsoros88: hmmm I didn't understand that :confused: I don't understand why this is bad either

northstar/src/runtime/mod.rs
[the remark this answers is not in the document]
alfonsoros88: Well separating these two things avoids to implement `Future` for `Runtime` and allows other type of "events" to trigger a shutdown if we so require in the future. I think this makes the handling of the runtime simpler and cleaner. Why is it better the way it was before?.

northstar/src/runtime/mod.rs
alfonsoros88: Ah got it, yeah the wait is done by someone else. The one owning the runtime is the one waiting for it. That's related to the other discussion of course.

PR #281, opened 2021-02-02
Otimize NPK usage

flxo:
There's just one NPK instanced needed during mount.


flxo:
I'm not sure if passing a fd opened non block to `losetup` is fine.

bzld (approved):
Looks great

bzld:
> I'm not sure if passing a fd opened non block to `losetup` is fine.

I didn't test that either, but I'd hope the kernel just cares about the fd, no matter which mode...

PR #282, opened 2021-02-02
Cache manifest and version in NPK

flxo:
Instead of parsing the manifest and version on each access, store a copy inside `Npk`.

### DIFF
--- a/northstar/src/runtime/loopdev.rs
+++ b/northstar/src/runtime/loopdev.rs
@@ -113,14 +113,13 @@ impl LoopDevice {
 
     pub fn attach_file(
         &self,
-        bf: &mut fs::File,
+        file_fd: RawFd,
         offset: u64,
         sizelimit: u64,
         read_only: bool,
         auto_clear: bool,
     ) -> Result<(), Error> {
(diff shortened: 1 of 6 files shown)


bzld:
I think we should also be loading most things accessable via pub accessors into RAM initially, so especially `hashes` & `fsimg_offset` & `fsimg_size` & `verity_header`.

Generally, the nicer way would be to do all this directly in `Npk::new`:
- Open ZIP
- Parse ZIP comment to get version (we may need to have different parsing logic depending on version in future)
- Read hashes file
- Check signature of hashes file directly? I think that's pretty fast anyway, and prevents using the data unverified by accidentally forgetting to call `verify` before reading? Maybe optional by having a if `Option<Key>` as added parameter?
- Parse hashes/verity data, keep in memory
- Parse manifest file, keep in memory
- Read fs.img metadata (location, size), keep in memory

flxo:
`Check signature of hashes file directly? I think that's pretty fast anyway, and prevents using the data unverified by accidentally forgetting to call verify before reading? Maybe optional by having a if Option<Key> as added parameter?` This totally makes sense.

inorick (approved):

PR #285, opened 2021-02-03
Generate docker style hello world container during build

marcmo:
The container is created at build time and subsequently
added to the application as a binary blob.
When executing northstar, a repository with only the
hello-world container is created and will be available to
test the most basic container.


northstar/Cargo.toml
```diff
@@ -92,3 +92,6 @@ runtime = [
 
 [build-dependencies]
 vergen = "3.1.0"
+tempfile = "3.2.0"
+npk = { path = "../npk" }
+tokio = { version = "1.1.1", features = ["rt"], optional = true }
```
flxo, line 97: Does this need `rt`?
What does the `optional` flag do in `build-dependencies`?

Cargo.toml
```diff
@@ -1,5 +1,4 @@
 [workspace]
-
 members = [
```
flxo, line 2: This should include `hello_world`

northstar/hello_world/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "hello_world"
+version = "0.1.0"
+authors = ["ESRLabs"]
+edition = "2018"
+
+[workspace]
```
flxo, line 7: We should get rid of one `hello`. There's examples/containers/hello` which is the very same as this here.

northstar/src/runtime/state.rs
```diff
@@ -135,10 +136,41 @@ impl fmt::Display for Application {
...
+    Ok(())
+}
+
 impl State {
     /// Create a new empty State instance
     pub(super) async fn new(config: &Config, tx: EventTx) -> Result<State, Error> {
         let mut repositories = HashMap::new();
+        if prepare_hello_world_repo(&mut repositories).await.is_err() {
+            warn!("Could not prepare hello-world repository");
+        }
```
flxo, line 173: This should be limited to debug builds.

npk/src/npk.rs
```diff
@@ -453,15 +453,15 @@ pub async fn pack(dir: &Path, out: &Path, key: Option<&Path>) -> Result<(), Erro
...
+    key: Option<PathBuf>,
     squashfs_opts: SquashfsOpts,
 ) -> Result<(), Error> {
     let manifest = read_manifest(dir).await?;
     let name = manifest.name.clone();
     let version = manifest.version.clone();
     let mut builder = Builder::new(dir, manifest);
     if let Some(key) = key {
-        builder = builder.key(key);
+        builder = builder.key(&key);
```
flxo, line 464: The key is passed up to here a s `Option<PathBuf>` just to reference it. This doesn't make sense.

What's wrong with `key: Option<&Path>`?

northstar/src/runtime/state.rs
```diff
@@ -135,10 +136,41 @@ impl fmt::Display for Application {
     }
 }
 
+async fn prepare_hello_world_repo(
```
flxo, line 139: Hide this for non debug builds.

northstar/hello_world/manifest.yaml
```diff
@@ -0,0 +1,12 @@
...
+init: /hello_world
+uid: 1000
+gid: 1000
+mounts:
+    /lib:
+      host: /lib
+    /lib64:
+      host: /lib64
+    /system:
+      host: /system
```
flxo, line 12: yaml indentation is usually 2 spaces. Might be wrong in the yaml we have..

northstar/build.rs
```diff
@@ -12,12 +12,75 @@
...
+        comp: None,
+        block_size: None,
+    };
+    runtime
+        .block_on(npk::npk::pack_with(
+            &pack_tmp_dir.path(),
+            &out_dir_path,
+            None, // without a key
+            squashfs_opts,
+        ))
```
flxo, line 84: Ah shit. That's the drawback....

northstar/build.rs
```diff
@@ -12,12 +12,75 @@
...
+    }
+    fs::copy(hello_exe_path, root_dir.join("hello_world"))
+        .expect("Could not copy hello_world file");
+    fs::copy(
+        hello_example_manifest,
+        pack_tmp_dir.path().join("manifest.yaml"),
+    )
+    .expect("Could not copy hello_world manifest file");
+
+    let runtime = tokio::runtime::Builder::new_multi_thread()
```
flxo, line 68: Is multithread needed?

northstar/Cargo.toml
marcmo: what I need is is a runtime to execute a task..is there a less intrusive way?
 the optional is wrong

northstar/build.rs
```diff
@@ -12,12 +12,75 @@
...
+pub fn package_hello_example() {
+    use std::{fs, path::Path, process::Command};
+    use tempfile::tempdir;
+
+    let out_dir_var = env::var_os("OUT_DIR").unwrap();
+    let out_dir_path = Path::new(&out_dir_var);
+    let out_dir_string = out_dir_var.to_string_lossy();
+
+    let args = ["build", "--release", "--target-dir", &out_dir_string];
+    let mut cmd = Command::new("cargo");
```
alfonsoros88, line 41: In the integration tests we use `escargot` to build the `test_container`. Could it be maybe useful here?

northstar/build.rs
marcmo: nope...will remove

northstar/build.rs
marcmo: never heard about that before ;) but using cargo here seems pretty straightforward to me

northstar/build.rs
marcmo: hmmm...with `new_current_thread` this does not work: 
thread 'main' panicked at 'can call blocking only when running on the multi-threaded runtime'

PR #287, opened 2021-02-04
Remove unshare fs type mount argument

flxo:
The fs type is not needed when updating the mount propagation.
Remove it from the config and replace with a None/NULL.


flxo:
@bzld Checked this after your hint. Thanks.

PR #289, opened 2021-02-05
Support unix socket in northstar console

alfonsoros88:
The `console_address` option in `northstar.tom` is replaced by `console` of type `Option<String>`. If the parameter is not set (`None`) the northstar console is not opened. The input string is now an `url` that specified a `tcp` connection e.g. `tcp://localhost:4200` or a `unix` socket e.g. `unix:path/to/socket`.

### DIFF
--- a/northstar/src/runtime/console.rs
+++ b/northstar/src/runtime/console.rs
@@ -17,6 +17,7 @@ use crate::{
     api::{self},
     runtime::{state::State, EventTx},
 };
+use async_trait::async_trait;
 use futures::{sink::SinkExt, StreamExt};
 use log::{debug, error, info, warn};
 use std::{collections::HashMap, path::PathBuf};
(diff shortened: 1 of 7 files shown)


northstar/src/runtime/console.rs
```diff
@@ -54,20 +56,112 @@ pub(crate) struct Console {
...
+    async fn accept_connection(&self) -> io::Result<Box<dyn ClientStream>> {
+        Ok(Box::new(self.accept().await?.0))
+    }
+}
+
+/// Creates either a Unix or Tcp socket as source of connections
+async fn create_listener(url: &Url) -> Result<Box<dyn Listener>, Error> {
+    match url.scheme() {
+        "tcp" => {
+            let address = format!("{}:{}", url.host().unwrap(), url.port().unwrap_or(4200));
```
marcmo, line 104: instead of `unwrap` I'd use a `ok_or_else` to map a parsing error to `Error`

marcmo:
@flxo what's the rational that we need to support unix sockets for our remote API?
Security and access control will need to be addressed anyway.
This complicates the code quite a bit IMOH

flxo:
@alfonsoros88 I rebased and refactored some things -> merged. Thanks!

PR #290, opened 2021-02-05
Simplification of builtin container

flxo:
The cargo build run from the build.rs is weird. The cross compilation
use case and the race around the build dir from the parallel running
cargo instances.
Replace the binary with an `echo` invocation. Dump the manifest during
build time instead of maintaining it in a file.

The tempdir which is the internal repository must be kept in `State`,
otherwise it get's removed.


northstar/src/runtime/state.rs
```diff
@@ -206,6 +208,8 @@ impl State {
             config: config.clone(),
             minijail,
             mount_control: MountControl::new(&config).await.map_err(Error::Mount)?,
+            #[cfg(debug_assertions)]
+            internal_repository,
```
marcmo, line 212: why do we need to keep this as a field in the struct?

flxo: Because the directory is removed when the enclosing `Tempdir` is dropped: https://docs.rs/tempfile/3.2.0/tempfile/struct.TempDir.html

PR #296, opened 2021-02-16
Add --manifest and --root options to sextant

inorick:
`sextant` no longer expects a `manifest.yaml` and `root` directory inside the provided container directory. Instead, both manifest and root can now be specified separately using the `--manifest` and `--root` CLI parameters. The old `--dir` option has been removed.

### DIFF
--- a/examples/build_examples.sh
+++ b/examples/build_examples.sh
@@ -122,6 +122,7 @@ build_example() {
   echo "${bold}Building ${NAME}${normal} (target: ${PLATFORM})"
 
   local ROOT_DIR="${TMP_DIR}/root"
+  local MANIFEST_DIR="${TMP_DIR}/manifest.yaml"
   exe rm -rf "${ROOT_DIR}"
   exe mkdir -p "${ROOT_DIR}"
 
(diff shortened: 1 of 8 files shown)


examples/build_examples.sh
```diff
@@ -122,6 +122,7 @@ build_example() {
   echo "${bold}Building ${NAME}${normal} (target: ${PLATFORM})"
 
   local ROOT_DIR="${TMP_DIR}/root"
+  local MANIFEST_DIR="${TMP_DIR}/manifest.yaml"
```
flxo, line 125: This var is not a dir but more a file. Rename it to `MANIFEST`

northstar/build.rs
```diff
@@ -73,14 +73,15 @@ mounts:
         "android" => MANIFEST_ANDROID,
         _ => MANIFEST,
     };
-    std::fs::write(out_dir.join("manifest.yaml"), &manifest)
-        .context("Failed to create manifest")?;
+    let manifest_dir = out_dir.join("manifest.yaml");
```
flxo, line 76: Rename `manifest_dir` to `manifest_file`

npk/src/npk.rs
```diff
@@ -375,7 +375,7 @@ impl Builder {
             context: "Failed to create temporary directory".to_string(),
             error: e,
         })?;
-        let tmp_root = copy_src_root_to_tmp(&self.root, &tmp).await?;
+        let tmp_root = copy_root_dir_to_tmp(&self.root, &tmp).await?;
```
flxo, line 378: The copy should not be necessary if the manifest is at it's own location. Probably the whole temper is not needed anymore since the root can be packed directly from where it lives.

flxo (changes requested):

npk/src/npk.rs
inorick: The copy is now be avoided. We still need the tmp dir to write the `fs.img` squashfsh to.

flxo (approved):

PR #298, opened 2021-02-17
Replace repo overlay with temporal repo

alfonsoros88:


northstar_tests/src/runtime.rs
```diff
@@ -75,36 +79,42 @@ impl Runtime {
...
-            .repositories
-            .get_mut("default")
-            .ok_or_else(|| eyre!("No \"default\" repository in configuration"))?
-            .dir;
-
-        // Discard changes to the default repository at the end of the test
-        *default_repository =
-            tokio::task::block_in_place(|| overlay_directory(default_repository))?;
+        // Test scoped repository
+        let repository = TempDir::new()?.into_path();
```
flxo, line 83: This persists the TempDir on disc and it will never be deleted:
https://docs.rs/tempfile/3.2.0/tempfile/struct.TempDir.html#method.into_path

northstar_tests/src/runtime.rs
```diff
@@ -65,7 +64,12 @@ pub fn timeout_on<R>(f: impl Future<Output = R>) -> Result<R> {
 }
 
 /// A running instance of northstar.
-pub struct Runtime(runtime::Runtime);
+pub struct Runtime {
+    runtime: runtime::Runtime,
+    /// Temporal repository for install/uninstall contaienrs in tests.
+    /// It is removed when shutting down the runtime.
+    repository: std::path::PathBuf,
```
flxo, line 71: Should be `path::PathBuf`

flxo (changes requested):

northstar_tests/src/runtime.rs
alfonsoros88: it is deleted in the shutdown

flxo: What happens with a panic?

A race in the Drop of Runtime sounds suspicious. The `Drop` for `TempDir` does: https://docs.rs/tempfile/3.2.0/src/tempfile/dir.rs.html#393-400.

The TempDir should be also free from opened fds.

alfonsoros88: It should yes, but it fails unfortunately. My guess would be that the fs is not immediately synced after the drop but I'm not sure.

flxo: Did you check which fd make the leak test fail? This would be interesting.

alfonsoros88: It is definitely related to the test directory, can't 100% be sure because it's just a number. I will try to have an other look at this.

flxo: > It is definitely related to the test directory, can't 100% be sure because it's just a number. I will try to have an other look at this.

Add a long sleep before the `Drop` of the Tmpdir and do a `ls -la` in /proc/<PID>/fd` with and without the tempdir. This should point you to right thing.

northstar/src/runtime/minijail.rs
```diff
@@ -509,84 +521,93 @@ impl Debug {
...
+                    Ok(n) => Ok(n),
+                    // read(2) on a nonblocking file (O_NONBLOCK) returns EAGAIN or EWOULDBLOCK in
+                    // case that the read would block. That case is handled by `try_io`.
+                    Err(nix::Error::Sys(e)) => Err(std::io::Error::from_raw_os_error(e as i32)),
+                    Err(e) => Err(std::io::Error::new(
+                        std::io::ErrorKind::Other,
+                        format!("{}", e),
+                    )),
+                }
+            }) {
```
flxo, line 575: Whats the advantage compared to the current version?

northstar/src/runtime/minijail.rs
```diff
@@ -685,6 +706,7 @@ impl ProcessSync {
         // is owned by the parent.
         let (childfd, parentfd) =
             pipe().map_err(|e| Error::Os("Failed to create pipe".to_string(), e))?;
+        error!("opening pipe ({}, {})", childfd, parentfd);
```
flxo, line 709: Logs start with a capital.

Is this an `error!`?

northstar_tests/tests/integration_tests.rs
```diff
@@ -267,28 +269,24 @@ test!(mount_containers_without_verity, {
...
-            .unwrap();
-        while let Ok(Some(e)) = dir.next_entry().await {
-            fds.insert(e.path());
-        }
-        Ok(fds)
+    fn fds() -> Result<HashSet<PathBuf>> {
+        Ok(std::fs::read_dir("/proc/self/fd")?
+            .filter_map(Result::ok)
+            .map(|entry| entry.path())
+            .collect())
```
flxo, line 276: nice!

northstar/src/runtime/minijail.rs
```diff
@@ -509,84 +521,93 @@ impl Debug {
...
-        let non_blocking =
-            AsyncFd::new(non_blocking).map_err(|e| Error::Io("Async fd".to_string(), e))?;
-
-        Ok(AsyncPipe {
-            non_blocking,
-            blocking,
-        })
-    }
+#[derive(Debug)]
+struct PipeFd(RawFd);
```
flxo, line 534: From the semantic point of view this is a `RawFd` that is closed on drop. It doesn't have any relation to pipes at all.

Can we do the close explicitly and remove `PipeFd`? This is just the case in `shutdown`.

northstar/src/runtime/minijail.rs
```diff
@@ -509,84 +521,93 @@ impl Debug {
     }
 }
 
-/// Pipe with async tokio impls
-#[derive(Debug)]
-struct AsyncPipe {
-    non_blocking: AsyncFd<std::fs::File>,
-    blocking: RawFd,
+/// Creates a pipe with the read endpoint implementing tokio::io::AsyncRead
+async fn create_pipe() -> Result<(PipeRead, PipeWrite), Error> {
```
flxo, line 525: Instead of a `create_pipe` there should be a `struct Pipe` that has a `fn split(self) -> (PipeLeft, PipeRight)`. 

Maybe we should move the pipe wrapper to a dedicated mod mit tests.

northstar/src/runtime/minijail.rs
alfonsoros88: whoops this was for debugging

northstar/src/runtime/minijail.rs
alfonsoros88: I think this version is a bit more transparent when handling the pipe fd. Mainly trying to avoid any surprise when delegating this to `std::fs::File`'s implementation.

northstar/src/runtime/minijail.rs
alfonsoros88: It is a new type rather than an alias, it holds one of the fds of the pipe, therefore the name. I wanted some type to manage the lifetime of the fd to be sure the pipe would always be closed.

We can call it differently if you want (maybe `FdGuard` or something) but manually closing the fds feels like it can potentially cause similar problems that the ones we had :(

northstar/src/runtime/minijail.rs
alfonsoros88: I also think it would be a good idea to make a mod out of this functionality. Even better would be to get rid of it entirely :sweat_smile:

flxo (commented):
Extracting the pipe creation stuff into a mod without making it generically usable doesn't make sense - then we can leave it where it is.
Here's a proposal how this could look like:

```rust
pub enum Error {
    // ...
}

pub struct Pipe {
    // ...
}

impl std::io::Read for Pipe {
    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
        todo!()
    }
}

impl std::io::Write for Pipe {
    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
        todo!()
    }

    fn flush(&mut self) -> io::Result<()> {
        todo!()
    }
}

impl Pipe {
    pub fn new() -> Result<Pipe, Error> {
        todo!()
    }

    pub fn split(self) -> (Endpoint, Endpoint) {
        todo!()
    }
}

pub struct Endpoint {
    file: std::fs::File,
}

impl AsRawFd for Endpoint {
    fn as_raw_fd(&self) -> RawFd {
        self.file.as_raw_fd()
    }
}

impl std::io::Write for Endpoint {
    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
        self.file.write(buf)
    }

    fn flush(&mut self) -> io::Result<()> {
        self.file.flush()
    }
}

impl std::io::Read for Endpoint {
    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
        self.file.read(buf)
    }
}

impl TryFrom<Endpoint> for AsyncEndpoint {
}

pub struct AsyncEndpoint {
}

impl AsyncRead for AsyncEndpoint {
    fn poll_read(
        self: Pin<&mut Self>,
        cx: &mut Context<'_>,
        buf: &mut ReadBuf<'_>,
    ) -> Poll<io::Result<()>> {
        todo!()
    }
}

impl AsyncWrite for AsyncEndpoint {
    fn poll_write(
        self: Pin<&mut Self>,
        cx: &mut Context<'_>,
        buf: &[u8],
    ) -> Poll<Result<usize, io::Error>> {
        todo!()
    }

    fn poll_flush(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), io::Error>> {
        todo!()
    }

    fn poll_shutdown(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Result<(), io::Error>> {
        todo!()
    }
}
```

alfonsoros88:
What would be the non "generically usable" part?

northstar/src/runtime/minijail.rs
```diff
@@ -520,6 +531,7 @@ impl AsyncPipe {
     fn new() -> Result<AsyncPipe, Error> {
         let (non_blocking, blocking) =
             pipe().map_err(|e| Error::Os("Failed to create pipe".to_string(), e))?;
+        error!("opening pipe ({}, {})", non_blocking, blocking);
```
flxo, line 534: Thinks this one is still left.

flxo (changes requested):

PR #299, opened 2021-02-17
Fix nstar issues

alfonsoros88:
This pr addresses the following issues with `nstar`:
- Re-introduces the reconnection loop.
- Fixes some issues with the handling of stdin/stdout that produced some weird artifacts.
- Validates the user input with `clap`
- Add missing help messages for commands


nstar/src/main.rs
```diff
@@ -55,10 +55,21 @@ async fn send_request(
...
+const RECONNECT_RETRIES: u64 = 5;
+
+/// Tries to connect to northstar `RECONNECT_RETRIES` times with 1 sec intervals before failing
+async fn try_connect<W: Write>(output: &mut W, url: &url::Url) -> Result<api::client::Client> {
+    for i in 0..RECONNECT_RETRIES {
+        match api::client::Client::new(url).await {
+            Ok(client) => return Ok(client),
+            Err(e) => {
+                time::sleep(time::Duration::from_secs(1)).await;
+                writeln!(output, "{}", e.to_string().red())?;
```
flxo, line 67: I'd flip the lines and extend the message with a message:

```rust
 writeln!(output, "Failed to connect to {}: {}", url, e.to_string().red())?;
 time::sleep(time::Duration::from_secs(1)).await;
```

nstar/src/main.rs
```diff
@@ -55,10 +55,21 @@ async fn send_request(
...
 
-async fn create_client(host: &str) -> Result<api::client::Client> {
-    api::client::Client::new(&url::Url::parse(host)?)
-        .await
-        .context(format!("Failed to stablish connection to {}", host))
+const RECONNECT_RETRIES: u64 = 5;
+
+/// Tries to connect to northstar `RECONNECT_RETRIES` times with 1 sec intervals before failing
+async fn try_connect<W: Write>(output: &mut W, url: &url::Url) -> Result<api::client::Client> {
+    for i in 0..RECONNECT_RETRIES {
```
flxo, line 62: What happens after `RECONNECT_RETRIES`? Wouldn't it make sense to retry forever in interactive mode? 

During development it's convenient to have a console funning forever that connects whenever the rt is reachable instead of having it to restart every while and then.

nstar/src/main.rs
```diff
@@ -71,61 +82,73 @@ async fn print_response<W: Write>(mut out: &mut W, response: Response, json: boo
...
+            client = try_connect(&mut std::io::stdout(), &url).await?;
         }
+        Ok(())
+    });
 
-        writeln!(terminal, "Disconnected")?;
-        time::sleep(time::Duration::from_secs(1)).await;
+    // Shutdown the runtime
+    // Also stops the terminal blocked reading from stdin
+    runtime.shutdown_background();
```
flxo, line 151: Why is this needed when exiting anyways?

alfonsoros88: This is needed because the terminal tasks that reads from `stdin` is blocked waiting for input at this point.

nstar/src/main.rs
alfonsoros88: We can make it infinite, no problem. I just thought that maybe a limit would make sense.

nstar/src/main.rs
```diff
@@ -71,61 +80,73 @@ async fn print_response<W: Write>(mut out: &mut W, response: Response, json: boo
...
+                            break;
+                        }
+                    }
+                    input = input.next() => {
+                        let input: &str = if input.is_some() {
+                            input.as_deref().unwrap()
+                        } else {
+                            break 'outer;
+                        };
+                        match parse_prompt(input) {
```
flxo, line 118: This is easier to read with

```rust
                        let input = if let Some(input) = input {
                            input
                        } else {
                            break 'outer;
                        };
                        match parse_prompt(&input) {
```

nstar/src/main.rs
flxo: The documentation os `shutdown_backgroud` explains that the tasks are just dropped and leaked. Feels a little bit hacky. Is there a way to unblock the terminal task and terminate gracefully?

nstar/src/main.rs
```diff
@@ -55,10 +55,19 @@ async fn send_request(
...
-async fn create_client(host: &str) -> Result<api::client::Client> {
-    api::client::Client::new(&url::Url::parse(host)?)
-        .await
-        .context(format!("Failed to stablish connection to {}", host))
+/// Tries to connect to northstar indefinitely
+async fn try_connect<W: Write>(output: &mut W, url: &url::Url) -> Result<api::client::Client> {
+    loop {
+        match api::client::Client::new(url)
+            .await
+            .context("Failed to Instantiate client")
```
flxo, line 63: That a client is instantiated is a impl details that's not interesting for the user. I'd change this text to "Failed to connect to {}"

nstar/src/main.rs
```diff
@@ -55,10 +55,19 @@ async fn send_request(
...
+        {
+            Err(e) => writeln!(output, "{}", e.to_string().red())?,
+            client => break client,
+        }
+        time::sleep(time::Duration::from_secs(1)).await;
+        writeln!(output, "Retrying for connection")?;
```
flxo, line 69: "Retrying for connection" feels wrong. "Reconnecting..."?

flxo (changes requested):
When the rt exits the output is broken:

nstar/src/main.rs
alfonsoros88: I tried a couple of approaches and nothing seems to stop the blocked task. This is the only thing that "worked" so far.

nstar/src/main.rs
alfonsoros88: Replaced it with:
```rust
let input = input.ok_or_else(|| anyhow!("Failed to read input"))?;
```
It's even nicer ;)

flxo:
@alfonsoros88 Just tested this: When `nstar` is connected and the runtime disconnects, `nstar` doesn't try to reconnect again as we had it before the changes. This is handy when developing northstar or  containers: just keep a `nstar` instance running that automatically reconnects whenever a rt instance is available.

A EOF (ctrl-d) doesn't completely shutdown nstar. This was also the case before the changes.

flxo:
The input prompt doesn't work as expected. During the connection attempts there shall be no prompt. In the current state of this PR a prompt is displayed in between the logs about the connection attempt. This can be solved by either dropping there terminal (and making sure the readline task breaks) whenever the nstar is disconnected, or starting the `readline` call in `Terminal` only on request for one input.

```
reconnect...
reconnect...
connected.
unblock readline task -> readline task blocks until eof or input and sends this
readline tasks blocks
...
disconnected...
reconnect...
reconnected...
connected.
unblock readline task etc..
```

alfonsoros88:
after spending a lot of time trying to find a workaround this, I'm not sure what you are asking is even possible with `rustyline` and the way `readline` is implemented. Once blocked on waiting for `stdin` there is not much you can do to interrupt that. I just had to check the "old" version of `nstar` to see how different was it respect to now but it does behave exactly in the same way. That said, I would propose to move forward as it is right now and hopefully we can find some replacement for `rustyline`.

northstar/src/runtime/minijail.rs
```diff
@@ -497,24 +497,26 @@ impl Debug {
...
             .debug
             .as_ref()
             .and_then(|debug| debug.strace.as_ref())
-        {
-            Some(Strace::new(strace, manifest, &config.log_dir, pid).await?)
-        } else {
-            None
-        };
+            .map(|strace| Strace::new(strace, manifest, &config.log_dir, pid))
+            .into();
```
marcmo, line 505: Didn't know about this OptionFuture...works but does not make the code more readable...or is the behavior changed in any way?

alfonsoros88: Nope, no difference at all. The reason for this change is to address a complain from clippy nightly that we should use `Option::map` here :man_shrugging:

nstar/src/main.rs
```diff
@@ -99,37 +114,39 @@ async fn main() -> Result<()> {
...
             }
         }
 
-        writeln!(terminal, "Disconnected")?;
-        time::sleep(time::Duration::from_secs(1)).await;
-
         // Try to reconnect
-        client = create_client(&opt.host).await?;
+        writeln!(terminal, "Disconnected, trying to reconnect")?;
+        client = try_connect(&mut terminal, &url).await?;
```
marcmo, line 150: while this is already very good, we could also do a less verbose reconnect by catching failed connects and just adding a `.` to the output

marcmo:
works much better :) ...I like it

nstar/src/main.rs
alfonsoros88: Oh yes, I like this idea :)

flxo:
Hej. Just check this PR. Something seems to be really strange and lock up. After typing some series of commands the prompt is completely messed or locked up somewhere.
Screenshot 2021-03-22 at 15 37 37
Screenshot 2021-03-22 at 15 37 51

Screenshot 2021-03-22 at 15 43 00

This happens when the client connected and disconnected from the rt a couple of times. Seems broken.

I think that `rustyline` is the root of all the problems.  Let's discuss the complete removal of rusty line and replace the convenience features with a nice shell completion script that runs `containers` internally to make stateful suggestions.

alfonsoros88:
@flxo it seems that I introduced that error by trying to implement the `Connecting...` message idea. But it seems it's not worth the trouble. I just reverted that change, it should work okay now.

flxo:
@alfonsoros88 If there's input while `nstar` is in it's reconnecting loop and it finally reconnects it locks up.

Steps to reproduce:

1. Start runtime
2. Start nstar
3. Stop runtime
4. Input something in nstar and hit return
5. Start the runtime

flxo:
Merging this. I'll add a issue to remove the interactive mode.

PR #303, opened 2021-02-24
Turn AsyncPipe into mod

alfonsoros88:
Turn AsyncPipe into mod

### DIFF
--- a/northstar/src/runtime/pipe.rs
+++ b/northstar/src/runtime/pipe.rs
@@ -1,0 +1,255 @@
+// Copyright (c) 2019 - 2020 ESRLabs
+//
+//   Licensed under the Apache License, Version 2.0 (the "License");
+//   you may not use this file except in compliance with the License.
+//   You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+//   Unless required by applicable law or agreed to in writing, software
+//   distributed under the License is distributed on an "AS IS" BASIS,
+//   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//   See the License for the specific language governing permissions and
+//   limitations under the License.
+
+use nix::{fcntl, fcntl::OFlag, unistd};
+use std::{
+    convert::TryFrom,
+    io,
+    os::unix::io::{AsRawFd, RawFd},
+    pin::Pin,
+    task::{Context, Poll},
+};
+use tokio::io::{unix::AsyncFd, AsyncRead, AsyncWrite, ReadBuf};
+
+/// Opens a pipe(2) with both ends blocking
+pub(crate) fn pipe() -> io::Result<(PipeReader, PipeWriter)> {
+    let (readfd, writefd) = unistd::pipe().map_err(from_nix)?;
+    Ok((PipeReader(readfd), PipeWriter(writefd)))
+}
+
+/// Read end of a pipe(2)
+#[derive(Debug)]
+pub(crate) struct PipeReader(RawFd);
+
+impl std::io::Read for PipeReader {
+    fn read(&mut self, buf: &mut [u8]) -> io::Result<usize> {
+        unistd::read(self.0, buf).map_err(from_nix)
+    }
+}
+
+impl AsRawFd for PipeReader {
+    fn as_raw_fd(&self) -> RawFd {
+        self.0
+    }
+}
+
+impl Drop for PipeReader {
+    fn drop(&mut self) {
+        unistd::close(self.0).unwrap();
+    }
+}
+
+/// Write end of a pipe(2)
+#[derive(Debug)]
+pub(crate) struct PipeWriter(RawFd);
+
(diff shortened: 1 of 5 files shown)


northstar/src/runtime/pipe.rs
```diff
@@ -0,0 +1,256 @@
...
+impl AsRawFd for PipeReader {
+    fn as_raw_fd(&self) -> RawFd {
+        self.0
+    }
+}
+
+impl Drop for PipeReader {
+    fn drop(&mut self) {
+        if let Err(e) = unistd::close(self.0) {
+            log::error!("Failed to close pipe: {}", e);
```
flxo, line 50: There's not much we can do when the close fails. I'd panic instead of just logging it. Retrying doesn't make sense.

northstar/src/runtime/pipe.rs
```diff
@@ -0,0 +1,256 @@
...
+
+/// Write end of a pipe(2)
+#[derive(Debug)]
+pub(crate) struct PipeWriter(RawFd);
+
+impl std::io::Write for PipeWriter {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        unistd::write(self.0, buf).map_err(from_nix)
+    }
+    fn flush(&mut self) -> io::Result<()> {
```
flxo, line 63: newline missing.

northstar/src/runtime/pipe.rs
```diff
@@ -0,0 +1,256 @@
...
+impl AsRawFd for PipeWriter {
+    fn as_raw_fd(&self) -> RawFd {
+        self.0
+    }
+}
+
+impl Drop for PipeWriter {
+    fn drop(&mut self) {
+        if let Err(e) = unistd::close(self.0) {
+            log::error!("Failed to close pipe: {}", e);
```
flxo, line 77: Same as in `PipeReader`

northstar/src/runtime/pipe.rs
```diff
@@ -0,0 +1,256 @@
...
+    }
+}
+
+/// Pipe's synchronous reading end
+#[derive(Debug)]
+pub(crate) struct AsyncPipeReader(AsyncFd<PipeReader>);
+
+impl TryFrom<PipeReader> for AsyncPipeReader {
+    type Error = io::Error;
+    fn try_from(reader: PipeReader) -> Result<Self, Self::Error> {
```
flxo, line 88: newline missing

northstar/src/runtime/pipe.rs
```diff
@@ -0,0 +1,256 @@
...
+    }
+}
+
+/// Pipe's synchronous writing end
+#[derive(Debug)]
+pub(crate) struct AsyncPipeWriter(AsyncFd<PipeWriter>);
+
+impl TryFrom<PipeWriter> for AsyncPipeWriter {
+    type Error = io::Error;
+    fn try_from(writer: PipeWriter) -> Result<Self, Self::Error> {
```
flxo, line 132: newline missing

northstar/src/runtime/pipe.rs
```diff
@@ -0,0 +1,256 @@
...
+    flags |= OFlag::O_NONBLOCK;
+    fcntl::fcntl(fd, fcntl::FcntlArg::F_SETFL(flags)).map_err(from_nix)?;
+    Ok(())
+}
+
+/// Maps an nix::Error to a io::Error
+fn from_nix(error: nix::Error) -> io::Error {
+    match error {
+        nix::Error::Sys(e) => io::Error::from_raw_os_error(e as i32),
+        e => io::Error::new(io::ErrorKind::Other, format!("{}", e)),
```
flxo, line 187: For `ErrorKind::Other` we can put any error into the `io::Error`. No need to convert it to a `String`.

` e => io::Error::new(io::ErrorKind::Other, e),`

northstar/src/runtime/pipe.rs
flxo, line 88: Think this drops `reader`. With the drop the `Drop` impl of `PipeReader` is executed which closes the socket? Can we check this?

If yes:
An option would be to either mark the `PipeReader` to *not* close or put the fd into an `Option` that is taken upon the `try_from` to make the `PipeReader` not close.

flxo (changes requested):

northstar/src/runtime/pipe.rs
alfonsoros88: I think `reader` is just moved here, there is no reason to drop it. Otherwise the integration tests would fail when checking for container output.

flxo: `reader` is moved and dropped at the end of `try_from` when it goes out of scope.

flxo: ah. got it - it's moved into AsyncPipeReader.. Sorry for the confusion.

flxo:
The buidld 0cf368d failed. Merging anyway with the added info in the leak test. Thanks @alfonsoros88

PR #305, opened 2021-03-01
Add NPK guide

inorick:


guide/src/sextant/pack.md
```diff
@@ -0,0 +1,100 @@
...
+
+```bash
+$ ls target/release/hello
+target/release/hello
+```
+
+Next, we can crate the destination directory if it does not already exist:
+
+```bash
+$ mkdir -p target/northstar/repository
```
alfonsoros88, line 67: Maybe it would be good to explain here why this is the chosen path or make some reference to what container repositories are and how containers are "installed"

alfonsoros88 (approved):
looks great! :+1: Maybe it would be nice to add a link in the main `README.md` to the gh page.

guide/src/sextant/inspect.md
```diff
@@ -0,0 +1,87 @@
...
+    host: /lib
+  /lib64:
+    host: /lib64
+  /system:
+    host: /system
+io:
+  stdout:
+    log:
+      level: DEBUG
+      tag: hello
```
marcmo, line 44: is it possible to include another file into this file? then we could include the manifest from our example and ensure we are always up to date

guide/src/sextant/npk_format_details.md
```diff
@@ -0,0 +1,239 @@
...
+cgroups:
+  memory:
+    limit_in_bytes: 10000000
+    swappiness: 0
+io:
+  stdout:
+    log:
+      - DEBUG
+      - memeater
+```
```
marcmo, line 38: probably better for that one since the first manifest does not even need to be completely visible in the first place

guide/src/sextant/pack.md
inorick: I added a small passage as to why we are choosing that directory. Not sure what to write about 'installing' though.

guide/src/sextant/inspect.md
```diff
@@ -0,0 +1,68 @@
...
+$ sextant inspect target/northstar/repository/hello-0.0.1.npk 
+# inspection of 'target/northstar/repository/hello-0.0.1.npk'
+## NPK Content
+signature.yaml
+manifest.yaml
+fs.img
+
+## manifest.yaml
+{{#include ./../../../examples/container/hello/manifest.yaml}}
+
```
marcmo, line 26: nice

PR #306, opened 2021-03-01
Lazy mounting and test refactoring

flxo:


northstar/src/runtime/state.rs
```diff
@@ -841,120 +686,27 @@ impl<'a> State<'a> {
...
-                .mount_control
-                .mount_npk(&entry.path(), repo, &self.config.run_dir)
-                .await
-                .map_err(Error::Mount)?;
-            containers.push(container);
-        }
+/// Dump the hello world npk created at compile time into a tmpdir that acts as internal
+/// repository.
+#[cfg(debug_assertions)]
+async fn prepare_internal_repository(name: &str) -> Result<(tempfile::TempDir, Repository), Error> {
```
marcmo, line 700: internal repository is not a good description...rather demo repository or s.th. like that

northstar/src/runtime/state.rs
```diff
@@ -109,615 +89,464 @@ impl ProcessContext {
...
+            .repositories
+            .get(key.repository())
+            .and_then(|r| r.key.as_ref());
+        let npk = self
+            .npks
+            .get(&key)
+            .ok_or_else(|| Error::UnknownApplication(key.clone()))?;
+
+        // Load NPK
+        let npk = Npk::from_path(npk, repository_key)
```
marcmo, line 157: this is really misleading...it's either an npk or an npk-path. too bad rust allows for reassigning to the same identifier.

northstar/src/runtime/state.rs
```diff
@@ -109,615 +89,464 @@ impl ProcessContext {
...
+            mounted.push(key.clone());
+            self.containers.insert(key.clone(), container);
+            manifest
+        } else {
+            warn!("Unknown application {}", key);
+            return Err(Error::UnknownApplication(key.clone()));
+        };
+
+        // Find to be mounted resources
+        for key in manifest.mounts.values().filter_map(|m| match m {
```
marcmo, line 277: Here we mount a container and then check the needed resources and need to roll-back when something is missing. Would be better to first check if all resources are available and can be mounted, then try to mount the container

northstar/src/runtime/state.rs
```diff
@@ -109,615 +89,464 @@ impl ProcessContext {
...
+                                self.umount(&k).await.ok();
+                            }
+                            return Err(e);
+                        }
+                    }
+                } else {
+                    for k in mounted.drain(..) {
+                        self.umount(&k).await.ok();
+                    }
+                    return Err(Error::UnknownResource(key));
```
marcmo, line 313: handling error could also be easier with a dedicated mount-required-resources function that can fail.
not so many loose ends to tie up

northstar/src/runtime/state.rs
```diff
@@ -109,615 +89,464 @@ impl ProcessContext {
...
                 }
             }
         }
 
-        // Spawn process
-        info!("Starting {}", app);
+        // This must exist
+        let container = self.containers.get(key).unwrap();
```
marcmo, line 319: and no ugly "let's unwrap since it was already checked" code

northstar/src/runtime/container_key.rs
```diff
@@ -0,0 +1,84 @@
...
+use super::{Name, RepositoryId, Version};
+use serde::{Deserialize, Serialize};
+use std::{
+    convert::TryFrom,
+    fmt::{self, Display},
+    sync::Arc,
+};
+
+#[derive(Clone, Eq, PartialEq, Debug, Hash, Serialize, Deserialize)]
+pub struct ContainerKey {
```
marcmo, line 24: would not call this `Key`. yes it's used as a key in a hashmap but conceptually it is still a container id

northstar/src/runtime/container_key.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[serde(flatten)]
+    inner: Arc<Inner>,
```
marcmo, line 26: not sure why you make this so complicated...no need to put this in an `Arc`.
A key/an id is also a value object and should be cloned not used as  reference

northstar/src/runtime/state.rs
flxo: It's in `debug` builds only so it's not important: `demo` or `example`?

northstar/src/runtime/state.rs
flxo: It's type safe and within a scope unique. I love this feature :-)

northstar/src/runtime/state.rs
flxo: That's a optimisation.
The mounts have to be paralleled in future anyway which is an open todo:
- assemble a list of npks to be mounted (including the "root" container)
- mount all of those containers in parallel (today it's sequential).

northstar/src/runtime/state.rs
flxo: The mounting code will be rewritten soon. See previous comment - so I didn't care much about it.
Agree that this is too noisy.

northstar/src/runtime/state.rs
flxo: This is the result of a borrow checker fight. I couldn't solve it. It because above a `Container` needs to be inserted into `self.containers` which mutably borrows.

Can you find a solution?

northstar/src/runtime/container_key.rs
flxo: The name key has been chosen because it's also used as the key for maps. Agree that `ContainerKey` is long and not ideal. My personal perference is `Container` and the current `Container` is renamed to `ContainerInstance`.

Thoughts?

northstar/src/runtime/container_key.rs
flxo: Right it should be cloned - that's why it's done this way to make the clone cheap. Instead of cloning two strings and a version just clone the `Arc`.

northstar/src/runtime/repository.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    pub fn npks(&self) -> &HashMap<(Name, Version), PathBuf> {
+        &self.containers
+    }
+
+    pub async fn add(&mut self, key: &ContainerKey, src: &Path) -> Result<(), Error> {
+        let dest = self
+            .dir
+            .join(format!("{}-{}.npk", key.name(), key.version()));
```
marcmo, line 60: adding a container to a repository should also update the `last_modified` timestamp

northstar/src/runtime/state.rs
marcmo: I know it is technically typesafe but still does not contribute to understandable code in some instances such as this. I would not use it here.

northstar/src/runtime/state.rs
marcmo: yes, parallelizing will be good, but still better in this way:
* assemble a list of npks to be mounted *including* necessary resources
* mount needed resource containers
* mount *only* those application containers that can be mounted
that will have the advantage that we do not do unnecessary work (roll-back)

northstar/src/runtime/container_key.rs
marcmo: not sure...what it it is a unique identifier for a container-instance, yes...
but then *-Instance is some what redundant since a lot of things are instances of some kind.
Just leave it as `ContainerKey` or `ContainerId` then

northstar/src/runtime/container_key.rs
marcmo: ok...still a typical example of POITROAE

flxo: I think it's worth to save the clones since it's anyway not visible to the user of `ContainerKey`.

A clone would clone 2 Strings, 3 u64 (yes that's cheap) and 2 `Vec`s that potentially contain `Strings`.

northstar/src/runtime/container_key.rs
flxo: So `ContainerId` and `Container`? I'd like to get rid of `key`.

northstar/src/runtime/repository.rs
flxo: Yes. Added.

northstar/src/runtime/repository.rs
```diff
@@ -14,44 +14,70 @@
...
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub(super) struct Repository {
     pub(super) id: RepositoryId,
     pub(super) dir: PathBuf,
     pub(super) key: Option<PublicKey>,
-    last_modified: Option<SystemTime>,
-    containers: HashMap<(Name, Version), PathBuf>,
+    pub(super) containers: HashMap<Container, PathBuf>,
+    pub(super) npks: HashMap<Container, Npk>,
```
marcmo, line 36: keeping `containers` and `npks` around seems redundant and probably should be consolidated

marcmo:
looks already pretty good. I'd suggest to merge it now since it involves a lot of changes and take it as a basis.

marcmo (changes requested):
please update the CHANGELOG

northstar/src/runtime/state.rs
```diff
@@ -109,617 +95,494 @@ impl ProcessContext {
...
     }
 
-    /// Return an iterator over all known resources
-    pub fn resources(&self) -> impl iter::Iterator<Item = &Container> {
-        self.resources.values()
+    /// Mount `container`
+    async fn mount(
+        &self,
+        container: &Container,
+    ) -> Result<impl Future<Output = Result<MountedContainer, Error>>, Error> {
```
alfonsoros88, line 159: Just as before I don't understand the use for the double future.

northstar_tests/src/logger.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) 2019 - 2020 ESRLabs
 //
 //   Licensed under the Apache License, Version 2.0 (the "License");
-//   you may not use this file except in compliance with the License.
+//   you may not use thpub(crate) is file except in compliance with the License.
```
alfonsoros88, line 4: whoops?

northstar_tests/test_container/src/main.rs
```diff
@@ -39,30 +40,36 @@ enum TestCommands {
         path: PathBuf,
     },
     Whoami,
-    Loop,
+    LeakMemory,
```
alfonsoros88, line 43: what happened to loop?

northstar_tests/tests/tests.rs
```diff
@@ -0,0 +1,323 @@
...
+
+    for client in &clients {
+        assert!(client.containers().await.is_err());
+    }
+    clients.clear();
+
+    result
+});
+
+// test!(cgroups_memory, {
```
alfonsoros88, line 309: is this test no longer relevant?

northstar_tests/test_container/manifest.yaml
```diff
@@ -3,6 +3,10 @@ version: 0.0.1
 init: /test_container
 uid: 1000
 gid: 1000
+# cgroups:
+#   memory:
+#     limit_in_bytes: 10000000
+#     swappiness: 0
```
alfonsoros88, line 9: what happened to these configs?

alfonsoros88 (commented):
Quite a long one :sweat:  maybe we can do short ones next time. Btw couldn't compile `nstar` with `cargo build -p nstar`. There are some details but i guess we can fix whatever issue shows up in the future.

northstar/src/api/client.rs
```diff
@@ -346,18 +370,64 @@ impl Client {
             _ => Err(Error::Protocol),
         }
     }
+
+    /// Mount a list of containers
+    /// TODO
```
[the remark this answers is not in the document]
flxo: The docs.

northstar/src/runtime/console.rs
```diff
@@ -47,8 +45,9 @@ pub(crate) enum Request {
 pub(crate) struct Console {
     event_tx: EventTx,
     url: Url,
-    notification_tx: broadcast::Sender<api::model::Notification>,
-    token: CancellationToken,
+    notification_tx: broadcast::Sender<Notification>,
+    stop: CancellationToken,
+    stopped: CancellationToken,
```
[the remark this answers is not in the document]
flxo: Jup. done.

northstar/src/runtime/console.rs
```diff
@@ -117,19 +124,38 @@ impl Console {
                                     }
                                 }
                             }
-                            _ = token.cancelled() => break,
+                            _ = stop.cancelled() => {
+                                debug!("Closing listener on {}", address);
+                                drop(listener);
+                                // TODO: Wait for all connections to be closed
```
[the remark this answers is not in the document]
flxo: Done! Thanks for the hint. Used a `FuturesUnordered` to keep that thing clear from dead connections by polling it regularly.

northstar/src/runtime/console.rs
```diff
@@ -61,13 +60,15 @@ pub enum Error {
...
+
         Ok(Self {
-            event_tx: tx.clone(),
-            url,
+            event_tx,
+            url: url.clone(),
             notification_tx,
-            token: CancellationToken::new(),
+            stop: CancellationToken::new(),
+            stopped: CancellationToken::new(),
```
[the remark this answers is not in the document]
flxo: Done.

northstar/src/runtime/state.rs
flxo: If `Npk::from_path` would not be async this would not be needed.

northstar_tests/test_container/src/main.rs
flxo: `Loop` was unused. It's all `Io` and `Os` (nix) erros.

northstar_tests/tests/tests.rs
flxo: Not at all!

cgroups mem doesn't trigger in the GH actions runs. Commented for now and needs to be taken care of.

northstar_tests/test_container/manifest.yaml
flxo: Commented along the the cgroup mem test. See above.

northstar/src/runtime/state.rs
flxo: Updated the patch with the following behaviour: Whenever a mount fails the container is kept mounted. In a worst case the rt would need to be abled to handle every present npk to be mounted so a umount is not needed upon start errors.

northstar/src/runtime/state.rs
flxo: Think we can do this in an follow up.

flxo: Created #307

northstar/src/runtime/state.rs
flxo: Done in `8835f8a`.

northstar/src/runtime/container_key.rs
flxo: Done in `8835f8a`

northstar/src/api/client.rs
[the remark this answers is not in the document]
flxo: Done.

northstar_tests/src/logger.rs
flxo: oops. Fixed.

northstar_tests/tests/tests.rs
flxo: Created #309

northstar_tests/test_container/manifest.yaml
flxo: Created #309

flxo:
> please update the CHANGELOG

Done.

PR #310, opened 2021-03-09
Remove unshare_root workaround

marcmo:
by introducing a bind-mount of the run-dir
and marking the mount propagation for that one as private

### DIFF
--- a/northstar/src/main.rs
+++ b/northstar/src/main.rs
@@ -59,15 +59,6 @@ fn main() -> Result<(), Error> {
         .map(|r| r.disable_mount_namespace)
         .unwrap_or(false)
     {
-        // Set the mount propagation of unshare_root to MS_PRIVATE
-        nix::mount::mount(
-            Option::<&'static [u8]>::None,
-            config.devices.unshare_root.as_os_str(),
-            Option::<&str>::None,
-            nix::mount::MsFlags::MS_PRIVATE,
-            Option::<&'static [u8]>::None,
-        )?;
-
         // Enter a mount namespace. This needs to be done before spawning
         // the tokio threadpool.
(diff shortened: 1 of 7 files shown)


marcmo:
Not sure that is all. Running on my machine the mount is not visible to other processes.
tests are now broken but first wanted to get some feedback from @flxo

flxo:
@marcmo The tests fail with:

```
The application panicked (crashed).
956
Message:  Test failed: 
957
   0: Failed to remove tmpdir
958
   1: Device or resource busy (os error 16) at path "/tmp/.tmpOnsgZJ"
```

This is because the bind mount is not umounted upon shutdown. This missing  when the runtime task breaks.

northstar/src/main.rs
```diff
@@ -59,15 +59,6 @@ fn main() -> Result<(), Error> {
         .map(|r| r.disable_mount_namespace)
         .unwrap_or(false)
     {
-        // Set the mount propagation of unshare_root to MS_PRIVATE
```
flxo, line 62: This needs to be remove from integration_tests::macros.rs as well

northstar_tests/src/runtime.rs
```diff
@@ -94,7 +94,6 @@ impl Northstar {
             cgroups,
             // This sections matches most common x86 Linux distributions
             devices: config::Devices {
-                unshare_root: PathBuf::from("/"),
```
flxo, line 97: The `unshare_root` is in the example northstar.toml. Think we should add `#[serde(deny_unknown_fields)]` to not allow unknown fields in all struct in `config.rs`.

flxo (changes requested):

PR #311, opened 2021-03-10
Fix dependency from api to runtime

alfonsoros88:
After the changes for lazy mounting, a dependency from api -> runtime
was introduced. The api models for ExitCode and Notification used the
homologue in runtime. Additionally the api model used the
runtime::continer module.

This pr solves this issue by:
- Inverting the dependency for notification and exitcode.
- Moves runtime::container module to api

### DIFF
--- a/northstar/src/runtime/mod.rs
+++ b/northstar/src/runtime/mod.rs
@@ -40,7 +40,6 @@ use tokio_util::sync::CancellationToken;
 mod cgroups;
 pub mod config;
 mod console;
-mod container;
 #[allow(unused)]
 mod device_mapper;
 mod error;
@@ -56,8 +55,7 @@ pub(self) mod state;
 
 pub(self) type EventTx = mpsc::Sender<Event>;
 pub(self) type RepositoryId = String;
-pub use container::*;
-pub(self) use npk::manifest::{Name, Version};
+pub use api::container::*;
 pub(self) use repository::Repository;
 pub(self) use state::MountedContainer;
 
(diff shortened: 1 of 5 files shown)


northstar/src/runtime/mod.rs
```diff
@@ -56,8 +55,7 @@ pub(self) mod state;
 
 pub(self) type EventTx = mpsc::Sender<Event>;
 pub(self) type RepositoryId = String;
-pub use container::*;
-pub(self) use npk::manifest::{Name, Version};
+pub use api::container::*;
```
flxo, line 58: Hm. I'm not sure whether this super central type `Container` should live in the `api` module. In general: there are some type that are used by `api`, `runtime` and even `npk`. 

Proposal: Create a `common` module that houses `Container`, `Notification` etc.. and is used by `runtime` `api` etc? This would avoid the need to have the `runtime` feature enabled when someone just uses `api` (means currently `nstar`?

What do you think?

flxo (commented):
Thanks for the finding! You probably hit this when installing nstar.

alfonsoros88:
@flxo actually I commented about this issue on the review yesterday :sweat_smile:

northstar/src/runtime/mod.rs
alfonsoros88: Currently `nstar`  uses only "api" from northstar. I don't see much trouble as it is currently. Maybe let us fix this first and then think about module organization later.

flxo: Ok. Fine for me! I'll create a ticket!

PR #315, opened 2021-03-12
Handle new connections during shutdown

alfonsoros88:
Currently it is possible to accept new connections during a shutdown.
This causes a sort of dead lock where the shutdown cannot terminate as
log as connections remain open but requests cannot be answer because the
runtime is just waiting for the shutdown.

To handle this situation, if a shutdown request is issued, the console
is closed for new connections and request in the event loop are
ignored.

### DIFF
--- a/northstar/src/runtime/console.rs
+++ b/northstar/src/runtime/console.rs
@@ -22,7 +22,7 @@ use futures::{
     future::join_all,
     sink::SinkExt,
     stream::{self, FuturesUnordered},
-    StreamExt,
+    Future, StreamExt, TryFutureExt,
 };
 use log::{debug, error, info, trace, warn};
 use std::{path::PathBuf, unreachable};
@@ -69,6 +69,14 @@ pub enum Error {
     Protocol(String),
     #[error("IO error: {0} ({1})")]
     Io(String, #[source] io::Error),
+    #[error("Event loop closed, request cannot be processed")]
+    EventLoopClosed,
+}
+
+macro_rules! io_err {
+    ($($msg:expr),*) => {
+        |e| Error::Io(format!($($msg,)*), e)
+    };
 }
 
 impl Console {
@@ -93,130 +101,18 @@ impl Console {
         // Stop token for self *and* the connections
         let stop = self.stop.clone();
 
-        match self.url.scheme() {
-            "tcp" => {
-                let addresses = self
-                    .url
-                    .socket_addrs(|| Some(4200))
-                    .map_err(|e| Error::Io("Invalid console address".into(), e))?;
-                let address = addresses
-                    .first()
-                    .ok_or_else(|| {
-                        Error::Io(
-                            "Invalid console url".into(),
-                            io::Error::new(io::ErrorKind::Other, ""),
-                        )
-                    })?
-                    .to_owned();
-
-                debug!("Starting console on {}", &address);
-
-                let listener = TcpListener::bind(&address).await.map_err(|e| {
-                    Error::Io(format!("Failed to open tcp listener on {}", &address), e)
-                })?;
-
-                debug!("Started console on {}", &address);
-
-                let task = task::spawn(async move {
-                    // Connection tasks
-                    let mut connections = FuturesUnordered::new();
-                    loop {
-                        select! {
-                            stream = listener.accept() => {
-                                match stream {
-                                    Ok(stream) => {
-                                        connections.push(task::spawn(Self::connection(
-                                            stream.0,
-                                            stream.1.to_string(),
-                                            stop.clone(),
-                                            event_tx.clone(),
-                                            notification_tx.subscribe(),
-                                        )));
-                                    }
-                                    Err(e) => {
-                                        warn!("Error listening: {}", e);
-                                        break;
-                                    }
-                                }
-                            }
-                            _ = connections.next(), if !connections.is_empty() => {},
-                            _ = stop.cancelled() => {
-                                drop(listener);
-                                debug!("Closed listener on {}", address);
-                                if ! connections.is_empty() {
-                                    debug!("Waiting for connections to be closed");
-                                    while connections.next().await.is_some() {}
-                                }
-                                break;
-                            }
-                        }
-                    }
-                });
-                self.tasks.push(task);
-            }
-            "unix" => {
-                let address = PathBuf::from(self.url.path());
-
-                debug!("Starting console on {}", address.display());
-
-                if address.exists() {
-                    fs::remove_file(&address)
-                        .await
-                        .map_err(|e| Error::Io("Failed to remove unix socket".into(), e))?;
-                }
-
-                let listener = UnixListener::bind(&address).map_err(|e| {
-                    Error::Io(
-                        format!("Failed to open unix listener on {}", address.display()),
-                        e,
-                    )
-                })?;
-
-                debug!("Started console on {}", address.display());
-
-                let task = task::spawn(async move {
-                    // Connection tasks
-                    let mut connections = FuturesUnordered::new();
-                    loop {
-                        select! {
-                            stream = listener.accept() => {
-                                match stream {
-                                    Ok(stream) => {
-                                        connections.push(task::spawn(Self::connection(
-                                            stream.0,
-                                            format!("{:?}", &stream.1),
-                                            stop.clone(),
-                                            event_tx.clone(),
-                                            notification_tx.subscribe(),
-                                        )));
-                                    }
-                                    Err(e) => {
-                                        warn!("Error listening: {}", e);
-                                        break;
-                                    }
-                                }
-                            }
-                            _ = connections.next(), if !connections.is_empty() => {},
-                            _ = stop.cancelled() => {
-                                drop(listener);
-                                debug!("Closed listener on {}", address.display());
-                                if address.exists() {
-                                    fs::remove_file(&address)
-                                        .await.expect("Failed to remove unix socket");
-                                }
-                                if ! connections.is_empty() {
-                                    debug!("Waiting for connections to be closed");
-                                    while connections.next().await.is_some() {}
-                                }
-                                break;
-                            }
-                        }
-                    }
-                });
-                self.tasks.push(task);
-            }
-            _ => unreachable!(),
-        }
+        let task = match Listener::new(&self.url)
+            .await
+            .map_err(|e| Error::Io("Failed to remove unix socket".into(), e))?
+        {
+            Listener::Tcp(listener) => task::spawn(async move {
+                handle_connections(|| listener.accept(), event_tx, notification_tx, stop).await
+            }),
+            Listener::Unix(listener) => task::spawn(async move {
+                handle_connections(|| listener.accept(), event_tx, notification_tx, stop).await
+            }),
+        };
+        self.tasks.push(task);
 
         Ok(())
     }
@@ -235,7 +131,7 @@ impl Console {
 
     async fn connection<T: AsyncRead + AsyncWrite + Unpin>(
         stream: T,
-        peer: String,
+        peer: ClientId,
         stop: CancellationToken,
         event_tx: EventTx,
         mut notification_rx: broadcast::Receiver<Notification>,
@@ -347,100 +243,20 @@ impl Console {
                     } else {
                         break;
                     };
-                    let message_id = message.id.clone();
 
                     trace!("{}: --> {:?}", peer, message);
-
-                    let mut keep_file = None;
-
-                    let request =
-                        if let api::model::Payload::Request(
-                            api::model::Request::Install(repository, size)) =
-                            message.payload
-                        {
-                            debug!(
-                                "{}: Received installation request with size {}",
-                                peer,
-                                bytesize::ByteSize::b(size)
-                            );
-                            info!("{}: Using repository \"{}\"", peer, repository);
-                            // Get a tmpfile name
-                            let tmpfile = match tempfile::NamedTempFile::new() {
-                                Ok(f) => f,
-                                Err(e) => {
-                                    warn!("Failed to create tempfile: {}", e);
-                                    break;
-                                }
-                            };
-
-                            // Create a tmpfile
-                            let mut file = match fs::File::create(&tmpfile.path()).await {
-                                Ok(f) => f,
-                                Err(e) => {
-                                    warn!("Failed to open tempfile: {}", e);
-                                    break;
-                                }
-                            };
-
-                            // Receive size bytes and dump to the tempfile
-                            let start = time::Instant::now();
-                            match io::copy(
-                                &mut io::AsyncReadExt::take(&mut network_stream, size),
-                                &mut file,
-                            )
-                            .await
-                            {
-                                Ok(n) => {
-                                    debug!(
-                                        "{}: Received {} in {:?}",
-                                        peer,
-                                        bytesize::ByteSize::b(n),
-                                        start.elapsed()
-                                    );
-                                }
-                                Err(e) => {
-                                    warn!("{}: Connection error: {}", peer, e);
-                                    break;
-                                }
-                            }
-
-                            let tmpfile_path = tmpfile.path().to_owned();
-                            keep_file = Some(tmpfile);
-                            Request::Install(repository, tmpfile_path)
-                        } else {
-                            Request::Message(message)
-                        };
-
-                    // Create a oneshot channel for the runtimes reply
-                    let (reply_tx, reply_rx) = oneshot::channel();
-
-                    // Send the request to the runtime
-                    event_tx
-                        .send(Event::Console(request, reply_tx))
-                        .await
-                        .expect("Internal channel error on main");
-
-                    // Wait for the reply from the runtime
-                    select! {
-                        // If the runtime shuts down the connection shall be closed and not wait for
-                        // a reply
-                        _ = stop.cancelled() => break,
-                        Ok(response) = reply_rx => {
-                            keep_file.take();
-
-                            // Report result to client
-                            let message = api::model::Message {
-                                id: message_id,
-                                payload: api::model::Payload::Response(response),
-                            };
-
-                            trace!("{}: <-- {:?}", peer, message);
-                            if let Err(e) = network_stream.send(message).await {
-                                warn!("{}: Connection error: {}", peer, e);
-                                break;
-                            }
+                    let response = match process_request(&peer, &mut network_stream, &event_tx, message).await {
+                        Ok(response) => response,
+                        Err(e) => {
+                            warn!("Failed to process request: {}", e);
+                            break;
                         }
-                        else => break,
+                    };
+                    trace!("{}: <-- {:?}", peer, response);
+
+                    if let Err(e) = network_stream.send(response).await {
+                        warn!("{}: Connection error: {}", peer, e);
+                        break;
                     }
                 }
             }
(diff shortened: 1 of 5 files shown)


northstar/src/runtime/console.rs
```diff
@@ -357,15 +357,16 @@ impl Console {
...
+                    trace!("{:?} -> event loop", request);
+                    let response = match event_tx.send(Event::Console(request, reply_tx)).then(|_| reply_rx).await {
+                        Ok(response) => response,
+                        Err(e) => {
+                            // This could happen if the runtime is in process of shutting down
+                            warn!("Failed to get response from event loop: {}", e);
+                            break;
+                        }
+                    };
+                    trace!("{:?} <- event loop", response);
```
flxo, line 369: If the connections are closed, before everything else it shut down the `expects` are fine. This patch logs things but don't fail on the internal bug when the main loop is down or the runtime core doesn't respond.

northstar/src/runtime/minijail.rs
```diff
@@ -69,16 +70,25 @@ impl<'a> Minijail<'a> {
...
+                        match line.chars().next() {
+                            Some('D') => debug!("{}", l),
+                            Some('I') => info!("{}", l),
+                            Some('W') => warn!("{}", l),
+                            Some('E') => error!("{}", l),
+                            _ => trace!("{}", line),
+                        }
+                    }
+                    _ = task_stop.cancelled() => break,
+                    else => break,
```
flxo, line 91: A nice pattern is to avoid the need to find a name doesn't fit (just because a clone is needed). Put it into braces. Clone everything needed on the top. This also underlines what elements are moved into the task.
```rust
{
   let stop_token = stop_token.clone();
   ....
}
```

northstar/src/runtime/mod.rs
```diff
@@ -206,7 +206,18 @@ async fn runtime_task(config: &'_ Config, stop: CancellationToken) -> Result<(),
...
             // The runtime os commanded to shut down and exit.
-            Event::Shutdown => break state.shutdown().await,
+            Event::Shutdown => {
+                log::debug!("Stopping to process events");
+                drop(event_rx);
+
+                log::debug!("Stopping console");
+                if let Some(console) = console.take() {
+                    console.shutdown().await.map_err(Error::Console)?;
+                }
```
flxo, line 216: Shouldn't this be the other way round. Stop the console and it's connections *before* the main loop?

northstar/src/runtime/process.rs
```diff
@@ -93,9 +93,13 @@ pub(crate) async fn waitpid(
...
-        event_handle
-            .blocking_send(Event::Exit(container, status.clone()))
-            .expect("Internal channel error on main event handle");
+        //
+        // TODO this is needed so that the runtime does not panic on shutdown
+        //
+        match event_handle.blocking_send(Event::Exit(container, status.clone())) {
+            Ok(()) => (),
+            Err(e) => log::warn!("Failed to send container exit status: {}", e),
+        };
```
flxo, line 102: If this is just logged there's a optional hard error that leaks a process ignored. This error must panic!

northstar/src/runtime/state.rs
```diff
@@ -756,10 +757,8 @@ impl<'a> State<'a> {
     }
 
     async fn notification(&self, n: Notification) {
-        self.events_tx
-            .send(Event::Notification(n))
-            .await
-            .expect("Internal channel error on main");
+        // TODO do not ignore this
+        let _ = self.events_tx.send(Event::Notification(n)).await;
```
flxo, line 761: Sending on `event_tx` must be an error.

northstar/src/runtime/mod.rs
alfonsoros88: The event loop has to be stopped first. This _signals_ the open connections that they won't be getting their responses. Then the next step is to wait for those connections to terminate and close the console. Finally the state is shut down.

northstar/src/runtime/minijail.rs
alfonsoros88: Sure, but this function has already too many levels :sweat_smile:

northstar/src/runtime/console.rs
alfonsoros88: The problem are the connections that are at this point waiting for a response. They won't close and therefore won't permit that the console is finally closed. That's the actual issue here.

northstar/src/runtime/process.rs
alfonsoros88: Yes, I need to fine a solution for this one.

northstar/src/runtime/mod.rs
marcmo: Conceptually the event loop should remain processing until no further events are generated. 
We need another way to signal open connections that will not be serviced anymore.

northstar/src/runtime/console.rs
flxo: Is the "waiting for a response" uninterruptible? Should be doable with the `CancelationToken` you pass into the connection loop.

northstar/src/runtime/minijail.rs
flxo: But short ones ;-)

northstar/src/runtime/mod.rs
alfonsoros88: There is this `shutdown` method on the connections, I will try to invert the order with that.

alfonsoros88: @marcmo, what should be the behavior with a shutdown request followed by n other requests? the remaining request should be processed before the shutdown?

northstar/src/runtime/minijail.rs
alfonsoros88: as you wish :)

flxo:
@alfonsoros88 I think the shutdown procedure can be simplified by closing any open connections and listeners (and wait for them as you do) *before* stopping the state. I sketched this in https://github.com/flxo/northstar/tree/connections_during_shutdown
There can't be any console commands in flight when the `Shutdown` event is processes so from my understanding closing the console must be enough to prevent the initial problem that this PR tries to solve.

alfonsoros88:
@flxo I had a look to your suggestion but it seems that after receiving a shutdown request it would stop processing events right after that. @marcmo suggested that we should handle those remaining events in the queue like exits and notifications before closing it. Also when the state shuts down, it potentially produces more of those events.

northstar/src/runtime/minijail.rs
```diff
@@ -469,8 +470,12 @@ impl Process {
     /// it is SIGKILLed.
     pub async fn terminate(&mut self, timeout: time::Duration) -> Result<ExitStatus, Error> {
         debug!("Sending SIGTERM to {}", self.pid);
-        signal::kill(unistd::Pid::from_raw(self.pid as i32), Some(SIGTERM))
-            .map_err(|e| Error::Os(format!("Failed to SIGTERM {}", self.pid), e))?;
+
+        match signal::kill(unistd::Pid::from_raw(self.pid as i32), Some(SIGTERM)) {
+            Err(nix::Error::Sys(Errno::ESRCH)) => return (&mut self.exit_status).await,
```
flxo, line 475: `ESRCH` means that libc couldn't find a matching process. Isn't this an error condition? We're trying to terminate a process and it's not found. Must be something wrong in the state.

flxo:
> @flxo I had a look to your suggestion but it seems that after receiving a shutdown request it would stop processing events right after that. @marcmo suggested that we should handle those remaining events in the queue like exits and notifications before closing it. Also when the state shuts down, it potentially produces more of those events.

The events that might be in the queue don't make sense when the runtime already know that it's gonna shut down. E.g start soemthing. I think also the shutdown notification is nonsense since this event is signaled well via the graceful shutdown of the connection.

northstar/src/runtime/minijail.rs
alfonsoros88: Peter found this issue, this is a race if where the process can exit just before this point in which case its pid is no longer valid. Maybe I should put this as a comment.

alfonsoros88:
@flxo I wouldn't say that it is nonesensical, we handle such events in `State` with the intention to cleanup after containers. Currently we are not leaking things like cgroups because for example both `stop` and `on_exit` call `destroy()` but probably the correct would be to let the cleanup to `on_exit`. In that case we probably would like to handle the remaining events. 

It is similar for the notifications, there is some sense to it. From the point of view of the client there is no difference between a graceful shutdown and a broken connection. The client would probably be interested in listening notifications related to those containers it started.

flxo:
> @flxo I wouldn't say that it is nonesensical, we handle such events in `State` with the intention to cleanup after containers. Currently we are not leaking things like cgroups because for example both `stop` and `on_exit` call `destroy()` but probably the correct would be to let the cleanup to `on_exit`. In that case we probably would like to handle the remaining events.

If the "cleanup" would be just done on `on_exit` the shutdown would need to be synchronised to the last `on_exit` event etc. The topic here's asynchronous stop: Stop containers in parallel like the start. We do not do that currently because the miliseconds don't count on the stopping case as they do when starting.

> It is similar for the notifications, there is some sense to it. From the point of view of the client there is no difference between a graceful shutdown and a broken connection.

There is a difference.

 > The client would probably be interested in listening notifications related to those containers it started.

I don't get this. Why should a client want to wait for notifications after sending a shutdown?

northstar/src/runtime/minijail.rs
flxo: Ok! Thank for pointing out.

A comment would be great.

alfonsoros88:
> If the "cleanup" would be just done on `on_exit` the shutdown would need to be synchronised to the last `on_exit` event etc. The topic here's asynchronous stop: Stop containers in parallel like the start. We do not do that currently because the miliseconds don't count on the stopping case as they do when starting.

Probably  I'm misunderstanding something but I don't see what's the problem. Still sounds like handling the "cleanup" in `on_exit` is the right thing to do. And if as a side product that allows us to have a parallel stopping of the containers then why not? :sweat_smile: 

> There is a difference.

If I fire 2 instances of `nstar` for example, and call `shutdown` from one of them, the message I receive in both instances is "Disconnected, trying to reconnect". Perhaps we are referring to different things.

> I don't get this. Why should a client want to wait for notifications after sending a shutdown?

I meant that if at some point, a user is scripting some client using `northstar::api::client`, I would probably be interested in the notifications, even those sent during shutdown. Again, maybe because currently there seems to be no way to distinguish between a lost of connectivity or a shutdown procedure.

northstar/src/runtime/minijail.rs
```diff
@@ -82,19 +83,30 @@ impl Launcher for Minijail {
...
+                                Some('D') => debug!("{}", l),
+                                Some('I') => info!("{}", l),
+                                Some('W') => warn!("{}", l),
+                                Some('E') => error!("{}", l),
+                                _ => trace!("{}", line),
+                            }
+                        }
+                        _ = stop_token.cancelled() => break,
+                        else => break,
+                    }
```
flxo, line 106: Think this is unrelated to the PR.

northstar/src/runtime/mod.rs
```diff
@@ -272,42 +272,68 @@ async fn runtime_task(config: &'_ Config, stop: CancellationToken) -> Result<(),
...
+                // debug!("Stop accepting new connections");
+                // if let Some(console) = &console {
+                //     console.stop_new_connctions();
+                // }
+
+                debug!("Shutting down runtime");
+                state.shutdown().await.unwrap();
+
+                debug!("Closing event loop");
+                event_rx.close();
```
flxo, line 315: Closing `event_rx` will make this loop break. The flag `shutdown_issued` is never read with value `true`.

northstar_tests/tests/tests.rs
```diff
@@ -347,6 +348,50 @@ test!(connect_version, {
...
+                    .ok(),
+                Err(_) => break,
+            };
+        }
+    });
+
+    let result = runtime.shutdown().await;
+
+    let (fst, snd) = tokio::join!(start_container, stop_container);
+    assert!(fst.is_ok() && snd.is_ok());
```
flxo, line 390: This just checks for spawn errors. The closures in the task return `()`.

flxo (commented):
I like the idea to refactor the listeners into an enum - see comments on the `accept`.

The test is broken and the `shutdown_issued` flag seems wrong to me. Maybe I didn't get the intention.

I'd really vote for closing the listeners and connection upon a shutdown and that's it. That is a graceful close which can be identified by clients.

northstar/src/runtime/console.rs
```diff
@@ -93,130 +101,18 @@ impl Console {
...
-                self.tasks.push(task);
-            }
-            _ => unreachable!(),
-        }
+        let task = match Listener::new(&self.url)
+            .await
+            .map_err(|e| Error::Io("Failed to remove unix socket".into(), e))?
+        {
+            Listener::Tcp(listener) => task::spawn(async move {
+                handle_connections(|| listener.accept(), event_tx, notification_tx, stop).await
```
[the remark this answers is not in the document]
alfonsoros88: yes, `handle_connections` is has a loop to handle new connections.

northstar/src/runtime/console.rs
```diff
@@ -69,6 +69,14 @@ pub enum Error {
     Protocol(String),
     #[error("IO error: {0} ({1})")]
     Io(String, #[source] io::Error),
+    #[error("Event loop closed, request cannot be processed")]
+    EventLoopClosed,
+}
+
+macro_rules! io_err {
```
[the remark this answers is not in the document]
alfonsoros88: :+1:  didn't know that

northstar/src/runtime/minijail.rs
alfonsoros88: yes, partially, I needed this at some point to be able to stop the `tokio` runtime after the shutdown.

northstar/src/runtime/console.rs
```diff
@@ -452,6 +268,197 @@ impl Console {
...
+/// Generic handling of connections. The first parameter is a function that when called awaits for
+/// a new connection. The connections are represented as a pair of a stream and some client
+/// identifier.
+///
+/// All the connections container stored the tasks corresponding to each active connection. As
+/// these tasks terminate, they are removed from the connections container. Once a stop is issued,
+/// the termination of the remaining connections will be awaited.
+///
+async fn handle_connections<AcceptConnection, Connection, Stream, Client, E>(
+    accept: AcceptConnection,
```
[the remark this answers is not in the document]
alfonsoros88: I'm not sure if it has to be a stream, it is _something_ that awaits for a new connection.

northstar/src/runtime/mod.rs
alfonsoros88: https://docs.rs/tokio/1.5.0/tokio/sync/mpsc/struct.Receiver.html#method.close The documentation mentions that a call to `close` will not drop the receiver, i.e. all the events generated in `state.shutdown()` will remain there (e.g. all the stop containers events and notifications). The `shutdown_issued` flag is used in the first arm of this match expression to reply to all those possibly queued console requests with an error.

northstar_tests/tests/tests.rs
alfonsoros88: yes, that's fine. This just checks that those results are ok.

alfonsoros88:
> I like the idea to refactor the listeners into an enum - see comments on the `accept`.
> 
> The test is broken and the `shutdown_issued` flag seems wrong to me. Maybe I didn't get the intention.
> 
> I'd really vote for closing the listeners and connection upon a shutdown and that's it. That is a graceful close which can be identified by clients.

Well yes, at this point this pr is just a scrap yard, the question is if there is still anything of value here. If we want to take the refactorings or not and if we want to change the order of things in the shutdown or not. I will stick to whatever @flxo  and @marcmo  decide.

northstar/src/runtime/console.rs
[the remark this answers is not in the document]
flxo: But doesn't a `listener.accept()` actually just return a future because of of 

```rust
pub async fn accept(&self) -> Result<(TcpStream, SocketAddr)>
```
?

My understanding about futures is that it only can resolve once e.g: [play](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=9f9d15b6b620576b67202799088857ca)

I would get it sorted if the the argument would be `listener.accept` without `()`.

northstar/src/runtime/console.rs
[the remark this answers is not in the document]
flxo: Quite new - you would have seen it during a rebase.

We should have done something like this macro or the fn earlier because this `String` sucks.

flxo:
I'd like to merge the Listener implication because it removes slightly duplicate code.  Same for the minijail log shutdown which can't do any harm. 

I have still a hard time to not just close the connections and send "we're closing soon and ignore your request" before closing the connection. There's no win for the client.

northstar/src/runtime/console.rs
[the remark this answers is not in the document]
alfonsoros88: The first argument is actually a closure that calls `accept`. The closure is called inside the loop in `handle_connections`. It's kind of a trick that I used to avoid writing a new trait for `UnixListener` and `TcpListener` and just use a `AcceptConnection: Fn() -> NewConnection` generic.

northstar/src/runtime/console.rs
```diff
@@ -93,130 +101,18 @@ impl Console {
...
-        }
+        let task = match Listener::new(&self.url)
+            .await
+            .map_err(|e| Error::Io("Failed to remove unix socket".into(), e))?
+        {
+            Listener::Tcp(listener) => task::spawn(async move {
+                handle_connections(|| listener.accept(), event_tx, notification_tx, stop).await
+            }),
+            Listener::Unix(listener) => task::spawn(async move {
+                handle_connections(|| listener.accept(), event_tx, notification_tx, stop).await
```
marcmo, line 112: quite elegant how you removed the code duplication here

northstar/src/runtime/console.rs
```diff
@@ -452,6 +268,197 @@ impl Console {
...
+    notification_tx: broadcast::Sender<Notification>,
+    stop: CancellationToken,
+) where
+    AcceptConnection: Fn() -> Connection,
+    Connection: Future<Output = Result<(Stream, Client), E>>,
+    Stream: AsyncWrite + AsyncRead + Unpin + Send + 'static,
+    Client: Into<ClientId>,
+    E: std::fmt::Debug,
+{
+    let mut connections = FuturesUnordered::new();
```
marcmo, line 408: didn't know about this `FuturesUnordered`...where did you learn about that?

northstar/src/runtime/minijail.rs
marcmo: makes sense

northstar/src/runtime/console.rs
```diff
@@ -452,6 +268,197 @@ impl Console {
...
+            send_event(&event_loop, Request::Message(message)).await?
+        };
+
+    Ok(api::model::Message {
+        id: message_id,
+        payload: api::model::Payload::Response(response),
+    })
+}
+
+/// Sends an `Event::Console` to the event loop and await its response
```
marcmo, line 336: more clear now

marcmo:
looks rather good for my taste...I'd merge it

northstar/src/runtime/console.rs
alfonsoros88: this one is used already in master, @flxo  introduced it in the last update to `console.rs`

marcmo:
oh...just saw that this pr seems to be obsolete and you added #331 instead? if so let's close this @alfonsoros88

PR #320, opened 2021-03-23
Revert "minijail: use pid namespace"

pclesr:
This reverts commit efd315d1cc9287b3dd4e7755d040a97e16d143d0.

When launching multiple containers in parallel, libc can
deadlock when using pid namespaces. The root cause is still
under investigation, but all threads end up waiting on a futex.

### DIFF
--- a/northstar_tests/src/runtime.rs
+++ b/northstar_tests/src/runtime.rs
@@ -27,7 +27,7 @@ use northstar::{
 };
 use std::{collections::HashMap, convert::TryInto, path::PathBuf, time::Duration};
 use tempfile::TempDir;
-use tokio::{fs, select, time};
+use tokio::fs;
 
 pub struct Northstar {
     config: config::Config,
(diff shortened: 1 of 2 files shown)


alfonsoros88 (approved):
I think there is some problem with the unit tets, I don't think this patch is the cause though. I'll see if I can get the tests fixed.

flxo:
I would expect this change to touch the `minijail` module only. Why are there changes at all at the tests? Those can't be related to the pid ns.

alfonsoros88:
@flxo, the removal of `run_as_init()` in minijail changes the expected exit status of some containers in the tests. This is because now we get the exit status of the actual process instead of the fake init. For example, we should expect a `Signaled` with `SIGABORT` in the case of the crashing containers tests instead of some `ExitCode`.

northstar_tests/src/runtime.rs
```diff
@@ -212,23 +212,19 @@ impl Northstar {
 
     // TOOD: Queue the notifications in the runtime struct. Currently there's a race
     // if the notification is faster.
-    pub async fn assume_notification<F>(&mut self, mut pred: F, timeout: u64) -> Result<()>
+    pub async fn assume_notification<F>(&mut self, mut pred: F, timeout: Duration) -> Result<()>
```
flxo, line 215: I'd not use `Duration` here since it's bulky to type and also keep it in sync with the logger assumption API that accepts seconds also.

northstar_tests/tests/tests.rs
```diff
@@ -214,16 +220,24 @@ test!(crashing_container, {
...
                         status: ExitStatus::Exit(254),
-                    }
+                    };
+
+                    let aborted = Notification::Exit {
+                        container: TEST_CONTAINER.try_into().unwrap(),
+                        status: ExitStatus::Signaled(6),
+                    };
+
+                    n == &exited || n == &aborted
```
flxo, line 238: Shouldn't the test container always exit with the same result?

flxo:
> @flxo, the removal of `run_as_init()` in minijail changes the expected exit status of some containers in the tests. This is because now we get the exit status of the actual process instead of the fake init. For example, we should expect a `Signaled` with `SIGABORT` in the case of the crashing containers tests instead of some `ExitCode`.

Ok. Got it. But shouldn't the test container then exit always with the same result? See my comment above.

northstar_tests/src/runtime.rs
alfonsoros88: changed back but i think such assumptions and internal conversions are not nice for a API

northstar_tests/tests/tests.rs
alfonsoros88: Yes, you are right. It is now expected to exit via an abort signal.

alfonsoros88:
@pclesr closing this pr because the main patch got its way to master by other means and what remains makes no longer sense.

PR #323, opened 2021-03-26
Check process status when SIGCHLD is received

alfonsoros88:
We use `waitpid` inside a `spawn_blocking` task. This change intends to
replace it with async `spawn` by awaiting for `SIGCHLD` to then check
the process status with `waitpid` with `WNOHANG` set.

### DIFF
--- a/northstar/src/runtime/island/mod.rs
+++ b/northstar/src/runtime/island/mod.rs
@@ -21,13 +21,13 @@ use super::{
     Event, EventTx, ExitStatus, Pid,
 };
 use async_trait::async_trait;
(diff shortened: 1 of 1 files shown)


flxo:
@alfonsoros88 I ported your patch to island. I replaced the `oneshot` channel with the future returned from `tokio::task::spawn`.

alfonsoros88:
Then i think we should close this one :)

flxo:
@alfonsoros88 I hijacked this PR for the port ;-). Would ne nice if you can have a look at 28da17c.

PR #324, opened 2021-04-07
Remove nstar interactive mode

### LINKED ISSUE #321: Remove interactive nstar mode
The use of `rustyline` is pain since month because of the not interruptible blocking readline fn. Todays modern shell provide almost the same feature set as rusty line and with some hand crafted completion script in addition to the generated ones `from` clap there should be no need fo the interactive mode. When removing the interactive mode, `nstar` should get a mode that just dumps notifications since a single command passed on the cli will not print notifications.

flxo:
The blocking read in the interactive mode has been broken since it exits. This patch removes the interactive mode and uses the existing subcommand parsing structure as interface. Fix the json mode that was broken for installations. Additionally:

* Print api version in help text
* Add a notification dump only mode
* Add a shell completion generation subcommand
* (Finally) implement pretty printing for notifications
* Add non zero exit codes upon errors
* Add connect timeout option with 10 seconds default
* Cleanup of dependencies

Fixes #321


northstar/src/api/codec.rs
```diff
@@ -23,21 +23,34 @@ use std::{
...
+    pub fn into_parts(self) -> FramedParts<T, Codec> {
+        self.inner.into_parts()
+    }
+
+    /// Consumes the Framed, returning its underlying I/O stream.
+    pub fn into_inner(self) -> T {
+        self.inner.into_inner()
+    }
+}
+
```
alfonsoros88, line 44: I couldn't understand what `into_parts()` does so i tried to look for where it is used but it seems it is not used? The same seems to happen with `into_inner()`. I'm probably missing something, what is this `impl` block for?

nstar/src/main.rs
```diff
@@ -12,141 +12,250 @@
...
-            return match client.install(&npk, &repo_id).await {
-                Ok(()) => Ok(Response::Ok(())),
-                Err(e) => Ok(Response::Err(api::model::Error::Npk(e.to_string()))),
-            };
-        }
-        Northstar::Uninstall { name, version } => Request::Uninstall(Container::new(name, version)),
-        Northstar::Shutdown => Request::Shutdown,
-    };
+// About help string generated by build.rs because of static requirements
+const ABOUT: &str = env!("NSTAR_ABOUT");
```
alfonsoros88, line 32: Just curious, why is this about taken from an environment variable defined in the `build.rs`?

nstar/src/main.rs
```diff
@@ -12,141 +12,250 @@
...
+    /// Connect timeout in seconds
+    #[structopt(short, long, default_value = "10", parse(try_from_str = parse_duration))]
+    pub timeout: time::Duration,
+    /// Command
+    #[structopt(subcommand)]
+    pub command: Subcommand,
+}
+
+/// Parse a str containing a u64 into a `std::time::Duration` and take the value
+/// value as seconds
```
alfonsoros88, line 129: value is written twice

nstar/src/main.rs
```diff
@@ -12,141 +12,250 @@
...
+fn parse_duration(src: &str) -> Result<time::Duration, anyhow::Error> {
```
alfonsoros88, line 130: nit: maybe consider naming it to something like `parse_secs` just to signal the unit casting in the signature.

alfonsoros88 (approved):
Looks good to me :+1:

northstar/src/api/codec.rs
flxo: We wrap the tokio_util::codec::Framed in an struct to make to fix the codec type. This hides the API from `tokio_util::codec::Framed` (`into_parts` and `into_inner`) which is needed if the raw transport shall be obtained back from the struct.

nstar/src/main.rs
flxo: The reason is that the `structopt` about needs a string with static lifetime. The version is obtained via `northstar::api::model::version` which is not `'static`. The env variable is one way to solve this. Otherwise we could use `lazy_static` but I wanted to avoid the additional crate dep.

nstar/src/main.rs
flxo: Done

nstar/src/main.rs
flxo: Done.

marcmo:
CHANGELOG needs to be updated

nstar/src/main.rs
alfonsoros88: Got it! There is this trick to get a `'static` life time out of things, I just tried it and it works, maybe you can replace this with:

```
fn about() -> &'static str {
    Box::leak(Box::new(format!("Nstar version {}", version())))
}

...
#[structopt(name = "nstar", author, about = about(), global_setting(AppSettings::ColoredHelp))]
```

and remove `build.rs`. It would be a bit simpler I think

northstar/src/api/codec.rs
alfonsoros88: Understood, it is just that those methods seem to not be currently used in the project.

nstar/src/main.rs
flxo: Ah. Thanks. That makes sense! Will add this.

flxo:
> CHANGELOG needs to be updated

done.

marcmo (approved):

PR #325, opened 2021-04-09
minijail: re-enable pid namespaces

pclesr:
This uses a consistent interface to fork/clone in minijail. Earlier
versions mixed library calls (which acquire futexes during atfork()
calls) with directly invoked syscalls, bypassing the library handling.
As documented in the minijail code, this can lead to deadlocks. Most
likely, when first developed, the library code was not available.

With this fix, and the recent fixes to the runtime for handling
process termination, a 'spawner' test of 20 clients starting and
stopping can run in the release build for more than 36 hours.

### DIFF
--- a/minijail/libminijail.c
+++ b/minijail/libminijail.c
@@ -2761,6 +2761,41 @@ pid_t API minijail_fork(struct minijail *j)
 	return minijail_run_config_internal(j, &config);
 }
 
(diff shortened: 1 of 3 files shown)


alfonsoros88 (approved):
Looks good to me! :+1:

PR #326, opened 2021-04-14
Introduce a Launcher trait

flxo:
The minijail crate is going to be replaced with something easier to debug
and maintain. This is a preparation for that in order to make `State`
generic over process launcher implementation.


northstar/src/runtime/mod.rs
```diff
@@ -60,10 +60,33 @@ pub(self) type RepositoryId = String;
 pub use api::container::*;
 pub(self) use repository::Repository;
 pub(self) use state::MountedContainer;
-
 pub(self) type ExitCode = i32;
 pub(self) type Pid = u32;
 
+#[async_trait]
+pub(self) trait Process {
```
marcmo, line 67: why do you use `pub(self)`? isn't it the same like using nothing?
If it is just used in the current module at least

northstar/src/runtime/mod.rs
```diff
@@ -60,10 +60,33 @@ pub(self) type RepositoryId = String;
...
+    fn pid(&self) -> Pid;
+    async fn start(&mut self) -> Result<(), Error>;
+    async fn stop(mut self, timeout: time::Duration) -> Result<ExitStatus, Error>;
+}
+
+#[async_trait]
+pub(self) trait Launcher {
```
marcmo, line 74: will probably need different visibility if we implement it in another module/crate

northstar/src/runtime/state.rs
```diff
@@ -36,61 +30,56 @@ use npk::{
...
-    fmt,
     path::{Path, PathBuf},
     result,
     sync::Arc,
 };
 use tokio::{sync::oneshot, task, time};
 
 #[derive(Debug)]
-pub(super) struct State<'a> {
+pub(super) struct State<'a, L>
```
marcmo, line 40: would be nice to get rid of the lifetime here as well

northstar/src/runtime/state.rs
```diff
@@ -36,61 +30,56 @@ use npk::{
...
-pub(super) struct ProcessContext {
+pub(super) struct ProcessContext<P> {
+    process: P,
     started: time::Instant,
-    process: Process,
     cgroups: Option<super::cgroups::CGroups>,
 }
 
-impl ProcessContext {
+impl<P: Process> ProcessContext<P> {
```
marcmo, line 77: shouldn't the restriction for the type parameter done in the definition of the `ProcessContext` struct?

northstar/src/runtime/mod.rs
flxo: Correct.

northstar/src/runtime/mod.rs
flxo: The `Launcher` is not publicly exchangeable by any user of `northstar::runtime`. This is not intended and would need to export a lot of more stuff we probably don't want to export.

northstar/src/runtime/state.rs
flxo: Why do you want to remove the lifetime? It's just used for the ref of the config to avoid a copy.

northstar/src/runtime/state.rs
flxo: The definition of `ProcessContext` doesn't need the restriction. This is why it's just in the `impl` block to ensure that `Process::stop` is implemented for `P`

PR #329, opened 2021-04-19
Replace homebrewn GH actions cache with Swatinem/rust-cache@v1

flxo:
The cache setup is complicated and misses a couple of cases. Use
Swatinem/rust-cache@v1 which has good defaults for this straight forward
rust project.


marcmo (approved):
if you verified it works as intended it's good

PR #330, opened 2021-04-19
Pr island

flxo:
Draft of a homegrown minijail replacement.


examples/container/inspect/src/main.rs
```diff
@@ -0,0 +1,100 @@
...
+            .trim_start_matches('(')
+            .trim_end_matches(')');
+        println!("    {:>10}: {:<10}: {}", pid, name, cmdline);
+    }
+
+    let mut sigs = Signals::new(SIGNALS).expect("install signal handler");
+    for signal in &mut sigs {
+        eprintln!("Received signal {:?}", signal);
+        let signal = signal.signal;
+        low_level::emulate_default_handler(signal).expect("default signal handler");
```
marcmo, line 98: even after reading the documentation I'm not sure what this is doing

northstar/src/runtime/pipe.rs
```diff
@@ -13,7 +13,7 @@
 //   limitations under the License.
```
marcmo, line 14: the pipe module is kind of hard to read and would certainly benefit from some explaining comments

northstar/src/runtime/debug.rs
```diff
@@ -131,9 +187,9 @@ impl Strace {
         self.token.cancel();
 
         // Stop strace - if not already existed - irgnore any error
+        let pid = self.child.id().unwrap();
```
marcmo, line 190: I would not unwrap() for a log message

northstar/src/runtime/island/init.rs
```diff
@@ -0,0 +1,586 @@
...
+macro_rules! cerror { ($($arg:tt)+) => ( log::error!("{}: {}", std::process::id(), format!($($arg)+))) }
+
+// Init function. Pid 1.
+pub(super) fn init(
+    config: &Config,
+    container: &Container<IslandProcess>,
+    mut fds: HashMap<i32, i32>,
+    groups: &[(String, Option<Gid>)],
+    mut intercom: Intercom,
+) -> ! {
```
marcmo, line 70: what's the `!` about?

northstar/src/runtime/island/init.rs
```diff
@@ -0,0 +1,586 @@
...
+    // We cannot use log after here because the fd to logd is closed on Android
+
+    Ok(())
+}
+
+// TODO: The container could be malformed and the mountpoint might be
+// missing. This is not a fault of the RT so don't expect it.
+// TODO: mount flags: nosuid etc....
+// TODO: /dev mounts from manifest: full or minimal
+fn rootfs(config: &Config, container: &Container<IslandProcess>) -> anyhow::Result<()> {
```
marcmo, line 217: the whole `init` would benifit from `thiserror` Error definitions

northstar/src/runtime/island/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+                // Wait for the notification from init that the child is launchend (8)
+                intercom.recv::<LaunchProtocol>().ok();
+
+                Ok(IslandProcess::Started {
+                    pid,
+                    exit_status,
+                    io: _io,
+                })
+            }
+            _ => unreachable!(),
```
marcmo, line 189: so it's always an error to call start on an already started process?

northstar/src/runtime/island/mod.rs
```diff
@@ -0,0 +1,348 @@
...
+        let (pid, mut exit_status) = match self {
+            IslandProcess::Created {
+                pid, exit_status, ..
+            } => (pid, exit_status),
+            IslandProcess::Started {
+                pid,
+                exit_status,
+                io: _io,
+            } => (pid, exit_status),
+            IslandProcess::Stopped { .. } => unreachable!(),
```
marcmo, line 208: same thing for stop...shouldn't it just be a noop if the process is already stopped?

marcmo:
that's quite a handful...from a rust point of view it already looks really good. The process handling also looks good but I don't understand it all.

examples/container/inspect/src/main.rs
flxo: This handler prints the received signal and does best effort to do the same as the default handler would do. Probably just exit.

northstar/src/runtime/debug.rs
flxo: What log message? 
Getting the pid of a strace child is very likely be good. Will change that to an `expect`

northstar/src/runtime/island/init.rs
flxo: It says the fn does never return.

northstar/src/runtime/island/init.rs
flxo: This would be just boilerplate since the errors are never passed outside of `init`. Using anyhow is convenient because the only thing what could be done with those is printing the in `init.

northstar/src/runtime/island/mod.rs
flxo: This should be handled in the `state` mod. The `island` mod doesn't care about.

northstar/src/runtime/island/mod.rs
flxo: Think this is a logical bug if someone calls stop on a stopped process.

marcmo:
Missing: CHANGELOG update

northstar/src/runtime/state.rs
```diff
@@ -766,10 +766,12 @@ impl<'a, L: Launcher> State<'a, L> {
...
-        self.events_tx
-            .send(Event::Notification(n))
-            .await
-            .expect("Internal channel error on main");
+        if !self.events_tx.is_closed() {
+            self.events_tx
+                .send(Event::Notification(n))
+                .await
+                .expect("Internal channel error on main");
+        }
```
marcmo, line 774: why not just ignore the send error...this seems racy and all the send error indicates is a closed or dropped channel anyway

PR #332, opened 2021-04-22
CI cleanup

flxo:
* Remove obsolete explicit `libcap-dev` install (libcap is committed to the workspace)
* Remove `cargo check`: a `cargo clippy` is a superset of tests compared to `cargo `check`
* Add caching of docker images


alfonsoros88 (approved):
Sweet <3

PR #336, opened 2021-05-04
Implement Codec in terms of tokio_util::codec::LinesCodec

rvarago:
Greetings, Northstar development team,

This PR implements Northstar's codec in terms of [tokio_util::codec::LinesCodec](https://docs.rs/tokio-util/0.6.6/tokio_util/codec/struct.LinesCodec.html), which relieves Northstar from handling newlines, etc. Therefore, leaving it with the sole responsibility of composing `tokio_utils` with `serde_json`.

Furthermore, to reduce, albeit _not_ eliminate, the chance of introducing a regression, I'd written a rather simple unit test before the refactoring. In the test, we leverage [proptest](https://docs.rs/proptest/1.0.0/proptest/index.html) to check for the codec consistency, such that encoding a message and then decoding it yields another message that compares equal to the original one. That is, modulo some details, e.g. `Result`, `BytesMut`, etc.

I am not entirely sure whether this PR is _really_ an improvement. Thus, I'd love to learn your thoughts.

Thank you ^^.

### DIFF
--- a/northstar/src/api/codec.rs
+++ b/northstar/src/api/codec.rs
@@ -13,7 +13,6 @@
 //   limitations under the License.
 
 use super::model;
-use bytes::{Buf, BufMut};
 use futures::Stream;
 use std::{
     cmp::min,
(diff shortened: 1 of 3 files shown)


northstar/src/api/codec.rs
```diff
@@ -155,3 +152,36 @@ impl<T: Unpin + AsyncRead + AsyncWrite> futures::sink::Sink<model::Message> for
...
+            assert_eq!(message, Some(initial_message));
+        }
+    }
+
+    fn mk_message() -> impl Strategy<Value = model::Message> {
+        mk_simple_payload().prop_map(model::Message::new)
+    }
+
+    fn mk_simple_payload() -> impl Strategy<Value = model::Payload> {
+        // TODO: Perhaps consider all the variants?
```
marcmo, line 184: allowing the strategy to pick all possible variants would be indeed nicer!

marcmo:
nice...thanks for the PR. using proptest here is quite appropriate. And the code is simplified using the LineCodec. @flxo also will take a look on it.

PR #352, opened 2021-05-20
Remove tokio dependency from npk

flxo:
The IO portion of the npk crate got concentrated some time agon in the "constructor" methods of `npk::Npk`. Since tokio itself
also doesn't do anything else than a block_in_place internally, removing tokio as a dep of npk shortens the list of dependencies and reduces compile time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1171,7 +1171,6 @@ dependencies = [
  "proc-mounts",
  "procinfo",
  "proptest",
- "regex",
  "serde",
  "serde_json",
  "tempfile",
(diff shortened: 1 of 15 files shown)


inorick (approved):
I think we are loosing some granularity in functions that contained several calls to `.await()` and are now just one big synchronous function (e.g. `Builder::build()`). If it is not a performance issue then I think the simplification is worth it.

flxo:
The "online" use case: read and parse an NPK spends most time in the io. The parsing of the yaml is fast so there's no real change on the performance topic.

"offline" means packing NPK isn't performance critical at all. Here the far most time is spent in `mksquashfs`.

PR #354, opened 2021-05-25
Add seccomp Allowlist struct

### LINKED ISSUE #342: Island: Avoid allocation in seccomp
The capability dropping in the island impl allocates and frees. This is problematic because when the rt clones there might be a lock somewhere in the libc held that causes a deadlock on *any* malloc/free in the cloned child process (in this case `init`).

Therefore after the `clone` of the `init` there should be *no* allocation. The seccomp code does in 

https://github.com/esrlabs/northstar/blob/fb430ec51dd34bd271bcae60a4d3e0f75ccfc77b/northstar/src/runtime/island/seccomp.rs#L140

and 

https://github.com/esrlabs/northstar/blob/fb430ec51dd34bd271bcae60a4d3e0f75ccfc77b/northstar/src/runtime/island/seccomp.rs#L116

Move the allocation to `clone` calling context. See #341 which is the same issue.

inorick:
Fixes #342


flxo:
Thanks for the typo fixes.

flxo (approved):
Looks good!

flxo:
@inorick I rebased the PR and tried on `aarch64-linux-android`. Looks like we cannot fail when a sys call lookup fails because this is an error of the manifest - not the rt.

PR #362, opened 2021-06-10
Startup and installation

flxo:
Various optimisations:

* Do not read the NPK twice when loading the repo and when starting. Read the npks (and verify) once on startup and upon installation
* Replace internal repo tmpdir with an in memory `memfd` to avoid leftovers
* Stream npks directly into repository during installation. Do not dump them to a tmp file and copy later on. Instead pass a Stream to the repository and dump directly to the final position
* Make `Npk` generic of the internal `Read + Seek`

### DIFF
--- a/northstar/src/runtime/console.rs
+++ b/northstar/src/runtime/console.rs
@@ -267,9 +267,7 @@ impl Console {
 ///
 /// # Errors
 ///
-/// Installing requests will cause this function to create a temporary file where to copy the
-/// incoming npk file. It can potentially produce an `Error::Io`.
-///
+/// If the streamed NPK is not valid and parseable a `Error::Npk(..)` is returned.
 /// If the event loop is closed due to shutdown, this function will return `Error::EventLoopClosed`.
 ///
 async fn process_request<S>(
@@ -300,7 +298,7 @@ where
         let event = Event::Console(request, reply_tx);
         event_loop.send(event).map_err(|_| Error::Shutdown).await?;
 
-        // If the connections breaks: just break. If the receiver is dropp: just break.
+        // If the connections breaks: just break. If the receiver is dropped: just break.
         let mut take = ReaderStream::new(BufReader::new(stream.take(size)));
         while let Some(Ok(buf)) = take.next().await {
             if tx.send(buf).await.is_err() {
--- a/northstar/src/runtime/console.rs
+++ b/northstar/src/runtime/console.rs
@@ -18,7 +18,7 @@ use crate::{
     runtime::{EventTx, ExitStatus},
 };
 use api::model;
-use bytes::BytesMut;
+use bytes::Bytes;
 use futures::{
     future::join_all,
     sink::SinkExt,
@@ -26,26 +26,25 @@ use futures::{
     Future, StreamExt, TryFutureExt,
 };
 use log::{debug, error, info, trace, warn};
-use std::{fmt, io::Write, path::PathBuf, unreachable};
-use tempfile::NamedTempFile;
+use std::{fmt, path::PathBuf, unreachable};
 use thiserror::Error;
 use tokio::{
     fs,
-    io::{self, AsyncRead, AsyncReadExt, AsyncWrite},
+    io::{self, AsyncRead, AsyncReadExt, AsyncWrite, BufReader},
     net::{TcpListener, UnixListener},
     pin, select,
-    sync::{self, broadcast, oneshot},
+    sync::{self, broadcast, mpsc, oneshot},
     task::{self},
     time,
 };
-use tokio_util::{either::Either, sync::CancellationToken};
+use tokio_util::{either::Either, io::ReaderStream, sync::CancellationToken};
 use url::Url;
 
 // Request from the main loop to the console
 #[derive(Debug)]
 pub(crate) enum Request {
     Message(model::Message),
-    Install(RepositoryId, PathBuf),
+    Install(RepositoryId, mpsc::Receiver<Bytes>),
 }
 
 /// A console is responsible for monitoring and serving incoming client connections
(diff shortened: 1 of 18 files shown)


northstar/src/runtime/console.rs
```diff
@@ -278,86 +277,57 @@ async fn process_request<S>(
     stream: &mut S,
     stop: &CancellationToken,
     event_loop: &EventTx,
-    message: api::model::Message,
-) -> Result<api::model::Message, Error>
+    message: model::Message,
+) -> Result<model::Message, Error>
```
marcmo, line 281: The comment above seems outdated

northstar/src/runtime/console.rs
```diff
@@ -278,86 +277,57 @@ async fn process_request<S>(
...
+        info!("{}: Using repository \"{}\"", client_id, repository);
+
+        // Send a Receiver<Bytes> to the runtime and forward n bytes to this channel
+        let (tx, rx) = mpsc::channel(10);
+        let request = Request::Install(repository, rx);
+        trace!("    {:?} -> event loop", request);
+        let event = Event::Console(request, reply_tx);
+        event_loop.send(event).map_err(|_| Error::Shutdown).await?;
+
+        // If the connections breaks: just break. If the receiver is dropp: just break.
```
marcmo, line 303: dropped

northstar/src/runtime/repository.rs
```diff
@@ -12,100 +12,305 @@
...
+        self.containers
+            .values()
+            .map(|(_, npk)| npk.clone())
+            .collect()
+    }
+}
+
+/// In memory repository
+#[derive(Default, Debug)]
+pub(super) struct MemRepository {
```
marcmo, line 235: nice idea

northstar/src/runtime/repository.rs
```diff
@@ -12,100 +12,305 @@
...
+                .map_err(|e| Error::Io("Failed to write npk".into(), e))?;
         }
-
-        // Copy the npk to the repository
-        fs::copy(src, &dest)
+        file.flush()
             .await
-            .map_err(|e| Error::Io("Failed to copy npk to repository".into(), e))?;
+            .map_err(|e| Error::Io("Failed to flush npk".into(), e))?;
+        drop(file);
```
marcmo, line 161: why do we need the drop here?

northstar/src/runtime/mount.rs
```diff
@@ -99,23 +102,11 @@ impl MountControl {
 
         task::spawn(async move {
             let start = time::Instant::now();
-
             let manifest = npk.manifest();
-            debug!("Mounting {}:{}", manifest.name, manifest.version);
-            let use_verity = key.is_some();
-
-            if npk.version() != &Manifest::VERSION {
```
marcmo, line 107: do we not need the version check anymore here?

northstar/src/runtime/repository.rs
```diff
@@ -12,100 +12,305 @@
...
+    sync::Arc,
+};
+use tokio::{fs, io::AsyncWriteExt, sync::mpsc, task, time::Instant};
+
+#[async_trait::async_trait]
+pub(super) trait Repository: fmt::Debug {
+    /// Add npk from `src` to repositoriy. Open the npk and parse content
+    async fn insert(&mut self, rx: &mut Receiver<Bytes>) -> Result<Container, Error>;
+
+    /// Add container from repository if present
```
marcmo, line 46: comment seems wrong

northstar/src/runtime/repository.rs
```diff
@@ -12,100 +12,305 @@
...
+        debug!("Loading buffer");
+        let npk = npk::Npk::from_reader(file, None).map_err(|e| Error::Npk("Memory".into(), e))?;
+        let manifest = npk.manifest();
+        let container = Container::new(manifest.name.clone(), manifest.version.clone());
+
+        self.containers
+            .insert(container.clone(), (data, Arc::new(npk)));
+
+        info!("Loaded {} into {}", container, self.id);
+
```
marcmo, line 261: why not pass in the channel and directly write from it? Then we can get rid of another copy operation also for the in-memory file system.

northstar/src/runtime/repository.rs
```diff
@@ -12,100 +12,305 @@
...
+#[async_trait::async_trait]
+impl<'a> Repository for MemRepository {
+    async fn insert(&mut self, rx: &mut Receiver<Bytes>) -> Result<Container, Error>
+    where
+        Self: Sized,
+    {
+        let mut buffer = Vec::new();
+        while let Some(r) = rx.recv().await {
+            buffer.extend(&r);
+        }
```
marcmo, line 288: this could be moved into the function that deals with the in-memory fs (see above)

marcmo:
very nice ideas implemented! I like it!

northstar/src/runtime/console.rs
flxo: Fixed.

northstar/src/runtime/console.rs
flxo: Fixed.

northstar/src/runtime/repository.rs
flxo: :-)

northstar/src/runtime/repository.rs
flxo: The drop is not needed explicitly because the `flush` should do the trick. Anyways I wanted to get the fd to be closed before the npk is opened again for parsing.

northstar/src/runtime/mount.rs
flxo: No because it's done when the npk is added to repository.

northstar/src/runtime/repository.rs
flxo: Fixed.

northstar/src/runtime/repository.rs
flxo: `memfd_create` needs a continuous memory region. Therefore we need to allocate `buf` first. `memfd_create` cannot operate on a stream.

northstar/src/runtime/repository.rs
flxo: `Repository::insert` is declared in the `Repository` trait. But the `MemRepository` needs an additional interface compared to `Repository` for loading the compiled in `hello-world` example (The npk included via `include_bytes!`. `MemRepository` needs also to satisfy `Repository` because it can also be used for installing npks via the console.

PR #366, opened 2021-06-15
Precalculate errors in mount()

inorick:
If a manifest's seccomp part contains an unknown syscall we now return an error to the runtime instead of disabling seccomp.

### DIFF
--- a/northstar/src/runtime/island/fs.rs
+++ b/northstar/src/runtime/island/fs.rs
@@ -39,15 +39,12 @@ pub(super) struct Mount {
     pub fstype: Option<&'static str>,
     pub flags: MsFlags,
     pub data: Option<String>,
+    pub err_str: String,
 }
 
 impl Mount {
     /// Execute this mount call
-    pub(super) fn mount(&self) {
-        if !self.target.exists() {
-            panic!("Missing mount point {}", self.target.display())
-        }
-
+    pub(super) fn mount(&self) -> Result<(), ()> {
         nix::mount::mount(
             self.source.as_ref(),
             &self.target,
@@ -55,7 +52,8 @@ impl Mount {
             self.flags,
             self.data.as_deref(),
         )
-        .unwrap_or_else(|_| panic!("Failed to mount {:?}", self));
+        .expect(&self.err_str);
+        Ok(())
     }
 }
 
@@ -114,6 +112,7 @@ pub(super) async fn prepare_mounts(
 fn proc(root: &Path) -> Mount {
     debug!("Mounting /proc");
     let target = root.join("proc");
+    let err_str = format!("Failed to mount {}", &target.display());
     let flags = MsFlags::MS_RDONLY | MsFlags::MS_NOSUID | MsFlags::MS_NOEXEC | MsFlags::MS_NODEV;
     Mount {
         source: Some(PathBuf::from("proc")),
@@ -121,6 +120,7 @@ fn proc(root: &Path) -> Mount {
         fstype: Some("proc"),
         flags,
         data: None,
+        err_str,
     }
 }
 
@@ -135,6 +135,7 @@ fn bind(root: &Path, target: &Path, host: &Path, options: &MountOptions) -> Vec<
             options.iter().collect::<Vec<_>>(),
         );
         let target = root.join_strip(target);
+        let err_str = format!("Failed to mount {}", &target.display());
         let mut flags = options_to_flags(&options);
         flags.set(MsFlags::MS_BIND, true);
         mounts.push(Mount {
@@ -143,6 +144,7 @@ fn bind(root: &Path, target: &Path, host: &Path, options: &MountOptions) -> Vec<
             fstype: None,
             flags: MsFlags::MS_BIND | flags,
             data: None,
+            err_str: err_str.clone(),
         });
 
         if !rw {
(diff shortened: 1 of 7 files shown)


flxo (commented):
Hm. The PR is tricky. We're not allowed to do any allocation (even in error handling) in the init process because this can deadlock in the malloc code. Formatting the context is a allocation. 
We could add a `description` to `Mount` that is passed to the `expect`. This description shall be formatted when the list of mounts is assembled prior the the first `clone`.

inorick:
Agreed. Alternatively, we could use global string constants for those message.

northstar/src/runtime/island/fs.rs
```diff
@@ -39,23 +39,23 @@ pub(super) struct Mount {
     pub fstype: Option<&'static str>,
     pub flags: MsFlags,
     pub data: Option<String>,
+    pub err_str: String,
 }
 
+// TODO: The container could be malformed and the mountpoint might be missing.
+// This is not a fault of the RT so don't expect it.
```
flxo, line 46: The mount points are in the the roofs and under the responsibility of the one that creates a NPK. The runtime cannot create the mount points because of the read onliness of squashfs. This means that malicious container might lack mount points.

northstar/src/runtime/island/fs.rs
```diff
@@ -114,13 +114,15 @@ pub(super) async fn prepare_mounts(
 fn proc(root: &Path) -> Mount {
     debug!("Mounting /proc");
     let target = root.join("proc");
+    let err_str = format!("Failed to mount {}", &target.display());
```
flxo, line 117: I'd include all options. E.g:

```rust
format!("Failed to mount {} of type {} on {} with flags {}, ...)
```

northstar/src/runtime/island/fs.rs
```diff
@@ -143,6 +146,7 @@ fn bind(root: &Path, target: &Path, host: &Path, options: &MountOptions) -> Vec<
             fstype: None,
             flags: MsFlags::MS_BIND | flags,
             data: None,
+            err_str: err_str.clone(),
```
flxo, line 149: This cries for `Mount::new(target, source, type, flags, data)` that internally constructs `err_str`.

northstar/src/runtime/island/fs.rs
```diff
@@ -39,23 +39,23 @@ pub(super) struct Mount {
     pub fstype: Option<&'static str>,
     pub flags: MsFlags,
     pub data: Option<String>,
+    pub err_str: String,
```
flxo, line 42: Possibly rename to `error_msg`

northstar/src/runtime/island/fs.rs
```diff
@@ -39,23 +39,23 @@ pub(super) struct Mount {
...
+// TODO: The container could be malformed and the mountpoint might be missing.
+// This is not a fault of the RT so don't expect it.
 impl Mount {
     /// Execute this mount call
-    pub(super) fn mount(&self) {
-        if !self.target.exists() {
-            panic!("Missing mount point {}", self.target.display())
-        }
-
+    pub(super) fn mount(&self) -> Result<(), ()> {
```
flxo, line 49: The `Result<(), ()>` doesn't make sense and is unused. There's the `expect` which is fine.

northstar/src/runtime/island/init.rs
```diff
@@ -68,7 +68,7 @@ pub(super) fn init(
         .expect("Failed to canonicalize root");
 
     // Mount
-    mount(&mounts);
+    mount(&mounts).expect("Failed to mount");
```
flxo, line 71: See above.

northstar/src/runtime/island/init.rs
```diff
@@ -147,10 +147,11 @@ pub(super) fn init(
 }
 
 /// Execute list of mount calls
-fn mount(mounts: &[Mount]) {
+fn mount(mounts: &[Mount]) -> Result<(), ()> {
```
flxo, line 150: The `Result<...>` doesn't make any sense.

northstar/src/runtime/island/seccomp.rs
```diff
@@ -36,28 +36,19 @@ const AUDIT_ARCH: u32 = bindings::AUDIT_ARCH_AARCH64;
...
-            // Return an error here. Extend runtime::Error with an error: InvalidManifest
-            warn!(
-                "Failed to whitelist {}: {}. Disabling seccomp",
-                &name.to_string(),
-                e
-            );
-            return AllowList::default();
-        };
+            // Continue here as a missing syscall on the allow list does not lead to insecure behaviour
+            warn!("Failed to allow syscall {}: {}", &name.to_string(), e);
```
flxo, line 50: Thoughts: We have a couple of places where we just warn but continue. This is another one. One approach could be to add a `strict` mode in the configuration and fail in such a situation. e.g not existent bind bind or such a malformed seccomp config. On the other hand every thing should be ok also during test and currently the "warn and continue" behaviour is just handy for testing...but smells.

flxo (changes requested):

northstar/src/runtime/island/seccomp.rs
inorick: Sounds like an equivalent to the compiler flag `-Werror`. I think it is a good idea. If we add this later we should go over all the `warn!` statements and check if we want to do abort instead.

flxo (approved):

PR #369, opened 2021-06-22
Readd autostart option

flxo:
The autostart manifest flag got somewhere. Readd it.
Implement result handling of bulk mounting via the console.

### DIFF
--- a/northstar/src/runtime/error.rs
+++ b/northstar/src/runtime/error.rs
@@ -19,34 +19,26 @@ use thiserror::Error;
 
 #[derive(Error, Debug)]
 pub enum Error {
-    /// The container is not known to the system
     #[error("Invalid configuration: {0}")]
     Configuration(String),
-    /// The container is not known to the system
     #[error("Invalid container {0}")]
     InvalidContainer(Container),
-    /// The container cannot be started because it's already running
+    #[error("Container {0} cannot be mounted because it is already mounted")]
+    MountBusy(Container),
     #[error("Container {0} cannot be unmounted: busy")]
     UmountBusy(Container),
-    /// The container cannot be started because it's already running
     #[error("Container {0} failed to start: Already started")]
     StartContainerStarted(Container),
-    /// The container cannot be started because it's a resource container
     #[error("Container {0} failed to start: Resources cannot be started")]
     StartContainerResource(Container),
-    /// The container cannot be started because it's missing a running resource container
     #[error("Container {0} failed to start: Resource {1} is missing")]
     StartContainerMissingResource(Container, Container),
-    /// The container cannot be started
     #[error("Container {0} failed to start: {1}")]
     StartContainerFailed(Container, String),
-    /// The container cannot be started because it's already running
     #[error("Container {0} failed to stop: Not started")]
     StopContainerNotStarted(Container),
-    /// The container is not known to the system
     #[error("Invalid repository {0}")]
     InvalidRepository(RepositoryId),
-    /// The container is not known to the system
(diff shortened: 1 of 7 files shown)


northstar/src/runtime/error.rs
```diff
@@ -25,7 +25,10 @@ pub enum Error {
     /// The container is not known to the system
     #[error("Invalid container {0}")]
     InvalidContainer(Container),
-    /// The container cannot be started because it's already running
+    /// The container cannot be umounted because it is busy
```
inorick, line 28: Comment does not match error. Maybe we can even remove these comments as the error message is pretty clear.

northstar/src/runtime/state.rs
```diff
@@ -130,21 +134,44 @@ impl ProcessContext {
 impl<'a> State<'a> {
     /// Create a new empty State instance
     pub(super) async fn new(config: &'a Config, events_tx: EventTx) -> Result<State<'a>, Error> {
-        let mut repositories = Repositories::default();
+        let repositories = Repositories::default();
+        let mount_control = Arc::new(MountControl::new(&config).await.map_err(Error::Mount)?);
```
inorick, line 138: Why do we need an Arc here?

northstar/src/runtime/state.rs
```diff
@@ -166,24 +193,44 @@ impl<'a> State<'a> {
...
+        // Mount (parallel)
+        let mounts = self.mount_all(&autostart).await;
+
+        for (result, container) in mounts.iter().zip(autostart) {
+            match result {
+                Ok(_) => {
+                    info!("Autostarting {}", container);
+                    if let Err(e) = self.start(&container).await {
+                        warn!("Failed to autostart {}: {}", container, e);
+                        // TODO: What shall we do with this error?
```
inorick, line 223: I think autostart is done on a best-effort basis so printing a warning and continuing is sensible.

northstar/src/runtime/state.rs
```diff
@@ -297,49 +344,47 @@ impl<'a> State<'a> {
...
-                }
-            } else {
-                need_mount.insert(container.clone());
+        // Check if the container is not a resource
+        if npk.manifest().init.is_none() {
+            warn!("Container {} is a resource", container);
+            return Err(Error::StartContainerResource(container.clone()));
+        }
+
+        // The to be started container
```
inorick, line 359: "The container to be started"

northstar/src/runtime/state.rs
flxo: The `mount_control` is used in a `task::spawn` and needs to be cloned for that. Otherwise we would borrow `&self`.

northstar/src/runtime/state.rs
flxo: I don't agree. Containers flagged with `autostart` are something that is *not* started by someone and therefore probably *nobody* takes care about them. Image a "root" container that takes care for starting all other containers. If this root container doesn't start, the whole system is probably broken. 

What about extending the `autostart` flags:
```rust
enum Autostart {
  /// Ignore errors when starting this application. Ignore the containers termination
  Relaxed,
  /// If the container doesn't start or exits with a non zero exit code the runtime exits with a failure
  Critical,
  /// Treat a failure when starting this container as a critical incident and terminate the runtime. Ignore any termination o the container
  CriticalStart,
}
...
autostart: Autostart,
```

flxo: @marcmo @inorick any opinion?

flxo:
@inorick Added the options as discussed.

inorick (approved):

PR #370, opened 2021-06-22
Monitor runtime death signal from init

inorick:


northstar/src/runtime/island/mod.rs
```diff
@@ -64,20 +65,23 @@ type Container = MountedContainer<IslandProcess>;
...
         io: (Option<io::Log>, Option<io::Log>),
-        _dev: Dev,
         checkpoint: Checkpoint,
+        _dev: Dev,
     },
     Started {
         pid: Pid,
         exit_status: Box<dyn Future<Output = Result<ExitStatus, Error>> + Unpin + Send + Sync>,
         io: (Option<io::Log>, Option<io::Log>),
+        checkpoint: Checkpoint,
```
flxo, line 84: Think the `Checkpoint` isn't needed one the process is started. This can be removed here.

northstar/src/runtime/island/mod.rs
```diff
@@ -64,20 +65,23 @@ type Container = MountedContainer<IslandProcess>;
 pub(super) struct Island {
     tx: EventTx,
     config: Config,
+    tripwire_read: PipeRead, // Used by child process to detect if the parent process has died
```
flxo, line 68: Convert the comment to `///` above the field.

northstar/src/runtime/island/mod.rs
```diff
@@ -131,22 +143,20 @@ impl Launcher for Island {
             .as_ref()
             .map(|l| seccomp::seccomp_filter(l.iter()));
 
+        // do not close child checkpoint fd as it will be needed to detect if the runtime process died
```
flxo, line 146: `Do` instead of `do`

flxo:
@inorick The `Clone` that we added to `PipeRead` and `PipeWrite` is a bad idea since the shared resource might be closed on an unintentional drop. The stupid `derive` is a footgun. I wrapped the fd in an `Arc` which makes the last one closes the fd .

northstar/src/runtime/island/mod.rs
```diff
@@ -107,10 +112,20 @@ impl Launcher for Island {
     where
         Self: Sized,
     {
-        Ok(Island { tx, config })
+        let (tripwire_read, tripwire_write) =
+            block(pipe).map_err(|e| Error::io("Open tripwire pipe", e))?;
+        block(|| tripwire_read.set_cloexec(true))
```
inorick, line 117: This causes the fd to automatically be closed during execve, right? Why do we need this for our tripwire?

northstar/src/runtime/pipe.rs
```diff
@@ -22,87 +22,98 @@ use std::{
...
-    }
-}
-
-/// Write end of a pipe(2)
-#[derive(Debug)]
+/// Write end of a pipe(2). Last dropped clone closes the pipe
+#[derive(Clone, Debug)]
 pub(crate) struct PipeWrite {
-    fd: RawFd,
+    inner: Arc<Inner>,
```
inorick, line 90: This is much safer :+1:

northstar/src/runtime/island/mod.rs
```diff
@@ -107,10 +112,20 @@ impl Launcher for Island {
     where
         Self: Sized,
     {
-        Ok(Island { tx, config })
+        let (tripwire_read, tripwire_write) =
```
inorick, line 115: This causes the fd to automatically be closed during execve, right? Why do we need this for our tripwire?

flxo: We safe an explicit `close` due it should also happen on `drop` of `PipeWrite`.

PR #378, opened 2021-06-28
Forbid identical container name and version

flxo:
If the configured repositories contain a container with the same name
and version twice do not add both to the index.

### DIFF
--- a/northstar/src/runtime/repository.rs
+++ b/northstar/src/runtime/repository.rs
@@ -28,8 +28,8 @@ use log::{debug, info, warn};
 use mpsc::Receiver;
 use npk::npk;
 use std::{
-    collections::HashMap,
-    ffi::{CStr, OsStr},
+    collections::{HashMap, HashSet},
+    ffi::CStr,
     fmt,
     io::{BufReader, Seek, SeekFrom, Write},
     os::unix::prelude::{FromRawFd, RawFd},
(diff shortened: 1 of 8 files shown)


northstar/src/runtime/repository.rs
```diff
@@ -85,38 +89,46 @@ impl DirRepository {
...
 
         let results = join_all(loads).await;
         for result in results {
             match result {
                 Ok((container, file, npk)) => {
-                    containers.insert(container, (file, Arc::new(npk)));
+                    // If the container name/version is already in there remove the present
+                    // container and skip the newly parsed one.
+                    if containers.remove(&container).is_some() {
+                        warn!("Skipping duplicate container {}", container);
```
inorick, line 128: Isn't the presence of a duplicate container a fatal error and we should fail?

flxo: I thought about that, but it can leave Northstar in a situation where it's not able to start anymore. Think we had the "be pedantic or not" topic somewhere else.

inorick: That was the autostart behaviour which ended up being configurable. Not sure what the best way to go is here.

flxo: Yes. Right. Regarding the autostart thing it was easy to push the decision to the user.

Issue #379, opened 2021-06-28
Replace Croups memory oom polling with eventfd


flxo:
The cgroups memory controller provides an eventfd interface for oom events. Currently the oom condition is polled with 2Hz. Using an eventfd uses less resources since it's just two open fds and a read.

Issue #365, opened 2021-06-10
Fix cgroup memory integration test


flxo:
The cgroup memory integration test is broken and disabled because it had permanent errors in GitHub actions. The test needs to be reenabled and probably the configuration slightly tweaked. The test runs fine on ordinary boxes and therefore should also in GH.

flxo:
Fixed by c8a678a

PR #381, opened 2021-06-29
Refactor cgroup memory event polling with eventfd

flxo:
Replace the polling oom task with a wait for an eventfd signal from the
cgroup memory controler. Activate cgroups integration test.

Fixes: #379 #365

### DIFF
--- a/northstar/src/runtime/cgroups.rs
+++ b/northstar/src/runtime/cgroups.rs
@@ -16,14 +16,26 @@ use super::{config, Container, Event, EventTx, Pid};
 use log::{debug, warn};
 use npk::manifest;
 use proc_mounts::MountIter;
-use std::path::{Path, PathBuf};
+use std::{
+    os::unix::io::AsRawFd,
+    path::{Path, PathBuf},
+};
 use thiserror::Error;
-use tokio::{fs, io, select, task, time};
+use tokio::{
+    fs,
+    io::{self, AsyncReadExt, AsyncWriteExt},
+    select,
+    sync::mpsc::error::TrySendError,
+    task::{self, JoinHandle},
+    time,
+};
+use tokio_eventfd::EventFd;
 use tokio_util::sync::CancellationToken;
 
 const OOM_CONTROL: &str = "memory.oom_control";
-const UNDER_OOM: &str = "under_oom 1";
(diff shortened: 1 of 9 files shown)


northstar/src/runtime/cgroups.rs
```diff
@@ -63,104 +134,144 @@ impl CGroups {
...
 /// Get the cgroup v1 controller hierarchy mount point
 fn mount_point(controller: &str) -> Result<PathBuf, Error> {
-    let controller = controller.to_owned();
-    MountIter::new()
-        .map_err(Error::MountInfo)?
-        .filter_map(Result::ok)
-        .find(|m| m.fstype == "cgroup" && m.options.contains(&controller))
-        .map(|m| m.dest)
-        .ok_or(Error::UnknownController(controller))
+    task::block_in_place(|| {
```
alfonsoros88, line 269: is this needed even if the function is not `async`?

alfonsoros88 (commented):
looks good to me :) the event_control thing is interesting

northstar/src/runtime/cgroups.rs
```diff
@@ -63,104 +134,144 @@ impl CGroups {
...
 
-    pub async fn destroy(self) -> Result<(), Error> {
-        self.stop.cancel();
-        for cgroup_dir in self.groups {
-            debug!("Destroying CGroup {}", cgroup_dir.display());
-            fs::remove_dir(&cgroup_dir)
-                .await
-                .map_err(|e| Error::Destroy(cgroup_dir.display().to_string(), e))?;
+    pub async fn destroy(mut self) {
+        for group in self.groups.drain(..) {
```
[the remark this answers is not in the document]
flxo: Yes. Thanks!

northstar/src/runtime/cgroups.rs
flxo: In this case yes because the fn is called from an async context (as anything inside northstar). I'd wish there would be some kind of annotation of in best case automatic detection of blocking things. Think the async-std guys tried something like that.

flxo:
Anything else @alfonsoros88 ? Otherwise I'd be happy to merge this.

alfonsoros88:
@flxo no, looks good! :+1: :)

PR #383, opened 2021-07-02
Avoid leaking temp dirs in tests

### LINKED ISSUE #382: NPK tests leak tempdirs
After a clean build and `cargo test -p npk` the following tempdirs are left behind. This should not be the case.

The problem is 

```rust
    fn create_tmp_dir() -> PathBuf {
        tempfile::TempDir::new()
            .expect("Create tmp dir")
            .into_path()
```
 
because `TempDir::into_path` consumes the TempDir without deleting directory on the filesystem, meaning that the directory will no longer be automatically deleted.

https://docs.rs/tempfile/3.2.0/tempfile/struct.TempDir.html#method.into_path

```
› ls /tmp/.tmp*
/tmp/.tmp954evw:
test_key.key  test_key.pub

/tmp/.tmpAFVnzK:
manifest.yaml

/tmp/.tmpAwHRIu:
different_manifest_name.yaml

/tmp/.tmpb5Rifc:
hello-0.0.2.npk

/tmp/.tmpg8Zurv:
manifest.yaml

/tmp/.tmpGm6nEn:
fs.img  manifest.yaml  signature.yaml  squashfs-root

/tmp/.tmpiF5f5P:
test_key.key  test_key.pub

/tmp/.tmpInTCsh:
test_key.key  test_key.pub

/tmp/.tmpjMlCtF:
test_key.key  test_key.pub

/tmp/.tmpmHctYU:
hello-0.0.2.npk

/tmp/.tmpMymp9i:
test_key.key  test_key.pub

/tmp/.tmpOIqmOT:
manifest.yaml

/tmp/.tmpOOW28K:
test_key.key  test_key.pub

/tmp/.tmpTY4fdy:

/tmp/.tmpucKZ12:

/tmp/.tmpvhzWYF:
hello-0.0.2.npk

/tmp/.tmpViGGIj:
test_key.key  test_key.pub

/tmp/.tmpvY7cWk:
hello-0.0.2.npk

/tmp/.tmpzvbhR7:
manifest.yaml
```

inorick:
Fixes #382

### DIFF
--- a/npk/tests/npk.rs
+++ b/npk/tests/npk.rs
@@ -40,16 +40,16 @@ env:
   HELLO: north
 ";
 
-    fn create_test_npk(dest: &PathBuf, manifest_name: Option<&str>) -> PathBuf {
+    fn create_test_npk(dest: &Path, manifest_name: Option<&str>) -> PathBuf {
         let src = create_tmp_dir();
         let key_dir = create_tmp_dir();
-        let manifest = create_test_manifest(&src.path().to_path_buf(), manifest_name);
-        let (_pub_key, prv_key) = gen_test_key(&key_dir.path().to_path_buf());
-        pack(&manifest, &src.path().to_path_buf(), &dest, Some(&prv_key)).expect("Pack NPK");
+        let manifest = create_test_manifest(&src.path(), manifest_name);
+        let (_pub_key, prv_key) = gen_test_key(&key_dir.path());
+        pack(&manifest, &src.path(), &dest, Some(&prv_key)).expect("Pack NPK");
         dest.join("hello-0.0.2.npk")
     }
 
-    fn create_test_manifest(dest: &PathBuf, manifest_name: Option<&str>) -> PathBuf {
+    fn create_test_manifest(dest: &Path, manifest_name: Option<&str>) -> PathBuf {
         let manifest = dest
             .join(manifest_name.unwrap_or("manifest"))
             .with_extension("yaml");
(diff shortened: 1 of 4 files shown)


npk/tests/npk.rs
```diff
@@ -39,12 +40,12 @@ env:
   HELLO: north
 ";
 
-    fn create_test_npk(dest: &Path, manifest_name: Option<&str>) -> PathBuf {
+    fn create_test_npk(dest: &PathBuf, manifest_name: Option<&str>) -> PathBuf {
```
flxo, line 43: Shouldn't this be `&Path`?  Think there's never a good reason to pass a immutable `&PathBuf`.

tools/sextant/src/inspect.rs
```diff
@@ -141,12 +142,12 @@ mounts:
...
-    fn create_test_npk(dest: &Path) -> PathBuf {
+    fn create_test_npk(dest: &PathBuf) -> PathBuf {
         let src = create_tmp_dir();
         let key_dir = create_tmp_dir();
-        let manifest = create_test_manifest(&src);
-        let (_pub_key, prv_key) = gen_test_key(&key_dir);
-        pack(&manifest, &src, &dest, Some(&prv_key)).expect("Pack NPK");
+        let manifest = create_test_manifest(&src.path().to_path_buf());
+        let (_pub_key, prv_key) = gen_test_key(&key_dir.path().to_path_buf());
+        pack(&manifest, &src.path().to_path_buf(), &dest, Some(&prv_key)).expect("Pack NPK");
```
flxo, line 150: `pack` takes a `&Path`. Why creating a `PathBuf` and converting it to `&Path` again? Think `src.path()` should be fine because of https://docs.rs/tempfile/3.2.0/tempfile/struct.TempDir.html#method.path already returns `&Path`.

flxo (changes requested):
Thanks for the patch. 
I think there's not need for `PathBufs` anywhere here. The `TempDir` is always kept in scope and passing `&Path` from `TempDir::path` should be the cheapest.

PR #384, opened 2021-07-02
Move run_dir bind mount to main

flxo:
The self mounted run_dir is something integration specific and therfore
should not be done withing the runtime lib. When running northstar for
example as an android service all this ceremony is not needed and done
by Android init.

Move responsibility of the configured directories also out of the lib
into it's user but keep the checks whether the dirs are read and
writeable for the runtime.

Dump backtraces on panics in tests.

### DIFF
--- a/northstar_tests/src/macros.rs
+++ b/northstar_tests/src/macros.rs
@@ -12,20 +12,49 @@
 //   See the License for the specific language governing permissions and
 //   limitations under the License.
 
+use super::logger;
+use nix::{mount, sched};
+use sched::{unshare, CloneFlags};
+use std::env;
+
 pub fn init() {
     color_eyre::install().unwrap();
-    crate::logger::init();
+    logger::init();
     log::set_max_level(log::LevelFilter::Debug);
 
     // TODO make the test independent of the workspace structure
     // set the CWD to the root
-    std::env::set_current_dir("..").unwrap();
+    env::set_current_dir("..").unwrap();
 
     // Enter a mount namespace. This needs to be done before spawning
     // the tokio threadpool.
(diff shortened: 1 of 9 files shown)


northstar_tests/src/macros.rs
```diff
@@ -12,18 +12,44 @@
...
     // Enter a mount namespace. This needs to be done before spawning
     // the tokio threadpool.
-    nix::sched::unshare(nix::sched::CloneFlags::CLONE_NEWNS).unwrap();
+    unshare(CloneFlags::CLONE_NEWNS).unwrap();
+
+    // Enable backtrace dumping
+    env::set_var("RUST_BACKTRACE", "1");
+
+    // Set the mount propagation to private on root. This ensures that *all* mounts
+    // the tests to get cleaned up upon process termination. The approach to bind mount
```
inorick, line 37: "the tests to" left over from copy/paste?

northstar_tests/src/macros.rs
```diff
@@ -12,18 +12,44 @@
...
+    // the run_dir only (this is where the mounts from northstar happen) doesn't work
+    // for the tests since the run_dir is a tempdir which is a random dir on every run.
+    // Checking at the beginning of the tests if run_dir is bind mounted - a leftover from
+    // a previous crash obviously doesn't work.
```
inorick, line 41: "crash obviously" -> "crash - obviously"

northstar_tests/src/macros.rs
```diff
@@ -12,18 +12,44 @@
...
+    // Technically there's only the need to the propagation on the parent mount of the run_dir
```
inorick, line 42: "Technically, it is only necessary to propagate $WHAT to the parent mount of the run_dir"

PR #386, opened 2021-07-02
Support file as sextant output parameter

### LINKED ISSUE #364: Sextant: Adjust out parameter: file instead of directory
Change the interface of `sextant` so that the `--out` parameter points to a file where the packed npk is dumped to. Currently `--out` specifies a directory and generates the npk filename from the manifest.

An options could be to keep the current behaviour of the parameter of `--out` is a directory that exists.

inorick:
Fixes #364

### DIFF
--- a/npk/src/npk.rs
+++ b/npk/src/npk.rs
@@ -447,7 +447,7 @@ impl Default for SquashfsOpts {
 /// # Arguments
 /// * `manifest` - Path to the container's manifest file
 /// * `root` - Path to the container's root directory
-/// * `out` - Directory where the resulting NPK will be written to
+/// * `out` - Target directory or filename of the packed NPK
 /// * `key` - Path to the key used to sign the package
 ///
 /// # Example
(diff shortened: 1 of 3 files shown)


npk/src/npk.rs
```diff
@@ -501,9 +501,13 @@ pub fn pack_with(
     }
     builder = builder.squashfs_opts(&squashfs_opts);
 
-    let npk_dest = out
-        .join(format!("{}-{}.", &name, &version.to_string()))
-        .with_extension(&NPK_EXT);
+    let mut npk_dest = out.to_path_buf();
+    if Path::is_dir(out) {
+        // Append filename from manifest if only a directory path was given
+        npk_dest = out
```
flxo, line 507: Easier to read would be:

```
npk_dest.push(format!(...));

flxo (commented):
Please adjust the help text of the parameter `out` of the `pack` subcommand.
Rest is fine!

flxo (approved):

Issue #367, opened 2021-06-17
Manifest verification


flxo:
The `serde` code for parsing the manifests is pretty ok but misses some semantic checks:

Uncomplete list of needed checks:

- [x] `env`, `init`, `args` may not not contains any null byte - needs to be convertible to a `CString`
- [x] mount points may not overlap
- [x] `name`, `version`, `uid` and `gid` are mandatory
- [x] `args`, `env` and others are not valid if `init` equals `None`
- [x] croups names must be convertible to `CString`
- [x] `uid` range
- [x] `gid` range
- [x] character range for all items especially `name`

flxo:
See #208

inorick:
Implemented by #387

PR #387, opened 2021-07-05
Verify manifest name

inorick:
This PR implements the list of checks in #367 

Additional check:
Relative mount points are not allowed.

Open TODO:
Move new types to own file.

### DIFF
--- a/npk/src/manifest.rs
+++ b/npk/src/manifest.rs
@@ -37,7 +37,7 @@ pub struct Name(String);
 
 #[derive(Error, Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
 #[error("Invalid character(s) in name")]
-pub struct NameError;
+pub struct InvalidNameChar(usize);
 
 impl Display for Name {
     fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), std::fmt::Error> {
@@ -54,22 +54,21 @@ impl Deref for Name {
 }
 
 impl TryFrom<String> for Name {
-    type Error = NameError;
+    type Error = InvalidNameChar;
 
     fn try_from(value: String) -> Result<Self, Self::Error> {
-        if value
-            .chars()
-            .all(|c| matches!(c, '0'..='9' | 'A'..='Z' | 'a'..='z' | '.' | '_' | '-'))
+        if let Some(pos) =
+            value.find(|c: char| !matches!(c, '0'..='9' | 'A'..='Z' | 'a'..='z' | '.' | '_' | '-'))
         {
-            Ok(Name(value))
+            Err(InvalidNameChar(pos))
         } else {
-            Err(NameError)
+            Ok(Name(value))
         }
     }
 }
 
 impl TryFrom<&str> for Name {
-    type Error = NameError;
+    type Error = InvalidNameChar;
 
     fn try_from(value: &str) -> Result<Self, Self::Error> {
         value.to_string().try_into()
@@ -79,6 +78,12 @@ impl TryFrom<&str> for Name {
 impl Name {
     pub fn to_str(&self) -> &str {
         self.0.as_str()
+    }
+}
+
+impl InvalidNameChar {
+    pub fn nul_position(&self) -> usize {
+        self.0
     }
 }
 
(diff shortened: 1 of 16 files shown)


npk/src/manifest.rs
```diff
@@ -12,23 +12,108 @@
...
+}
+
+// TODO: move out of manifest.rs
+#[derive(Clone, Eq, PartialOrd, PartialEq, Debug, Hash, Serialize, Deserialize)]
+pub struct NonNullString(String);
+
+#[derive(Error, Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
+pub enum NonNullStringError {
+    #[error("Invalid null byte in string")]
+    NonNullStringError,
```
flxo, line 79: The stdlib names the same `InteriorNul` and adds the position. Check std::ffi::CString` Can we do the same? In fact we have almost the same restrictions as the `CString` but don't want to use it for discussed reasons (the null termination sucks).

npk/src/manifest.rs
```diff
@@ -94,38 +179,58 @@ impl Manifest {
...
+        }
+
+        // Check for relative bind mounts
+        let sorted_bind_mounts: Vec<&PathBuf> = self
+            .mounts
+            .iter()
+            .filter(|(_, m)| matches!(m, Mount::Bind(_)))
+            .map(|(p, _)| p)
+            .sorted()
+            .collect();
```
flxo, line 215: We can safe the collect by directly iterating over the iterator returned from `sorted` and do the below relative check in there as well.

npk/src/manifest.rs
```diff
@@ -458,17 +563,17 @@ seccomp:
...
+        assert_eq!(args[0].to_string(), "one");
+        assert_eq!(args[1].to_string(), "two");
 
         assert_eq!(manifest.autostart, Some(Autostart::Critical));
         let env = manifest.env.ok_or_else(|| anyhow!("Missing env"))?;
         assert_eq!(
-            env.get("LD_LIBRARY_PATH"),
-            Some("/lib".to_string()).as_ref()
+            env.get(&"LD_LIBRARY_PATH".to_string().try_into()?),
+            Some("/lib".to_string().try_into()?).as_ref()
```
flxo, line 576: There are a couple of place where a 

`TryFrom<&str> for NotNullString` would be handy.

or even better a 

`impl<T> TryFrom<T> for NotNullString where T: ToString`

npk/src/manifest.rs
```diff
@@ -12,23 +12,108 @@
...
+    ops::Deref,
+    path::{Component, Component::RootDir, PathBuf},
     str::FromStr,
 };
 use thiserror::Error;
 
-pub type Name = String;
+// TODO: move out of manifest.rs
+#[derive(Clone, Eq, PartialOrd, PartialEq, Debug, Hash, Serialize, Deserialize)]
+pub struct Name(String);
```
flxo, line 35: Can we add a 

```rust
  fn to_str(&self) -> &str;
```
?

flxo (commented):
Cool!

npk/src/manifest.rs
inorick: We need `sorted_bind_mounts` for two checks: The relative mount check and the overlapping mount check. That's why it is saved.

flxo: But both checks can be done with one iteration. This would safe the additional copy.

npk/src/manifest.rs
inorick: The second version conflicted with a definition from core. I implemented the first variant.

flxo: Ah. Yes both traits are upstream.

npk/src/manifest.rs
inorick: Done

npk/src/manifest.rs
```diff
@@ -539,7 +680,50 @@ mounts:
...
+    }
 
+    /// Overlapping mounts are invalid
+    #[test]
+    fn overlapping_mount() -> Result<()> {
+        let manifest = "name: hello\nversion: 0.0.0\ninit: /binary\nuid: 1000\ngid: 1001
+mounts:
+  /lib/overlapping:
+    type: bind
+    host: /lib
```
marcmo, line 693: would be good also to test the other conditions (Two mount paths both starting with '/', ...)

PR #389, opened 2021-07-05
Simplify client by removing tasks and channels

flxo:
The internal channels in the client for responses and notifications are not needed and make it impossible to report error conditions to the user of the lib. Remove the task in between the connection to the runtime and the client API.

### DIFF
--- a/northstar/src/api/client.rs
+++ b/northstar/src/api/client.rs
@@ -13,34 +13,32 @@
 //   limitations under the License.
 
 use super::{
-    codec::{framed, Framed},
+    codec::{self, framed},
     model::{
-        self, Connect, Container, ContainerData, Message, MountResult, Notification, Payload,
-        RepositoryId, Request, Response,
+        self, Connect, Container, ContainerData, Message, MountResult, Notification, RepositoryId,
+        Request, Response,
     },
 };
 use futures::{SinkExt, Stream, StreamExt};
-use log::{debug, info};
+use log::debug;
 use npk::manifest::Version;
 use std::{
-    collections::HashSet,
-    path::{Path, PathBuf},
+    collections::{HashSet, VecDeque},
+    path::Path,
     pin::Pin,
     task::Poll,
 };
 use thiserror::Error;
 use tokio::{
-    fs,
-    io::{self, AsyncRead, AsyncWrite},
+    fs, io,
     net::{TcpStream, UnixStream},
-    select,
-    sync::{mpsc, oneshot},
-    task, time,
+    time,
 };
+use tokio_util::either::Either;
 use url::Url;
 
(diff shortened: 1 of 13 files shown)


northstar/src/api/codec.rs
```diff
@@ -153,35 +154,35 @@ impl<T: Unpin + AsyncRead + AsyncWrite> futures::sink::Sink<model::Message> for
...
+
+//     fn mk_message() -> impl Strategy<Value = model::Message> {
+//         mk_simple_payload().prop_map(model::Message::Request(model::Request::Containers))
+//     }
+
+//     fn mk_simple_payload() -> impl Strategy<Value = model::Payload> {
+//         // TODO: Perhaps consider all the variants?
+//         Just(model::Payload::Request(model::Request::Containers))
+//     }
+// }
```
marcmo, line 188: should not be part of this commit. and should remain in, this is a good test

northstar/src/api/client.rs
```diff
@@ -489,7 +484,39 @@ impl<'a> Client {
...
+    fn fuse(&mut self) {
+        self.fused = true;
+    }
+
+    /// Return Error::Stopped if the client is fused
+    fn fused(&self) -> Result<(), Error> {
+        if self.fused {
+            Err(Error::Stopped)
+        } else {
+            Ok(())
```
marcmo, line 519: using the term `fuse` here is really confusing!

northstar/src/api/client.rs
```diff
@@ -156,78 +155,16 @@ impl<'a> Client {
...
-
-        let mut connection = time::timeout(timeout, Self::connect(url, notifications, timeout))
+        let connection = time::timeout(timeout, Self::connect(url, notifications, timeout))
             .await
             .map_err(|_| Error::Timeout)??;
 
         debug!("Connected to {}", url);
 
-        task::spawn(async move {
-            loop {
```
marcmo, line 171: removing the task indeed simplifies the code...nice

marcmo: but how do notifications work when no request is pending?

northstar/src/api/codec.rs
flxo: Ah. The proptest api confused me and I forgot the readd it. Basically this tests `serde`.

northstar/src/api/client.rs
flxo: It's borrowed from `futures`: https://docs.rs/futures/0.3.15/futures/future/trait.FutureExt.html#method.fuse.

northstar/src/api/client.rs
flxo: The `Stream` impl of `Client` checks the  notification buffer and if that's empty it polls the connection.

PR #392, opened 2021-07-08
Move npk crate into northstar

inorick:
Move contents of npk crate to a new northstar module.
Move Container type from api module to new module.
Remove hello-world container from build.rs as the required pack function is not available there. If necessary, it can be reintroduced on several different ways.


northstar-tests/Cargo.toml
```diff
@@ -25,4 +24,4 @@ uuid = { version = "0.8.2", features = ["v4"] }
 [build-dependencies]
 tempfile = "3.2"
 escargot = "0.5.2"
-npk = { path = "../npk" }
+northstar = { path = "../northstar", features = ["runtime"] }
```
flxo, line 27: Shouldn't this need also feature `npk` because the `test-container` is built from the `build.rs`?

northstar/Cargo.toml
```diff
@@ -9,34 +9,44 @@ license-file = "LICENSE.md"
 [dependencies]
 async-stream = { version = "0.3.0", optional = true }
 async-trait = { version = "0.1.50", optional = true }
+base64 = "0.13.0"
```
flxo, line 12: This should be optional and depended by the feature `npk`.

northstar/Cargo.toml
```diff
@@ -9,34 +9,44 @@ license-file = "LICENSE.md"
...
 bincode = { version = "1.3", optional = true }
 bitflags = { version = "1.2", optional = true }
+byteorder = "1.4"
```
flxo, line 15: This should be optional and depended by the feature `npk`.

northstar/Cargo.toml
```diff
@@ -9,34 +9,44 @@ license-file = "LICENSE.md"
...
 bytes = { version = "1.0", optional = true }
 bytesize = { version = "1.0", optional = true }
 caps = { version = "0.5.1", optional = true }
 derive-new = { version = "0.5.9", optional = true }
 ed25519-dalek = { version = "1.0", optional = true }
 floating-duration = { version = "0.1.2", optional = true }
+fs_extra = "1.2"
```
flxo, line 22: This should be optional and depended by the feature `npk`.

northstar/Cargo.toml
```diff
@@ -9,34 +9,44 @@ license-file = "LICENSE.md"
...
 page_size = { version = "0.4.2", optional = true }
 proc-mounts = { version = "0.2.4", optional = true }
 procinfo = { version = "0.4.2", optional = true }
+rand = "0.7.3"
+semver = { version = "0.11.0", features = ["serde"] }
 serde = { version = "1.0", features = ["derive", "rc"] }
 serde_json = { version = "1.0", optional = true }
+serde_with = "1.9"
+serde_yaml = "0.8.17"
+sha2 = "0.9.3"
```
flxo, line 40: optional?

northstar/Cargo.toml
```diff
@@ -9,34 +9,44 @@ license-file = "LICENSE.md"
...
 lazy_static = { version = "1.4", optional = true }
 log = { version = "0.4.14", features = [ "serde"] }
 memoffset = { version = "0.6.3", optional = true }
 nix = { version = "0.20.0", optional = true }
-npk = { path = "../npk", optional = true }
 page_size = { version = "0.4.2", optional = true }
 proc-mounts = { version = "0.2.4", optional = true }
 procinfo = { version = "0.4.2", optional = true }
+rand = "0.7.3"
+semver = { version = "0.11.0", features = ["serde"] }
```
flxo, line 35: optional?

northstar/Cargo.toml
```diff
@@ -9,34 +9,44 @@ license-file = "LICENSE.md"
...
+serde_yaml = "0.8.17"
+sha2 = "0.9.3"
 tempfile = { version = "3.2", optional = true }
 thiserror = "1.0"
 tokio = { version = "1.5", features = ["full"] }
 tokio-eventfd = { version = "0.2.0", optional = true }
 tokio-util = { version = "0.6.6", features = ["codec", "io"], optional = true }
 url = { version = "2.2", features = ["serde"] }
 uuid = { version = "0.8.2", features = ["v4"], optional = true }
+which = "4.1"
```
flxo, line 48: This should be optional and depended by the feature `npk`.

northstar/Cargo.toml
```diff
@@ -45,8 +55,8 @@ proptest = "1.0"
 
 [features]
 default = []
-api = ["bytes", "derive-new", "npk", "serde_json", "tokio-util"]
-hello-world = []
+api = ["bytes", "derive-new", "npk", "serde_json", "tokio-util", "runtime"]
+npk = []
```
flxo, line 59: The npk feature must have some deps that are not needed for the other features. e.g `which`

northstar/src/common/name.rs
```diff
@@ -0,0 +1,64 @@
+use serde::{Deserialize, Serialize};
```
flxo, line 1: Copyright header is missing.

northstar/src/common/non_null_string.rs
```diff
@@ -0,0 +1,67 @@
+use serde::{Deserialize, Serialize};
```
flxo, line 1: Copyright header is missing

flxo (changes requested):

northstar-tests/Cargo.toml
inorick: I replaced "runtime" with "npk" as runtime was not really needed.

main/Cargo.toml
```diff
@@ -26,5 +26,4 @@ env_logger = "0.8.3"
 android-logd-logger = "0.1.3"
 
 [features]
-default = []
-hello-world = ["northstar/hello-world"]
\ No newline at end of file
+default = []
```
flxo: Think a empty default set can just be removed along with the whole `[features]` section.

northstar/Cargo.toml
```diff
@@ -9,34 +9,44 @@ license-file = "LICENSE.md"
...
 bytes = { version = "1.0", optional = true }
 bytesize = { version = "1.0", optional = true }
 caps = { version = "0.5.1", optional = true }
 derive-new = { version = "0.5.9", optional = true }
 ed25519-dalek = { version = "1.0", optional = true }
 floating-duration = { version = "0.1.2", optional = true }
+fs_extra = { version = "1.2", optional = true }
 futures = "0.3.14"
 hex = { version = "0.4.3", optional = true }
+humanize-rs = "0.1.5"
```
flxo, line 25: This should be also optional. Only npk feature depends on it. It's used for parsing the tmpfs size in the manifest.

northstar/Cargo.toml
```diff
@@ -9,34 +9,44 @@ license-file = "LICENSE.md"
...
+sha2 = { version = "0.9.3", optional = true }
 tempfile = { version = "3.2", optional = true }
 thiserror = "1.0"
 tokio = { version = "1.5", features = ["full"] }
 tokio-eventfd = { version = "0.2.0", optional = true }
 tokio-util = { version = "0.6.6", features = ["codec", "io"], optional = true }
 url = { version = "2.2", features = ["serde"] }
 uuid = { version = "0.8.2", features = ["v4"], optional = true }
+which = { version = "4.1", optional = true }
+zip = { version = "0.5.13", features = ["unreserved"] }
```
flxo, line 49: Zip is also optional and just used by `npk`.

main/Cargo.toml
inorick: I deleted `default` but how can we get rid of the whole `features` section? Those features are references in `lib.rs`. Should we get rid of these features everywhere?

flxo (approved):
Nice!

PR #396, opened 2021-07-14
Replace devicemapper and loopdev fork

flxo:
The devicemapper and loopdev crate got updated with support for all
features and targets northstar needs. Remove loopdev and dm code and
make use of the upstream crates.


flxo:
This is blocked by the build failure of `devicemapper-rs` for `must`.

PR #399, opened 2021-07-22
Various stability and cleanup modifications

flxo:
* Remove not needed `mut` of seccomp allow list
* Add `/dev` mount to `test-container`
* Deactivate cgroup memory test. The removal of the group is unstable and needs a fix
* Add `rerun-if-changed` output of test `build.rs`
* Remove tempdir leak in integration tests
* Accept a SIGABRT in integrations test because the test-container tends to abort if it receives a SIGTERM in it's early startup. This happens *after* the `execve` - so it's fine
* Cleanup of the `pipe` module
* Remove `signal-hook` from `inspect` example because this swallows signals when signalled at the initialisation phase
* Remove `Resources` from `ContainerData` because those values just reflect `init` - will be replaced with a cgroup based solution

### DIFF
--- a/northstar-tests/test-container/manifest.yaml
+++ b/northstar-tests/test-container/manifest.yaml
@@ -3,10 +3,10 @@ version: 0.0.1
 init: /test-container
 uid: 1000
 gid: 1000
-cgroups:
-  memory:
-    limit_in_bytes: 10000000
-    swappiness: 0
+# cgroups:
+#   memory:
+#     limit_in_bytes: 10000000
+#     swappiness: 0
 capabilities:
   - CAP_KILL
 mounts:
--- a/northstar-tests/test-container/manifest.yaml
+++ b/northstar-tests/test-container/manifest.yaml
@@ -10,6 +10,8 @@ cgroups:
 capabilities:
   - CAP_KILL
 mounts:
+  /dev:
+    type: dev
   /data:
     type: persist
   /lib:
--- a/northstar-tests/tests/tests.rs
+++ b/northstar-tests/tests/tests.rs
@@ -84,12 +84,12 @@ test!(start_stop_test_container_without_waiting, {
 
     for _ in 0..10u32 {
         runtime.start(TEST_CONTAINER).await?;
-        runtime.stop(TEST_CONTAINER, 1).await?;
-        assume(
-            "Stopped test-container:0.0.1 with status Signaled\\(SIGTERM\\)",
-            5,
-        )
-        .await?;
+        runtime.stop(TEST_CONTAINER, 5).await?;
+        // TODO: For an unknown reason the child processes exit with a SIGABRT instead
+        // of SIGTERM. The abort is generated in the child process *after* the execve
+        // which makes it impossible to handle by init or the runtime. Accept any exit
+        // status here, as long as the container exits at all.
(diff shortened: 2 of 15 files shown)


northstar-tests/tests/tests.rs
```diff
@@ -396,30 +396,30 @@ test!(check_api_version_on_connect, {
...
-        assume(
-            "Stopped test-container:0.0.1 with status Signaled\\(SIGTERM\\)",
-            10,
-        )
-        .await?;
-    }
-
-    runtime.shutdown().await
-});
+// test!(cgroups_memory, {
```
inorick, line 399: Better add a TODO comment so we remember, why this was deactivated.

northstar-tests/test-container/manifest.yaml
```diff
@@ -3,13 +3,15 @@ version: 0.0.1
 init: /test-container
 uid: 1000
 gid: 1000
-cgroups:
-  memory:
-    limit_in_bytes: 10000000
-    swappiness: 0
+# cgroups:
```
inorick, line 6: Better add a TODO comment so we remember, why this was deactivated.

northstar/src/runtime/island/mod.rs
```diff
@@ -477,53 +496,65 @@ fn seccomp_filter(container: &Container) -> Option<seccomp::AllowList> {
...
-#[derive(Debug, Serialize, Deserialize, PartialEq)]
-enum Start {
-    // Signal the child to go
-    Start,
-    // Signal the parent that go is received
-    Started,
-    // Signal the child that the parent has terminated
-    Died,
+/// Block all signals of this process and currenty thread
+fn signals_block() {
```
inorick, line 500: Blocking those signals until we can process them looks like a cleaner solution than the one we had before :+1:

northstar/src/runtime/island/seccomp.rs
```diff
@@ -89,7 +89,7 @@ impl AllowList {
 
         let sf_prog = sock_fprog {
             len: self.list.len() as u16,
-            filter: self.list.as_mut_ptr(),
+            filter: self.list.as_ptr() as *mut bindings::sock_filter,
```
inorick, line 92: How can a non-mut pointer be cast into a mut pointer like that?

northstar-tests/tests/tests.rs
flxo: Documented in #402

northstar-tests/test-container/manifest.yaml
flxo: Documented in #402

northstar/src/runtime/island/seccomp.rs
flxo: [Here](https://internals.rust-lang.org/t/what-is-the-real-difference-between-const-t-and-mut-t-raw-pointers/6127/3) is the full story about it.

inorick: If I understand it correctly, calling `apply()` twice would now cause two mutable pointers to `self.list` to exist. It sounds similar to this bug in `as_mut_ref` that was fixed in Rust 1.15.1: https://blog.rust-lang.org/2017/02/09/Rust-1.15.1.html

They went ahead and changed the parameter type to `&mut self`.

PR #408, opened 2021-07-28
Add support for seccomp arguments

inorick:
Syscall arguments can now be filtered. This can be done in two ways:

1. List all allowed values for a specific syscall argument in the manifest.
For example, the following manifest entry allows the second argument of the `access` syscall to be 4. If `access` is called with any other argument value, the container is terminated.
```
seccomp:
  allow:
    access:
      args:
        index: 1
        values: [
            4,
        ]
```
      
2. Define a bitmask of allowed syscall argument bits for a specific syscall argument in the manifest.
For example, the following manifest entry allows the first argument of the `sysfs` syscall to be 0, 1, 4 or 5.
```
seccomp:
  allow:
    sysfs:
      args:
        index: 1
        mask: 0x05
```


northstar/src/npk/manifest.rs
```diff
@@ -281,7 +316,30 @@ pub struct Seccomp {
...
-    pub allowlist: Option<HashSet<String>>,
+    pub allow: Option<HashMap<NonNullString, SyscallRule>>,
+}
+
+#[derive(Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
+//#[serde(untagged)]
+pub enum SyscallRule {
+    /// All syscall arguments are allowed
+    #[serde(rename = "all")]
+    All,
```
flxo, line 327: Think `any` is a better fit than `all`

northstar/src/runtime/island/mod.rs
```diff
@@ -113,7 +113,7 @@ impl Island {
         let (checkpoint_runtime, checkpoint_init) = checkpoints();
         let groups = groups(&manifest);
         let capabilities = capabilities(&manifest);
-        let seccomp = seccomp_filter(&manifest);
+        let seccomp = seccomp_filter(&manifest).expect("Failed to create seccomp filter");
```
flxo, line 116: What would be a potential error here? Is this fine to `expect`?

northstar/src/runtime/island/seccomp.rs
```diff
@@ -13,18 +13,31 @@
...
+};
 use thiserror::Error;
 
+// Required for platform check
+#[cfg(any(
+    not(any(target_arch = "aarch64", target_arch = "x86_64")),
+    target_pointer_width = "32",
+    target_endian = "big"
+))]
+use crate::runtime::island::seccomp::Error::UnsupportedPlatform;
```
flxo, line 39: Can we make this a compile time error?

https://doc.rust-lang.org/std/macro.compile_error.html

northstar-tests/test-container/src/main.rs
```diff
@@ -122,6 +130,15 @@ fn leak_memory() {
     }
 }
 
+/// Call the 'delete_module' syscall with an empty module name. This has no effect and just returns -1.
```
flxo, line 133: Nice.

northstar/src/runtime/island/seccomp.rs
```diff
@@ -35,38 +48,66 @@ mod bindings {
     pub const SECCOMP_RET_LOG: u32 = 2147221504;
```
flxo, line 48: @inorick Do you remember why we don't get `SECCOMP_RET_LOG` is not included in the generated bindings?

flxo (commented):
Wow. Looks like a hard piece of word. I do not understand the BPF internals but despite the couple of nitpicks everything around that looks good to me.

northstar/src/runtime/island/seccomp.rs
inorick: I think you added that in fb430ec51dd34bd271bcae60a4d3e0f75ccfc77b.

I do not know more than what the comment says:
> SECCOMP_RET_LOG is implemented by Linux 4.14 but not present in all C libs

northstar/src/runtime/island/mod.rs
inorick: Since `check_platform_requirements()` fails at compile time now, we do not need any error here at all. I removed it.

northstar-tests/test-container/src/main.rs
inorick: :pray:

PR #409, opened 2021-07-30
Single threadded runtime

### LINKED ISSUE #385: Integration tests are flaky and fail sporadically on test_container log assertions in GHA
The integration tests fail from time to time in the GH CI. Couldn't reproduce it locally. Some of the tests are sporadically not getting any output of the `test_container` and fail random log assertions. This started somewhere around adding the parent termination watch thread in `runtime::island::init::init`.

### LINKED ISSUE #401: Sporadic deadlock upon container start: setresgid
When putting load on the system a sporadic deadlock happens within `setresgid` (libc) This is very likely a result from the clone from a MT program. In order to reproduce run the integration tests  in a loop and wait...

Screenshot 2021-07-22 at 12 23 41

### LINKED ISSUE #412: Capability dropping is guarded by capability list in manifest
If the `capabilities` set in the manifest is empty no capability is dropped. Adjust `island::init` to take `Capabilities` instead of `Option<capabilities>` and perform the drop unconditionally.

flxo:
Forking/cloning from a multi threaded application does not work. After
a clone/fork only the calling thread is alive and if any of the other
threads allocated any resource in the c lib or kernel the clone can end
up in a deadlock. There's basically nothing we can do about that beside
all the advices from the net to *not* do anything crucial between the
clone/fork and the execve. Since the major thing of the runtime is to
adjust the environment of the child process prior to the exeve call this
is nearly impossible. Stress tests with the `tools/stress` utility show
that after a couple of shorthands of operations something fails.

In the northstar runtime there's no real need for multi threading since
almost everything is tied to the borrow of `State`. Everything that is
triggered by a "event" is handled in that context and not parallel. The
only two things where multi threading give any benefit is

* parallel bulk mounts (long time spent in kernel for each mount)
* handling of container io (stdout, stderr)

The mount are handled in thread pool from the futures crate and are
guaranteed to block and wait for anything when a container is cloned.
This could be done even more safe by tearing down the pool after a mount
operation that requires the pool (or has a speed benefit from the pool).

With the single thread approach the stop of containers needed to be
refactored because signaling the process and waiting for its termination
doesn't work. Instead of having the `stop` as a pair of `kill` and
`wait` it's split into `kill` and `wait` from the user perspective and
is reflected in the model for the console.

Fixes #412
Fixes #401
Fixes #385


northstar-tests/src/logger.rs
```diff
@@ -14,75 +14,81 @@
...
+    },
+    time,
+};
 
 lazy_static! {
+    /// Test started timestamp
     static ref START: Instant = Instant::now();
-    static ref QUEUE: (flume::Sender<String>, flume::Receiver<String>) = flume::unbounded();
+    /// Channel for log assumptions
+    static ref QUEUE: (UnboundedSender<String>, Mutex<UnboundedReceiver<String>>) = {
```
marcmo, line 36: interesting pattern...did you use a mutex here just to get rid of the `flume` dependencies?
is this preferable to using a broadcast channel with subscriptions?

flxo: The `Mutex` is for getting a `&mut` to the receiver because of [recv](https://docs.rs/tokio/1.9.0/tokio/sync/mpsc/struct.Receiver.html#method.recv).

A `BroadcastReceiver` would require a `subscribe` in *every* `assume` call. The resulting stream would lack all logs *before* the `subscribe` so this would not work.

The `broadcast` is a thing "to be avoided" if possible because you have to handle the `Lagged` errors properly and that's not easy.

marcmo: ok thanks

northstar/src/runtime/state.rs
```diff
@@ -667,6 +667,11 @@ impl<'a> State<'a> {
             }
         }
 
+        // Spawn mount tasks onto the executor
+        let mounts = mounts
+            .drain(..)
+            .map(|t| self.executor.spawn_with_handle(t).unwrap());
```
marcmo, line 673: why do you have to do this manually? I always thought that the API to use for spawning parallel tasks is to use `Task::spawn`

northstar/src/util.rs
```diff
@@ -39,3 +39,17 @@ pub(crate) async fn is_rw(path: &Path) -> bool {
...
+    fn join_strip<T: AsRef<Path>>(&self, w: T) -> PathBuf;
+}
+
+impl PathExt for Path {
+    fn join_strip<T: AsRef<Path>>(&self, w: T) -> PathBuf {
+        if w.as_ref().starts_with("/") {
+            self.join(w.as_ref().strip_prefix("/").unwrap())
+        } else {
+            self.join(w)
+        }
```
marcmo, line 53: this would avoid the `unwrap`
```
        self.join(match w.as_ref().strip_prefix("/") {
            Some(stripped) => stripped,
            None => w,
        })
```

marcmo:
looks fine to me 👍

northstar/src/runtime/state.rs
flxo: The executor the runtime is running on is a `CurrentThreadExecutor` If we would spawn the mount futures onto that executor we wouldn't have any benefit because they're executed one by one. By spawning them onto a MT executor the blocking mount calls can be done in parallel and we get a speedup when mounting a larger amount of containers at once.

marcmo: ok...that is of course true! very good!

northstar/src/util.rs
flxo: Fixed.

PR #410, opened 2021-07-30
Fix new clippy warnings

inorick:
Rust 1.54.0 has more rigorous clippy warnings that are fixed in this PR.


inorick:
As these were just clippy warnings, I merged them right away.

PR #411, opened 2021-08-02
Fix sextant installation and add tools to CI

flxo:
Add the utils to the CI runs and fix `cargo install --path utils/sextant` by adding the `npk` feature to the northstar dependency.


mies2910 (approved):
Looks good.

PR #421, opened 2021-09-08
Add strace_parser to gen. seccomp manifest entries

inorick:
Add a new CLI tool to parse strace files from container runs and generate seccomp manifest entries from it.
This entry can then be added to the manifest of the container to enable seccomp support for it.


Cargo.toml
```diff
@@ -14,6 +14,7 @@ members = [
     "northstar-tests",
     "northstar-tests/test-container",
     "tools/nstar",
+    "tools/seccomp_util",
```
flxo, line 17: Please replace the `_` with a `-` to align with the other crates in here.

tools/seccomp_util/src/main.rs
```diff
@@ -0,0 +1,118 @@
+// Copyright (c) 2019 - 2020 ESRLabs
```
flxo, line 1: The year should be 2021

tools/seccomp_util/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+
+fn generate(strace_file: PathBuf) -> Result<()> {
+    let lines = open_file(strace_file);
+    let names = parse_strace(lines);
+    let entry = gen_seccomp_entry(names).expect("Failed to generate seccomp entry");
+    println!("{}", &entry);
+    Ok(())
+}
+
+fn open_file(strace_file: PathBuf) -> Vec<String> {
```
flxo, line 42: Can we inline those two and be consistent with either expecting or returning an `Result`?

tools/seccomp_util/src/main.rs
```diff
@@ -0,0 +1,118 @@
...
+            let name = caps.get(1).map_or("", |m| m.as_str());
+            let _args = caps.get(2).map_or("", |m| m.as_str());
+            names.push(name.to_string());
+        }
+    }
+    // Filter duplicates while retaining order
+    names.into_iter().unique().collect()
+}
+
+fn gen_seccomp_entry<I>(names: I) -> Result<String>
```
flxo, line 69: This could depend on `northstar` with feature `manifest` and use the `serde_yaml` instead of hand crafted yaml printing.  Probably useful when we do here something with the arguments (later).

northstar/src/npk/manifest.rs
```diff
@@ -319,6 +319,13 @@ pub struct Seccomp {
     pub allow: Option<HashMap<NonNullString, SyscallRule>>,
 }
 
+impl ToString for Seccomp {
+    fn to_string(&self) -> String {
+        // unwrap(): Can only if serialize fails which it does not for the types we use.
+        serde_yaml::to_string(self).unwrap()
+    }
+}
+
```
flxo, line 328: This feels odd.

It's a `to_yaml` and not a `to_string`.  It's also only used at a single place where even the potential error could be returned.

flxo (changes requested):

northstar/src/npk/manifest.rs
inorick: Agreed, I inlined the method and got rid of the unwrap().

inorick:
What part of the "npk" feature does it need?

flxo:
> What part of the "npk" feature does it need?

It's `npk::manifest::{Seccomp, SyscallRule}`. 

```
   Compiling seccomp-util v0.1.0 (/home/felix/northstar/tools/seccomp-util)
error[E0433]: failed to resolve: could not find `npk` in `northstar`
  --> tools/seccomp-util/src/main.rs:18:5
   |
18 |     npk::manifest::{Seccomp, SyscallRule},
   |     ^^^ could not find `npk` in `northstar`

error[E0433]: failed to resolve: use of undeclared type `SyscallRule`
  --> tools/seccomp-util/src/main.rs:47:75
   |
47 |                     syscalls.insert(NonNullString::try_from(m.as_str())?, SyscallRule::Any);
   |                                                                           ^^^^^^^^^^^ use of undeclared type `SyscallRule`

error[E0412]: cannot find type `SyscallRule` in this scope
  --> tools/seccomp-util/src/main.rs:39:46
   |
39 |     let mut syscalls: HashMap<NonNullString, SyscallRule> = HashMap::new();
   |                                              ^^^^^^^^^^^ not found in this scope

error[E0422]: cannot find struct, variant or union type `Seccomp` in this scope
  --> tools/seccomp-util/src/main.rs:52:17
   |
52 |     let entry = Seccomp {
   |                 ^^^^^^^ not found in this scope

error: aborting due to 4 previous errors

Some errors have detailed explanations: E0412, E0422, E0433.
For more information about an error, try `rustc --explain E0412`.
error: failed to compile `seccomp-util v0.1.0 (/home/felix/northstar/tools/seccomp-util)`, intermediate artifacts can be found at `/home/felix/northstar/target`

Caused by:
  could not compile `seccomp-util`

To learn more, run the command again with --verbose.
```

PR #422, opened 2021-09-14
Add secomp-util profile support

inorick:
Take the default seccomp profile into consideration when parsing strace files.
This behavior can be supressed with the `--no-default-profile` command line switch.

Move seccomp code to separate module.


flxo (approved):
LGTM

PR #424, opened 2021-09-16
CGroups V2 and container statistics

### LINKED ISSUE #18: Implement cgroups block io
blkio: Block IO controller - limit I/O usage

### LINKED ISSUE #309: Fix the cgroup memory test in GH actions
For some reason the integration test that waits for a cgroup memory enabled process to be under oom doesn't trigger when running in GH actions. This test is commeted in #306 and needs to be fixed and enabled.

### LINKED ISSUE #402: Refactor cgroups support
The northstar cgroups support is lazy. It's assuming everything is fine but it's not. The removal of the top level cgroup fails from time to time under load. Only a well defined subset of controllers is supported. Next the probably most important feature of cgroups is the statistics gathering regarding memory, cpu and io. The old approach to just query on the `init` pid lacks the resources used by the actual child application and has been removed.

The cgroup tests are currently disabled by #399. 
A draft implementation which makes use of `cgroups-rs` (a pre Rust cgroups wrapper) is in https://github.com/flxo/northstar/tree/pr-cgroups. This is blocked by some upstream changed needed.

flxo:
Gather container statistics via cgroups and expose via console. Add support for CGroups v2.

Fixes #402
Fixes #309
Fixes #18


flxo:
Fixes #402

northstar/src/api/model.rs
```diff
@@ -127,9 +142,16 @@ pub enum Response {
...
     Repositories(HashSet<RepositoryId>),
     Mount(Vec<MountResult>),
+    ContainerStats(ContainerStats),
     Err(Error),
 }
 
+#[derive(new, Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
+pub struct ContainerStats {
+    pub container: Container,
+    pub stats: HashMap<String, serde_json::Value>,
```
alfonsoros88, line 152: Maybe use `crate::runtime::stats::ContainerStats` or some alias for the type instead?

northstar/src/api/model.rs
```diff
@@ -127,9 +142,16 @@ pub enum Response {
     Containers(Vec<ContainerData>),
     Repositories(HashSet<RepositoryId>),
     Mount(Vec<MountResult>),
+    ContainerStats(ContainerStats),
     Err(Error),
 }
 
+#[derive(new, Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
+pub struct ContainerStats {
+    pub container: Container,
```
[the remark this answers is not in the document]
alfonsoros88: It seems that this struct is used in the `Response` to the client request, maybe it would be better to respond with a `(Container, ContainerStats)`, similar to the `(Container, ContainerEvent)` notification.

northstar/src/runtime/cgroups.rs
```diff
@@ -32,244 +37,331 @@ use tokio::{
...
+
+    fn v2(&self) -> bool {
+        self.inner.v2()
     }
 }
 
 #[derive(Debug)]
 pub struct CGroups {
-    groups: Vec<CGroup>,
+    container: Container,
```
alfonsoros88, line 127: maybe unnecessary

northstar/src/runtime/cgroups.rs
```diff
@@ -32,244 +37,331 @@ use tokio::{
...
+            let stop = token.clone();
+            task::spawn(async move {
+                debug!("Listening for oom events of {}", container);
+                let mut buffer = [0u8; 16];
+
+                'outer: loop {
+                    select! {
+                        _ = stop.cancelled() => break 'outer,
+                        _ = tx.closed() => break 'outer,
+                        _ = event_fd.read(&mut buffer) => {
```
alfonsoros88, line 268: is something meaningful written to `buffer`? is it just fine to ignore it?

northstar/src/runtime/cgroups.rs
```diff
@@ -32,244 +37,331 @@ use tokio::{
...
+
+                debug!("Stopped oom monitor of {}", container);
+            })
+        };
+
+        MemoryMonitor { token, task }
+    }
+
+    /// Construct a new cgroups v2 memory monitor
+    async fn new_v2(container: Container, path: &Path, tx: EventTx) -> MemoryMonitor {
```
alfonsoros88, line 293: It feels like some duplication in these functions but they are different enough to not make it obvious >:(

alfonsoros88 (commented):
Looks good! I was expecting more deletions but still it is better than managing cgroups directly plus the v2 support :)

northstar/src/api/model.rs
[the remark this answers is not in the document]
flxo: Right! Updated the 

```rust
#[derive(new, Clone, Eq, PartialEq, Debug, Serialize, Deserialize)]
pub enum Response {
    ...
    ContainerStats(Container, ContainerStats),
    ...
}
```

and replaced `ContainerStats` with 

```rust
pub type ContainerStats = HashMap<String, serde_json::Value>;
```

northstar/src/runtime/cgroups.rs
flxo: Since we're only watching for Modify events the content of buffer is always a modify event with some additional not relevant data. Thinks is safe to ignore.

northstar/src/runtime/cgroups.rs
flxo: Yes. Merging those two would remove trivial code add add fuzz ;-)

northstar/src/api/model.rs
flxo: Done.

northstar/src/api/codec.rs
```diff
@@ -187,8 +189,15 @@ mod tests {
...
-                "test:0.0.1".try_into().unwrap()
+            Just(Message::Notification(Notification::CGroup(
+                "test:0.0.1".try_into().unwrap(),
+                CgroupNotification::Memory(MemoryNotification {
+                    low: None,
+                    high: Some(10),
+                    max: Some(12),
+                    oom: None,
+                    oom_kill: Some(99),
+                })
```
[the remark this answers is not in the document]
flxo: Agree but since we're in fact just testing `serde-json` here, I'd vote for omitting the additional (beta) dependency to `proptest-derive`.

PR #427, opened 2021-09-20
Replace dev tmpdir with squashfs features

### LINKED ISSUE #404: Replace dev tmpdir with squashfs pseudo files
The new release of the squashfs tools 4.5 supports character and block device pseudo files. This allows to create a minimal dev version when packing the npk instead of creating the device files at runtime. This improves startup performance and removes the need of a tmpdir.

[Release notes](https://sourceforge.net/p/squashfs/mailman/message/37324506/)

```
9. New Pseudo file definitions with timestamps
==============================================

The Pseudo file definitions previously supported did not allow the time
of the file to be set.  New extended pseudo file definitions have been added
which take a <time> timestamp.  These are distinquished from the previous
definitions by using an upper-case type character.  For example the "D"
definition is identical to the previous "d" definition (it creates
a directory), but it takes a <time> timestamp.

The list of extended definitions are:

filename D time mode uid gid
filename M time mode uid gid
filename B time mode uid gid major minor
filename C time mode uid gid major minor
filename F time mode uid gid command
filename S time mode uid gid symlink

<time> can be either an unsigned decimal integer (which represents the
seconds since the epoch of 1970-01-01 01:00 UTC), or a "date string"
which is parsed and converted into an integer since the epoch, by calling
"date".

Because most date strings have spaces, they will need to be quoted, and if
entered on the command line, these quotes will need to be protected from the
shell by backslashes, i.e.

% mksquashfs x x.sqsh -p "file D \"1 jan 1980\" 0777 phillip phillip"

Obviously anything "date" accepts as a valid string can be used, such as
"yesterday", "last week" etc.
```

flxo:
The content of the minimal /dev can be created during npk creation. This has the advantage that no files/symlinks need to be created during runtime and the tmpdir assiciated with each container is ripped away.

Fixes #404

### DIFF
--- a/northstar/src/npk/npk.rs
+++ b/northstar/src/npk/npk.rs
@@ -100,6 +100,15 @@ pub enum Error {
     Version(Version, Version),
 }
 
+impl Error {
+    fn io<T: ToString>(context: T, error: io::Error) -> Error {
+        Error::Io {
+            context: context.to_string(),
+            error,
+        }
+    }
+}
+
 /// NPK archive comment
 #[derive(Debug, Serialize, Deserialize)]
 pub struct Meta {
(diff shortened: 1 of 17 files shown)


northstar/src/npk/npk.rs
```diff
@@ -635,58 +644,115 @@ fn sign_npk(key: &Path, fsimg: &Path, manifest: &Manifest) -> Result<String, Err
...
+                p.push(dir);
+                writeln!(file, "{} d {} {} {}", p.display(), mode, uid, gid).map_err(|e| {
+                    Error::Io {
+                        context: format!("Failed to write entry {} to temp file", p.display()),
+                        error: e,
+                    }
+                })?;
+                Ok(p)
+            })
+            .map(drop)
```
alfonsoros88, line 677: nit: This is a pretty cool iterator expression, but maybe in this particular case a good ol' for loop could be a bit more readable?

```rust
let mut p = PathBuf::from("/");

for d in dir.iter().skip(1) {
    p.push(d);
    writeln!(file, "{} d {} {} {}", p.display(), mode, uid, gid).map_err(|e| {
        Error::Io {
            context: format!("Failed to write entry {} to temp file", p.display()),
            error: e,
        }
    })?;
}
```

alfonsoros88: Also it looks like this function is big enough to be moved to the outer scope and ideally have it documented and unit tested?

northstar/src/npk/npk.rs
```diff
@@ -635,58 +644,115 @@ fn sign_npk(key: &Path, fsimg: &Path, manifest: &Manifest) -> Result<String, Err
...
+                        file,
+                        "{} c {} {} {} {} {}",
+                        target.join(dev).display(),
+                        666,
+                        uid,
+                        gid,
+                        major,
+                        minor
+                    )
+                    .map_err(|e| Error::io("Failed to write pseudofile", e))?;
```
alfonsoros88, line 731: nit: maybe move this bolck to a `add_chardevs` together with `add_directory` and have them documented/tested

alfonsoros88 (approved):
Looks good!

northstar/src/npk/npk.rs
flxo: Thanks for the hint!. I refactored the whole function. The writes on line were just too noisy.

alfonsoros88 (approved):
Much better!

PR #428, opened 2021-09-20
Remove redundant data in MountedContainer

alfonsoros88:
While working on the youki feature, noticed that `MountedContainer` had
some redundant information. This patch is a refactoring that tries to
remove the unnecessary `manifest` and `container` fields out of `MountedContainer`.

### DIFF
--- a/northstar/src/runtime/process/fs.rs
+++ b/northstar/src/runtime/process/fs.rs
@@ -14,11 +14,12 @@
 
 use super::Error;
 use crate::{
+    common::container::Container,
     npk::{
         manifest,
-        manifest::{MountOption, MountOptions, Resource, Tmpfs},
+        manifest::{Manifest, MountOption, MountOptions, Resource, Tmpfs},
     },
-    runtime::{config::Config, state::MountedContainer},
+    runtime::config::Config,
     util::PathExt,
 };
 use log::debug;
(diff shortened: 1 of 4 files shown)


flxo:
@alfonsoros88 I rebased because I merged #424

northstar/src/runtime/process/fs.rs
```diff
@@ -92,35 +93,34 @@ impl Mount {
...
 
     for (target, mount) in manifest_mounts {
         match &mount {
             manifest::Mount::Bind(manifest::Bind { host, options }) => {
-                mounts.extend(bind(&root, target, host, options));
+                mounts.extend(bind(root, target, host, options));
             }
             manifest::Mount::Persist => {
-                mounts.push(persist(&root, target, config, container).await?);
+                let source = config.data_dir.join(manifest.name.to_string());
```
flxo, line 112: We shouldn't forget to document that the version is intentionally not part of the path here. This allows upgrades with persistent data migration.

northstar/src/runtime/process/fs.rs
```diff
@@ -188,47 +188,49 @@ fn bind(root: &Path, target: &Path, host: &Path, options: &MountOptions) -> Vec<
 
 async fn persist(
     root: &Path,
+    source: &Path,
     target: &Path,
-    config: &Config,
-    container: &MountedContainer,
+    uid: u16,
+    gid: u16,
```
flxo, line 194: Like that!

northstar/src/runtime/state.rs
```diff
@@ -841,4 +850,12 @@ impl<'a> State<'a> {
                 .expect("Internal channel error on main");
         }
     }
+
+    /// Returns an iterator over the containers currently running
+    fn running_containers(&self) -> impl Iterator<Item = &Container> {
+        self.containers
+            .iter()
+            .filter(|(_, state)| state.process.is_some())
```
flxo, line 857: nit: There's a handy [filter_map](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map) and I think this should work:
```rust
.filter_map(|(container, state)| state.proces.map(|_| container))
```

flxo (approved):
LGTM!

alfonsoros88:
> @alfonsoros88 I rebased because I merged #424

whoops... didn't see this on time :see_no_evil:

flxo:
> > @alfonsoros88 I rebased because I merged #424
> 
> whoops... didn't see this on time 🙈

No worries - feel free to discard it...:-)

northstar/src/runtime/state.rs
alfonsoros88: I wasn't completely happy using either `state.proces.map(|_| container)` or ``state.proces.and(Some(container))`, it reads kind of weird. Fortunately `rust-analyzer` saved the day :tada: and suggested `state.process.is_some().then(|| container)` that is clearer and nicer to read. Didn't know those `.then` and `.then_some` methods on `bool` :)

northstar/src/runtime/process/fs.rs
alfonsoros88: Not sure I understand fully this but added your comment in a note above this line :blush:

PR #431, opened 2021-09-23
containerdb

alfonsoros88:
The intention is to simplify a bit the source of state.rs by moving to a new module all the logic for managing containers and repositories.

### DIFF
--- a/northstar/src/runtime/containerdb.rs
+++ b/northstar/src/runtime/containerdb.rs
@@ -1,0 +1,208 @@
+//! Defines [ContainerDB], a struct to manage containers
+//!
+//! Northstar organizes containers in repositories. Repositories can be any object that implements
+//! the [Repository][crate::runtime::repository::Repository] trait.
+
+use std::{collections::HashMap, ops::Deref, sync::Arc};
+
+use bytes::Bytes;
+use tokio::sync::mpsc::Receiver;
+
+use crate::{
+    common::container::Container,
+    npk::manifest::Manifest,
+    runtime::{
+        error::Error,
+        repository::{Repository, RepositoryId},
+        state::Npk,
+    },
+};
+
+/// Container Database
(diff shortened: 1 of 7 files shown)


northstar/src/runtime/containerdb.rs
```diff
@@ -0,0 +1,211 @@
...
+        }
+
+        self.find_npk(container).map(ManifestRef)
+    }
+
+    /// Find the repository that has the given container
+    pub fn find_repository(&self, container: &Container) -> Option<&dyn Repository> {
+        self.index_by_container
+            .get(container)
+            .and_then(|&idx| self.repositories.get(idx).map(|(_, b)| b.as_ref()))
```
flxo, line 84: Index operations scare me...

flxo (commented):
The relation between the lines added and removed or simplified by this patch is a little bit odd. 
The improvements in the `state` module where NPKs or repositories are searched are kind of small and the complexity with the patch is almost the same as before.

northstar/src/runtime/containerdb.rs
alfonsoros88: i'd agree, on the other hand this is not unsafe, should not panic if there is a bug,

alfonsoros88:
> The relation between the lines added and removed or simplified by this patch is a little bit odd. The improvements in the `state` module where NPKs or repositories are searched are kind of small and the complexity with the patch is almost the same as before.

Well yes, obviously there are more insertions due to the new module with its unit tests and so on. There is not much simplification in terms of changing things mainly moving and separating stuff. The intention here is to move the logic that wasn't inherently a responsibility of state. I see this as the first step and hopefully in subsequent patches we can better organize this part of the code. I think that the new module makes sense and its kind of clear what are its responsibilities.

alfonsoros88:
closing this one in favor o #438

PR #436, opened 2021-10-05
Double fork when spaning a process

flxo:
Instead of dedicated clone implementations use `fork` and unshare to
create the init process of containers. Communicate the pid of init via a
pipe.


northstar/src/runtime/process/init.rs
```diff
@@ -98,13 +89,10 @@ impl Init {
...
                 }
                 unistd::ForkResult::Child => {
-                    // If init dies, we want to die as well. This should normally never happen and is a error condition.
-                    set_parent_death_signal(SIGKILL);
-
                     // Unblock signals. The signal mask has been set in the runtime prior to the clone call. This
                     // avoids that init receives signals which are unhandled. Reminder: init doesn't have any
                     // signals handlers because it's init of a PID ns.
-                    super::signals_unblock();
+                    //super::signals_unblock();
```
inorick, line 95: Do we not need to unlock signals anymore? If so, we can delete this line.

PR #440, opened 2021-10-13
Version as optional argument for nstar commands

flxo:
In most cases the version of a container is unique during development. Save some seconds and make nstar query the container list and try to pick a version if it's unique.


marcmo:
good idea...looks good!

PR #445, opened 2021-10-15
Support all possible exit stati of childs

### LINKED ISSUE #444: Support full u8 range of container exit codes
If a container exits signalled, the runtime does a weird hack to "encode" the signal into the exit code of init. This is broken for exit codes > 127 aka `SIGNAL_OFFSET`.

Either increase the offset *or* adjust the process module and communicate the PID of the child of init to the runtime. The runtime `waitpids` on this pid instead of `init`. With that approach `init` could even exit. The child gets reparented and no dirty hack is needed for the exit status because the runtime can directly use the waitpid result of the child.

flxo:
The encoding of signals into the exit status has limitations. Therefore reuse the pipe for signaling the pid of init to the runtime to send the exit status of the child from init to the runtime.

Fixes #444


northstar/src/runtime/mod.rs
```diff
@@ -117,12 +119,12 @@ struct MemoryEvent {
...
 
 /// Container exit status
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub enum ExitStatus {
     /// Process exited with exit code
     Exit(ExitCode),
     /// Process was terminated by a signal
-    Signaled(signal::Signal),
+    Signalled(u8),
```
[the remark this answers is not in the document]
flxo: Signal are in the range 0..=64. The `repo(i32)` in `nix` is probably done because of the C lib API.

northstar/src/runtime/process/mod.rs
```diff
@@ -278,27 +299,45 @@ fn waitpid(
...
+        let exit_status = match channel.recv::<ExitStatus>().await {
+            Ok(exit_status) => {
+                debug!(
+                    "Received exit status of {} via channel: {}",
+                    container, exit_status
+                );
+
+                // Wait for init to exit
+                let pid = unistd::Pid::from_raw(pid as i32);
+                wait::waitpid(Some(pid), None).ok();
```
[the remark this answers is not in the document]
flxo: The `waitpid` is  needed because you can have the situation that the `init` process exits and within the exit it closes the pipe. The runtime starts to process this exit by tearing down cgroups but the process is not yet gone. Therefore wait until the pipe is closed *and* the process is completely exited.

northstar/src/runtime/process/mod.rs
```diff
@@ -165,7 +188,9 @@ impl Launcher {
                         .expect("Failed to wait for trampoline process");
 
                     // Create a future that resolves upon exit of init
-                    let exit_status = waitpid(container, pid, self.tx.clone(), stop);
+                    let container = container.clone();
+                    let tx = self.tx.clone();
+                    let exit_status = waitpid(container, pid, tx, stop, exit_status_channel);
```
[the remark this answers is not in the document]
flxo: The intention is that this future resolves to the `exit_status` of `pid`. The information that it's a task (aka `JoinHandle<>`) is not really needed.

northstar/src/runtime/process/mod.rs
```diff
@@ -278,27 +299,45 @@ fn waitpid(
     pid: Pid,
     tx: EventTx,
     stop: CancellationToken,
+    mut channel: pipe::AsyncChannelRead,
```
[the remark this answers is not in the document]
flxo: > after introducing this channel and removing the wait for `sigchld`, I'm not sure if the name `waitpid` is still valid for this function.

I kept the name `waitpid` intentionally because every system programmer knows what a `waitpid` does.

> I would actually suggest to inline this function. It would also nice to have some more verbosity on the variable names handled here. Using `pid`, `channel` and `exit_status` is confusing after so many `forks`. Maybe we could have something like:
> 
> ```rust
>    let process_exit_status = child_channel.rec::<ExitStatus>();
>    let init_exit_status = waitpid(init_pid);
> 
>     // here goes some explanation to why this is the way it is
>     let exit_status_fut = process_exit_status.or_else(|error| {
>         warn!("failed to ... {}" error);
>         init_exit_status
>     });
> 
>    let notify_on_exit_task = task::spwan(async move {
>        let exit_status = exit_status_fut.await.expext("no exit status");
>        stop.cancel();
>        tx.send(exit_status).await.ok();
>       // ...
>    });
> ```

Ok. I'll get back on this and try to sort things a little bit. Probably I spent too much time in this code and some things are obvious for me which aren't...

northstar/src/runtime/process/trampoline.rs
```diff
@@ -0,0 +1,26 @@
+use super::{init::Init, Checkpoint};
+use crate::runtime::pipe::Channel;
+use nix::{sched, unistd};
+use std::process::exit;
+
+pub(super) fn trampoline(init: Init, mut child_channel: Channel, checkpoint_init: Checkpoint) -> ! {
```
alfonsoros88, line 6: I have the impression that this pattern is repeated in a couple of places. Would it be possible to have a generalization of this like:

```rust
fn spawn_child(task: impl FnOnce() -> Result<(), Error>) -> impl Future<Output = Result<ExitStatus, Error>> {
    match unsafe { fork() }.expect("...") {
        Parent { child } => {
            waitpid(child)
        }
        Child => {
            match task() {
                Ok(()) => exit(0),
                Err(e) => // log or something and exit with -1 or something
            }
        }
    }
}
```

flxo: I think the three forks are very different and probably easier to read with a match for each since we read this `if fork() == -1 {.....` since years :-)

northstar/src/runtime/process/mod.rs
```diff
@@ -272,87 +320,50 @@ impl super::state::Process for Process {
...
-        );
-        exit_status
-    })
-    .map(|r| r.expect("Task join error"))
-}
-
-/// Get exit status of process with `pid` or None
-fn exit_status(pid: Pid) -> Option<ExitStatus> {
+/// Get exit status of process with `pid` if the process already exited. Panic otherwise.
+fn init_exit_status(pid: Pid) -> ExitStatus {
```
alfonsoros88, line 324: the `init_` part is making the function name confusing. The function is calling a blocking `waitpid` on a `pid` and panicking if something else to `exit` or `signaled` is returned. Maybe `reap_child` is a bit better name?

Alternatively, maybe it would be nicer to have and use a `async fn waitpid` similar the remove one:
```rust
async fn waitpid(pid: Pid) -> ExitStatus {
    let mut sigchld = signal::unix::signal(signal::unix::SignalKind::child()).expect("...");
    loop {
        sigchld.recv().await;
        if let Some(status) = exit_status(pid) {
            return status
        }
    }
}
```

northstar/src/runtime/process/trampoline.rs
```diff
@@ -0,0 +1,26 @@
...
+    // Create pid namespace
+    sched::unshare(sched::CloneFlags::CLONE_NEWPID).expect("Failed to create pid namespace");
+
+    // Fork the init process
+    match unsafe { unistd::fork() }.expect("Failed to fork init") {
+        unistd::ForkResult::Parent { child } => {
+            // Send the pid of init to the runtime and exit
+            let pid = child.as_raw() as i32;
+            child_channel.send(&pid).expect("Failed to send init pid");
+            exit(0);
```
alfonsoros88, line 16: I think that this causes the `init` process to be an orphan. The `trampolin` process should wait on its exit status.

northstar/src/runtime/process/mod.rs
```diff
@@ -143,32 +162,83 @@ impl Launcher {
...
+                                Ok(exit_status) => {
+                                    debug!(
+                                        "Received exit status of {} ({}) via channel: {}",
+                                        container, pid, exit_status
+                                    );
+
+                                    // Wait for init to exit. This is needed to ensure the init process
+                                    // exited before the runtime starts to cleanup e.g remove cgroups
+                                    let pid = unistd::Pid::from_raw(pid as i32);
+                                    wait::waitpid(Some(pid), None).ok();
```
alfonsoros88, line 208: Not sure this is even doing anything. The init process ends up being an orphan because the trampoline dies without waiting for it. This call should fail with a `ECHILD` but it is ignored either way.

alfonsoros88 (commented):
I think there is an issue with the trampolin process and the pid hierarchy

northstar/src/runtime/process/mod.rs
```diff
@@ -272,87 +320,50 @@ impl super::state::Process for Process {
...
-    .map(|r| r.expect("Task join error"))
-}
-
-/// Get exit status of process with `pid` or None
-fn exit_status(pid: Pid) -> Option<ExitStatus> {
+/// Get exit status of process with `pid` if the process already exited. Panic otherwise.
+fn init_exit_status(pid: Pid) -> ExitStatus {
     let pid = unistd::Pid::from_raw(pid as i32);
-    match wait::waitpid(Some(pid), Some(WaitPidFlag::WNOHANG)) {
+    let result = wait::waitpid(Some(pid), None);
```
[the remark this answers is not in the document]
flxo: With the `WNOHANG` we can have the situation that the init process closed the pipe but it not yet exited.
Yes - this blocks as we block in some other places as well. The assumption is that in `init` the pipe is never closed and the only possibility that the read makes progress is that init sends a exit stati of the child or exits itself. In both cases the `waitpid` is fine.

northstar/src/runtime/process/mod.rs
flxo: This is basically what we have had before. The drawback is that with the SIGCHLD we don't get the information about the pid that raised the SIGCHLD and we have to poll all running containers. As a tradeoff we call the (blocking) waitpid after we're sure that init exited because we either read the exit status of the child on the channel *or* the channel is closed.

northstar/src/runtime/process/trampoline.rs
flxo: Check `Launcher::start` which does an `prctl(PR_SET_CHILD_SUBREAPER)` to reap all childs that become orphans in this process tree. `init` is a child of the runtime after `trampoline` exited.

northstar/src/runtime/process/mod.rs
flxo: >  This call should fail with a `ECHILD` but it is ignored either way.

Good catch - have to check that. Maybe the call is too fast and init hasn't been adopted yet.

northstar/src/runtime/process/trampoline.rs
alfonsoros88: oh that's cool :)

northstar/src/runtime/process/mod.rs
flxo: I changed to the `ok` to a assertion on `.is_ok()`. Looks like `init` is reliably adopted by the runtime.

northstar/src/runtime/process/mod.rs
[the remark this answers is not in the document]
alfonsoros88: hmm okay but shouldn't we at least use some `spawn_blocking` or `block_in_place`?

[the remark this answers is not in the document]
flxo: We need to run on a single threaded runtime because we `clone`. There's no `spawn_blocking` or `block_in_place`.
The code in in init is supposed to work. If it's not - and init crashes - the `waitpid` also returns. The only chance that we hang forever is when the `waitpid` in `init` hangs forever - but then we're screwed anyways....

northstar/src/runtime/process/mod.rs
alfonsoros88: yeah, after thinking about it, it feels like the asynchronous `waitpid` is likely to suffer from race conditions with multiple child processes. But back to the original point, the name is a bit odd isn't it? how about a `waitpid` with some `block_in_place` inside?

flxo: There's no possibility to `block_in_place` on a single threaded executor. See comment above.

What's your suggestion for the fns name? `reap_child` is not enough since the result of `waitpid` is used if there's not exit status received via the pipe. (sigkill case).

northstar/src/runtime/process/mod.rs
alfonsoros88: well it looks like it'd be hard to be sure that such race could not ever happen in the current state.. but if you think it is fine this way then i ok too i guess ¯\_(ツ)_/¯

northstar/src/runtime/process/mod.rs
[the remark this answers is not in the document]
alfonsoros88: okay, looks like there is no way around then :sweat_smile:

northstar/src/runtime/process/mod.rs
alfonsoros88: `wait_exit_status(_blocking)`?

flxo: That's why I originally named it `waitpid` ;-)

`wait_exit_status` is fine for me.

alfonsoros88: well to be fair, `waitpid` was inlined and this is not the same :stuck_out_tongue_winking_eye:

northstar/src/runtime/process/mod.rs
alfonsoros88: well if it happens in the future we just put a `match` expression here and handle the `ECHLD` :)

flxo: > well if it happens in the future we just put a `match` expression here and handle the `ECHLD` :)

IMHO a `ECHLD` would be a bug. Let me replace the assertion with  a match to see what `waitpid actually returned before we panic. The assertion on `is_ok` would not show that.

PR #453, opened 2021-11-05
Add JsonSchema generation for console API

flxo:
Integrate (schemers)[https://docs.rs/schemars/0.8.6/schemars/] to generate JsonSchemas for the console API models. Add a small binary tool to dump the json schema to a file or stdout.

The probably most major change in this patch series is that no external types must be used in the API (server::Version, caps::Capability etc).


northstar/src/common/version.rs
```diff
@@ -1,2 +1,148 @@
...
+#[derive(Error, Debug, Display)]
+pub struct ParseError {
+    #[from]
+    source: semver::Error,
+}
+
 /// Container version
-pub type Version = semver::Version;
+#[derive(Clone, Eq, PartialEq, Hash, Debug)]
+pub struct Version {
```
alfonsoros88, line 20: just curious, why is another version type needed?

northstar/src/runtime/cgroups.rs
```diff
@@ -129,13 +132,13 @@ impl CGroups {
         );
 
         let resources = cgroups_rs::Resources {
-            memory: config.memory.clone().unwrap_or_default(),
+            memory: config.memory.clone().map(Into::into).unwrap_or_default(),
```
alfonsoros88, line 135: I see a lot of `.map(Into::into)` in this file. We have already a dependency to `itertools` that has a `.map_into()` not sure if that would improve anything tho :man_shrugging:

northstar/src/runtime/process/fs.rs
```diff
@@ -248,7 +248,7 @@ fn tmpfs(root: &Path, target: &Path, size: u64) -> Mount {
 
 fn options_to_flags(opt: &MountOptions) -> MsFlags {
     let mut flags = MsFlags::empty();
-    for opt in opt {
+    for opt in &**opt {
```
alfonsoros88, line 251: is this fine? :thinking:

alfonsoros88 (approved):
lgtm!

northstar/src/runtime/cgroups.rs
flxo: In that case its an ordinary option and not an iterator. Think converting the option into an iterator for getting `Itertools::map_into` doesn't sound like a obvious improvement.

alfonsoros88: yeah, wasn't suggesting that :sweat_smile: just thought all of these were iterators, never mind then :)

Issue #454, opened 2021-11-09
`nstar exec` subcommand


rosds:
This feature enables the execution of additional processes inside a running containers, just like `docker  exec`.

The considered approach will required to extend the interface with the init process running inside each container so that it can be commanded to start new processes. It would also have to keep a record of all the child processes and their respective status.

Each new spawned process will use a separated pseudo-terminal with its master file descriptor sent to the `nstar` client running the `exec` sub-command. The slave part is set to handle the `stdin`, `stdout` and `stderr` of the spawned process.

arosds:
This feature addresses a bit of #212

PR #456, opened 2021-11-16
init process api

alfonsoros88:
This is a first step towards #454.

The goal of this patch is to extend the functionality of `init`
processes inside each container. The extension consist of 2 points:

* Await and communicate the exit status of *every* child process.
  Currently, `init` only awaits for the single process that is specified
  in the container manifest.

* Spawn new processes at command. This will allow in the future to
  execute new processes inside the container using `exec`.


northstar/src/runtime/process/init.rs
```diff
@@ -76,34 +100,21 @@ impl Init {
...
-                            }
-                            Err(nix::Error::EINTR) => continue,
-                            e => panic!("Failed to waitpid on {}: {:?}", child, e),
-                        }
-                    }
+                    // FIXME: this is a workaround to be able to have a multi-threaded runtime
+                    // running inside the init process. It should be possible to remove this
+                    // once we implement have a separate process that does the forking.
+                    std::thread::spawn(move || {
+                        let rt = tokio::runtime::Builder::new_multi_thread().build().unwrap();
```
flxo, line 111: Is the MT rt needed to handle blocking stuff e.g `waitpid` in future? Currently a `current_thread` runtime would be cheaper and sufficient.

If we need a MT RT, we should probably fine tune the number of [worker_threads](https://docs.rs/tokio/1.14.0/tokio/runtime/struct.Builder.html#method.worker_threads).

northstar/src/runtime/process/init.rs
```diff
@@ -30,12 +30,36 @@ pub(super) struct Init {
     pub seccomp: Option<AllowList>,
 }
 
+/// Init process main function
+///
+/// This functions runs inside a multi-threaded tokio runtime.
+///
+async fn init_main(mut channel: Channel) {
+    loop {
+        match waitpid(None, None) {
```
flxo, line 39: If the `exec` feature comes true the runtime should not consider the container as exited when the `exec` process exits. Probably extending the messages that is sent via the channel make sense e.g

```rust
enum Exit {
  // The command specified in the manifest as init. Oh man -  this is confusing
  Init(ExitStatus),
  Other(Pid, ExitStatus),
}

File: northstar/src/runtime/process/init.rs (hunk 1 above)
Comment (alfonsoros88) [reply northstar/src/runtime/process/init.rs:111]: yeah, this is true, this was just an experiment :sweat_smile: i'll use a single thread and see if we can get away with that

File: northstar/src/runtime/process/init.rs (hunk 2 above)
Comment (alfonsoros88) [reply northstar/src/runtime/process/init.rs:39]: yep, this adds a new dimension to consider in the start/stop/mount/unmout/shutdown cases :cry:

File: northstar/src/runtime/process/fork.rs [hunk 3]
```diff
@@ -0,0 +1,24 @@
...
+where
+    F: Future<Output = Result<(), Error>>,
+{
+    match unsafe { unistd::fork()? } {
+        unistd::ForkResult::Parent { child } => Ok(child),
+        unistd::ForkResult::Child => {
+            if let Err(e) = f.await {
+                log::error!("Failed after fork: {:?}", e);
+                std::process::exit(-1);
+            }
```
marcmo, line 20: I thought using the tokio runtime here without spawning a new runtime in a thread is problematic

northstar/src/runtime/process/fork.rs
```diff
@@ -0,0 +1,25 @@
...
+
+use crate::runtime::error::Error;
+
+/// Fork a new process.
+///
+/// # Arguments:
+///
+/// * `f` - The closure to run in the child process.
+///
+pub fn fork_process<F>(f: F) -> nix::Result<Pid>
```
flxo, line 11: why not just `fork`? It will always be a process...

northstar/src/runtime/process/init.rs
```diff
@@ -30,11 +30,130 @@ pub(super) struct Init {
...
+    let command_tx = tx.clone();
+    let _command_thread = std::thread::spawn(move || loop {
+        let command = rt_rx
+            .recv()
+            .expect("Failed to receive init command from runtime");
+        command_tx.send(Payload::Command(command)).unwrap();
+    });
+
+    // This thread sends the exit status of the child processes
+    let exit_status_tx = tx;
```
flxo, line 73: This just a rename.

northstar/src/runtime/process/init.rs
```diff
@@ -30,11 +30,130 @@ pub(super) struct Init {
...
+                .expect("Failed to spawn child process");
+
+                // wake up the reaper thread
+                reaper_thread.thread().unpark();
+
+                rt_tx
+                    .send(InitMessage::NewChild {
+                        pid: child_pid.as_raw(),
+                    })
+                    .expect("Failed to send exit status");
```
flxo, line 143: Wakeup the reaper *after* sending the result to the child. If the child immediatel exits there's a chance that the runtime gets a `DeadChild` exit before a `NewChild`

northstar/src/runtime/process/init.rs
```diff
@@ -30,11 +30,130 @@ pub(super) struct Init {
     pub seccomp: Option<AllowList>,
 }
 
+/// This is the type used to communicate with the init process from the runtime.
+#[derive(Debug, Serialize, Deserialize)]
+pub enum InitMessage {
+    NewChild { pid: i32 },
+    DeadChild { pid: i32, exit_status: ExitStatus },
```
flxo, line 37: Can we rename this please? I have bad memories.

```rust
pub enum Message {
   Forked? { ... },
   Exit { ...},
}

File: northstar/src/runtime/process/init.rs [hunk 8]
```diff
@@ -30,11 +30,130 @@ pub(super) struct Init {
...
+    },
+}
+
+/// Init process main function
+fn init_main(
+    mut rt_tx: channel::Sender<InitMessage>,
+    mut rt_rx: channel::Receiver<InitApi>,
+    seccomp: Option<AllowList>,
+) -> ! {
+    enum Payload {
```
flxo, line 57: In the RT we called this `Event`. Would be a good fit here as well.

northstar/src/runtime/process/init.rs
```diff
@@ -30,11 +30,130 @@ pub(super) struct Init {
...
+        ChildExit { pid: i32, exit_status: ExitStatus },
+        Command(InitApi),
+    }
+    let (tx, rx) = mpsc::channel();
+
+    // This thread forwards the commands coming from the runtime
+    let command_tx = tx.clone();
+    let _command_thread = std::thread::spawn(move || loop {
```
flxo, line 65: The `JoinHandle` is not used. The thread will stop when `rx` is closed which can be gracefully done in the main loop. We could join the thread then.

Other wise detach the thread by dropping the handle.

northstar/src/runtime/process/init.rs
```diff
@@ -30,11 +30,130 @@ pub(super) struct Init {
...
+                        })
+                        .unwrap();
+                }
+                Ok(WaitStatus::Signaled(pid, status, _)) => {
+                    exit_status_tx
+                        .send(Payload::ChildExit {
+                            pid: pid.as_raw() as i32,
+                            exit_status: ExitStatus::Signalled(status as u8),
+                        })
+                        .unwrap();
```
flxo, line 91: We have to make sure that the whole process panics upon this unwrap.
When this thread panics, `tx` is dropped.  The other forward thread also owns a copy of tx and we will not notice that the reaper is gone.

northstar/src/runtime/process/mod.rs
```diff
@@ -53,17 +52,29 @@ pub(super) struct Launcher {
...
+}
+
 pub(super) struct Process {
-    pid: Pid,
-    checkpoint: Option<Checkpoint>,
+    init_tx: channel::AsyncSender<InitApi>,
+    init_rx: Option<channel::AsyncReceiver<InitMessage>>,
+    console_stop: Option<CancellationToken>,
+    container: Container,
+    container_app: ContainerApplication,
```
flxo, line 67: I'm not happy with `container_app`. A process in a os has also attributes like `cmd`, `args, `env`. Maybe just flatten it here?

northstar/src/runtime/process/mod.rs
```diff
@@ -322,8 +323,31 @@ impl super::state::Process for Process {
                 let process_group = unistd::Pid::from_raw(-(self.pid as i32));
                 let sigkill = Some(sys::signal::SIGKILL);
                 sys::signal::kill(process_group, sigkill).ok();
+                return Ok(());
             }
-        }
+        };
+
+        // spawn a task that waits for the process exit status
+        let tx = self.tx.clone();
```
flxo, line 331: What about moving `tx` to the args of `spawn`?

northstar/src/runtime/process/mod.rs
```diff
@@ -53,17 +52,29 @@ pub(super) struct Launcher {
...
+    tx: EventTx,
 }
 
 impl fmt::Debug for Process {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("Process")
             .field("pid", &self.pid)
-            .field("checkpoint", &self.checkpoint)
+            .field("container", &self.container)
             .finish()
```
flxo, line 78: Add `cmd`, `argv` and `env`

northstar/src/runtime/ipc/channel.rs
```diff
@@ -1,93 +1,194 @@
-use std::{
-    io::{ErrorKind, Read, Write},
-    os::unix::prelude::{AsRawFd, RawFd},
-};
+use std::io::{Read, Write};
```
flxo, line 1: This file has mixed style trait bound definitions. I'd just one bound everywhere so it's probably good to declare that with the `impl` instead of a `where` clause.

northstar/src/runtime/process/fork.rs
alfonsoros88: yeah, just wanted to signal that this is some kind of wrapper

northstar/src/runtime/process/init.rs
alfonsoros88: yeah, its just some verbosity to avoid wandering where the `tx` came from inside the thread. Can remove if if you want.

northstar/src/runtime/process/init.rs
alfonsoros88: sure :)

northstar/src/runtime/process/init.rs
alfonsoros88: You are right! good catch

northstar/src/runtime/process/init.rs
alfonsoros88: Hmm yes this sounds good, like we could add a `InitApi::Shutdown` command to stop init explicitly from the runtime or when the sender is dropped :thinking:

northstar/src/runtime/process/mod.rs
alfonsoros88: Yep, i don't like it much either, i'll inline it.

northstar/src/runtime/ipc/channel.rs
alfonsoros88: now that you mention it... :rofl: yeah, will try to make it more consistent

northstar/src/runtime/process/init.rs
alfonsoros88: I just tested adding a `panic!` after the reaper thread is `unparked` and the whole process panics. I think this is just because we use that `panic = "abort"` policy.

northstar/src/runtime/ipc/channel.rs
```diff
@@ -1,93 +1,218 @@
...
+    pipe: std::io::BufReader<PipeRead>,
+    _marker: std::marker::PhantomData<S>,
+}
 
-    /// Raw fds for the rx and tx pipe
-    pub fn as_raw_fd(&self) -> (RawFd, RawFd) {
-        (self.tx.as_raw_fd(), self.rx.get_ref().as_raw_fd())
+impl<S> Receiver<S>
+where
+    S: Serialize + DeserializeOwned,
```
flxo, line 60: Why is `Serialize` needed here?

northstar/src/runtime/ipc/channel.rs
```diff
@@ -1,93 +1,218 @@
...
+
+/// Asynchronous [`Sender`]
+pub struct AsyncSender<S> {
+    pipe: AsyncPipeWrite,
+    _marker: std::marker::PhantomData<S>,
+}
+
+impl<S> AsyncSender<S>
+where
+    S: Serialize + DeserializeOwned,
```
flxo, line 162: Same as for the `Receiver`. Why is the `DeserializeOwned` bound needed here?

northstar/src/runtime/process/init.rs
flxo: Yes. Since this crate is a lib the `panic` field in the manifest must be set in the user of the northstar crate. Needs to be added to the docs.

northstar/src/runtime/process/init.rs
```diff
@@ -30,11 +38,130 @@ pub(super) struct Init {
...
+                }
+                Ok(WaitStatus::Continued(_)) | Ok(WaitStatus::Stopped(_, _)) => continue,
+                Err(Errno::ECHILD) => {
+                    // wait till a child process is spawned
+                    // unparked by the consumer thread after a Exec is received
+                    std::thread::park();
+                }
```
flxo, line 106: If this thread is started with the first fork in this init, this branch is not needed and the Thread can just gracefully exit.

northstar/src/runtime/process/mod.rs
```diff
@@ -102,28 +111,15 @@ impl Launcher {
...
-        fds.remove(&channel.as_raw_fd().0);
-        fds.remove(&channel.as_raw_fd().1);
-
-        // Ensure that the checkpoint fds are not in the fds set and untouched
-        let (checkpoint_runtime, checkpoint_init) = checkpoints();
-        fds.remove(&checkpoint_runtime.as_raw_fd().0);
-        fds.remove(&checkpoint_runtime.as_raw_fd().1);
+        // Two-way communication channels with the init process.
+        let mut comm_to_init = channel::Channel::<InitApi>::new();
+        let mut comm_from_init = channel::Channel::<InitMessage>::new();
```
flxo, line 122: The file descriptors must be removed unconditionally from `fds`

northstar/src/runtime/ipc/channel.rs
alfonsoros88: Is not _needed_ for only the receiver implementation but in general a pair of `Sender` `Receiver` will share the same type and therefore it has to implement both `Serialize` and `Deserialize`. That's the reason i set these bounds everywhere.

northstar/src/runtime/process/init.rs
alfonsoros88: yes that's true, but this way only spawns this thread once.

northstar/src/runtime/process/init.rs
```diff
@@ -30,11 +38,130 @@ pub(super) struct Init {
...
+                    // is started. Therefore no explicit drop (close) of _checkpoint_notify is needed here.
+                    panic!(
+                        "Execve: {:?} {:?}: {:?}",
+                        &path,
+                        &args,
+                        unistd::execve(&path, &args, &env)
+                    )
+                })
+                .expect("Failed to spawn child process");
+
```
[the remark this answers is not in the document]
alfonsoros88: let's discuss the spawn on demand vs single spawn park/unpark in the previous commant :car:

northstar/src/runtime/process/mod.rs
```diff
@@ -165,155 +158,157 @@ impl Launcher {
...
-                        checkpoint: Some(checkpoint_runtime),
-                        exit_status: Some(Box::new(exit_status_fut)),
-                    })
-                }
-                unistd::ForkResult::Child => {
-                    // Forget writing ends of io which are stdout, stderr. The `forget`
-                    // ensures that the file descriptors are not closed
-                    forget(io);
+        // Channel just for receiving the init pid from the intermediate process
+        let mut init_pid_channel = channel::Channel::<u32>::new();
```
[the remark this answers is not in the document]
alfonsoros88: I added this second pipe for sending just the pid to avoid having the `InitPid` variant in the `InitMessage` enum. I think it is a bit more independent that way :man_shrugging:  but is not a strong opinion

northstar/src/runtime/process/mod.rs
alfonsoros88: Yeah, I changed the `file_descriptor` member in `init.rs` to take a `skip` file descriptor list there. I decided to do it there since at this point in the code neither process has closed the unused ends.

northstar/src/runtime/ipc/pipe.rs
```diff
@@ -32,14 +32,16 @@ impl From<RawFd> for Inner {
...
+    unistd::pipe2(OFlag::O_CLOEXEC)
+        .map_err(from_nix)
+        .map(|(read, write)| {
+            (
+                PipeRead { inner: read.into() },
+                PipeWrite {
+                    inner: write.into(),
+                },
+            )
+        })
```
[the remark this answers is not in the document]
alfonsoros88: sure!

northstar/src/runtime/process/mod.rs
alfonsoros88: do you mean making `self.tx` and `Option` and then `take()` it here?

northstar/src/runtime/process/fork.rs
flxo: @marcmo It is.
The current version of the patchset doesn't use any `async` code beyond the fork.

northstar/src/runtime/ipc/channel.rs
flxo: Makes sense.

northstar/src/runtime/process/init.rs
flxo: There must be just one reaper thread. My proposal is to spawn the thread after the first fork. Keep the `JoinHandle` in an `Option`. This avoids one `waitpid` + park the thread that *will* (very very likely) happen because the thread create is probably faster than the fork.

northstar/src/runtime/process/mod.rs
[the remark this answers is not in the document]
flxo: Let's use the same channel. It's not that complicated and we safe the creation of another pipe.

northstar/src/runtime/process/mod.rs
flxo: No. I mean

```rust
async fn spawn(&mut self, tx: EventTx) -> Result<(), Error> {
```

because after the `spawn`/`start` sequence `Process::tx` is never used again.

northstar/src/runtime/process/init.rs
```diff
@@ -71,60 +198,11 @@ impl Init {
...
-                            }
-                            Err(nix::Error::EINTR) => continue,
-                            e => panic!("Failed to waitpid on {}: {:?}", child, e),
-                        }
-                    }
-                }
-                unistd::ForkResult::Child => {
-                    drop(exit_status_channel);
+        // Avoid clossing accidentally the channel file descriptors
+        self.file_descriptors(&[rt_tx.as_raw_fd(), rt_rx.as_raw_fd()]);
```
flxo, line 202: The fd leak test is really our friend ;-)

northstar/src/runtime/process/mod.rs
flxo: Got it.

northstar/src/runtime/process/init.rs
alfonsoros88: I'm not sure if I'm following, is the point just avoiding the first `waitpid` + `park`?

northstar/src/runtime/process/mod.rs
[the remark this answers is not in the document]
alfonsoros88: sure :man_shrugging:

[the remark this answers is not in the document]
alfonsoros88: Well, on a second though :sweat_smile: using the same channel for sending the init's pid will force using `&mut` for passing the channel to the init and that required a `'static` for the receiver as we move it into a thread. In this case I think we are better with the second pipe :smile:

northstar/src/runtime/process/mod.rs
alfonsoros88: so changing the `Process` trait, I guess it's possible but isn't it a bit weird from the interface point of view? I agree that there is some smell here, I think it originates from the channel ownership and who should take care of notify when the process exits, but perhaps we should address that in the next iteration.

northstar/src/runtime/process/fork.rs
alfonsoros88: renamed

northstar/src/runtime/process/init.rs
alfonsoros88: meaning? :sweat_smile: what's the problem here?

northstar/src/runtime/process/init.rs
```diff
@@ -40,11 +36,130 @@ pub(super) struct Init {
...
+/// This is the type used to communicate with the init process from the runtime.
+#[derive(Debug, Serialize, Deserialize)]
+pub enum InitMessage {
+    Forked { pid: i32 },
+    Exit { pid: i32, exit_status: ExitStatus },
+}
+
+/// Init commands
+#[derive(Debug, Serialize, Deserialize)]
+pub enum InitApi {
```
marcmo, line 48: why is InitApi an enum?

northstar/src/runtime/process/init.rs
```diff
@@ -40,11 +36,130 @@ pub(super) struct Init {
...
+                    }
+
+                    // Checkpoint fds are FD_CLOEXEC and act as a signal for the launcher that this child
+                    // is started. Therefore no explicit drop (close) of _checkpoint_notify is needed here.
+                    panic!(
+                        "Execve: {:?} {:?}: {:?}",
+                        &path,
+                        &args,
+                        unistd::execve(&path, &args, &env)
+                    )
```
marcmo, line 138: is it correct that this panic will never execute if everything works as expected?

northstar/src/runtime/process/init.rs
alfonsoros88: I thought there could be other potential commands for the init in addition to `Exec`. One option would be the `Shutdown` command. But at the moment it is not 100% clear for me how the shutdown  should behave.

flxo: I think there's no need for a 'Shutdown'. The `init`lives while any child is alive. With the exit of the last child the init exits. A "shutdown" is a signal to the process group. The `init` itself doesn't handle signals and the childs exit. With the last child init exits as well...

northstar/src/runtime/process/init.rs
flxo: Yes.

alfonsoros88: yes, it was a bit confusing for me too :sweat_smile: but after the `execve` is executed this is a different program and that `panic` should not be trigger unless something goes wrong.

northstar/src/runtime/process/init.rs
```diff
@@ -40,11 +37,137 @@ pub(super) struct Init {
     pub seccomp: Option<AllowList>,
 }
 
+/// This is the type used to communicate with the init process from the runtime.
+#[derive(Debug, Serialize, Deserialize)]
+pub(crate) enum InitMessage {
```
alfonsoros88, line 42: `pub(crate)` is not needed, the `init` module is private inside `process`.

alfonsoros88:
Closing because contained in #489

PR #458, opened 2021-11-30
Mounting cleanup and dm device fix

flxo:


northstar/src/runtime/mount.rs
```diff
@@ -67,8 +71,17 @@ impl std::fmt::Debug for MountControl {
...
     pub(super) async fn new() -> Result<MountControl, Error> {
+        debug!("Opening loop control");
         let lc = LoopControl::open().map_err(Error::LoopDevice)?;
+        debug!("Opening device mapper control");
         let dm = devicemapper::DM::new().map_err(Error::DeviceMapper)?;
+
+        let dm_version = dm
+            .version()
+            .map_err(Error::DeviceMapper)
+            .map(Version::from)?;
```
alfonsoros88, line 82: Why not make `dm.version()` directly return a `Version` type? this seems to be the only place where it is called. Or even simpler, if it is used just for logging, then just debug-print it?

northstar/src/runtime/mount.rs
```diff
@@ -166,19 +162,27 @@ async fn mount(
...
     let start = time::Instant::now();
+
+    if !target.exists() {
+        debug!("Creating mount point {}", target.display());
+        std::fs::create_dir_all(&target).map_err(|e| {
+            Error::Io(
+                format!("Failed to create directory {}", target.display()),
+                e,
+            )
+        })?;
```
alfonsoros88, line 183: Maybe we should try to implement that `Context` trait for these mount results too

northstar/src/util.rs
```diff
@@ -38,3 +39,32 @@ impl PathExt for Path {
...
+
+        dur.as_secs() as f64 * 1_000.0 + dur.subsec_nanos() as f64 / 1_000_000.0
+    }
+
+    fn as_fractional_micros(&self) -> f64 {
+        let dur: &time::Duration = self.borrow();
+
+        dur.as_secs() as f64 * 1_000_000.0 + dur.subsec_nanos() as f64 / 1_000.0
+    }
+}
```
alfonsoros88, line 70: fancy! but isn't there already a `Duration::as_secs_f64`?

alfonsoros88 (approved):
lgtm! nice to see many deletions!

northstar/src/runtime/mount.rs
flxo: Or even better get rid of `runtime::mount::Error` and use `runtime::error::Error`

northstar/src/runtime/mount.rs
```diff
@@ -304,109 +309,67 @@ fn dmsetup(
...
-    dm.table_load(
-        &id,
-        &table,
-        devicemapper::DmOptions::default().set_flags(devicemapper::DmFlags::DM_READONLY),
-    )
-    .map_err(Error::DeviceMapper)?;
-
-    debug!("Resuming device {}", dm_dev.display());
-    dm.device_suspend(&id, devicemapper::DmOptions::default())
+    let load = || {
```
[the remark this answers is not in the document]
flxo: This is on purpose to catch errors and try to revert the device creation instead of bailing out somewhere in between and leaking dm devices etc..

northstar/src/util.rs
flxo: Perfect. Thanks - didn't know.

northstar/src/runtime/mount.rs
alfonsoros88: yes! that would be great!

PR #464, opened 2021-12-06
Derive Init from Serialize, Deserialize

flxo:
As preparation for the refactoring with an initial fork prior to the runtime setup, the Init structure must be sendable via a pipe. This is best accomplished by deriving it from serde's Serialize and Deserialize.

### DIFF
--- a/northstar/src/seccomp/bpf.rs
+++ b/northstar/src/seccomp/bpf.rs
@@ -9,6 +9,7 @@ use bindings::{
 };
 use log::trace;
 use nix::errno::Errno;
+use serde::{Deserialize, Serialize};
 use std::{
     collections::{HashMap, HashSet},
     mem::size_of,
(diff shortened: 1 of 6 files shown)


northstar/src/seccomp/bpf.rs
```diff
@@ -181,10 +182,29 @@ fn check_platform_requirements() {
     compile_error!("Seccomp is not supported on Big Endian architectures");
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct SockFilter {
```
alfonsoros88, line 186: Not sure if this helps much, but `serde` has a `remote` feature to make external types serializable: https://serde.rs/remote-derive.html. With it, I think it would be possible to avoid replacing `sock_filter` with `SockFilter`.

alfonsoros88 (approved):
lgtm!

northstar/src/seccomp/bpf.rs
flxo: Thanks. The remote feature would have worked here. I didn't use it because I wanted to avoid to use type from the c bindings to widespread and prefer the use of an internal type.

PR #471, opened 2022-01-11
Add support for selinux types in manifest files

inorick:
Sextant now supports adding selinux contexts to container files during packing.

An example entry in a manifest file looks like this:

```
selinux:
  type: northstar_t
```

**TODO / open points:**
 - Retag existing files or create copies?
 - Test on hardware

### DIFF
--- a/northstar/src/npk/manifest.rs
+++ b/northstar/src/npk/manifest.rs
@@ -1,6 +1,6 @@
 use crate::{
     common::{name::Name, non_null_string::NonNullString, version::Version},
-    seccomp::{Seccomp, SyscallRule},
+    seccomp::{Seccomp, Selinux, SyscallRule},
 };
 use derive_more::Deref;
 use itertools::Itertools;
@@ -47,10 +47,12 @@ pub struct Manifest {
     pub env: Option<HashMap<NonNullString, NonNullString>>,
     /// Autostart this container upon northstar startup
     pub autostart: Option<Autostart>,
-    /// CGroup config
+    /// CGroup configuration
     pub cgroups: Option<cgroups::CGroups>,
     /// Seccomp configuration
     pub seccomp: Option<Seccomp>,
+    /// SELinux configuration
+    pub selinux: Option<Selinux>,
     /// Capabilities
     pub capabilities: Option<HashSet<Capability>>,
     /// String containing group names to give to new container
@@ -149,6 +151,30 @@ impl Manifest {
                 }
                 _ => Ok(()),
             })?;
+
+        // Check selinux context type
+        if let Some(selinux) = &self.selinux {
+            // Maximum length since at least Linux v3.7
+            // (https://elixir.bootlin.com/linux/v3.7/source/include/uapi/linux/limits.h)
+            const XATTR_SIZE_MAX: usize = 65536;
+
+            if selinux.context_type.len() >= XATTR_SIZE_MAX {
+                return Err(Error::Invalid(format!(
+                    "Selinux context os too long. Maximum length in {}",
+                    XATTR_SIZE_MAX
+                )));
+            }
+            if !selinux
+                .context_type
+                .chars()
+                .all(|c| c.is_ascii_alphanumeric() || c == '_')
+            {
+                return Err(Error::Invalid(
+                    "Selinux context type must consist of alphanumeric ASCII characters or '_'"
+                        .to_string(),
+                ));
+            }
+        }
 
         // Check seccomp filter
         const MAX_ARG_INDEX: usize = 5; // Restricted by seccomp_data struct
--- a/northstar/src/npk/npk.rs
+++ b/northstar/src/npk/npk.rs
@@ -1,13 +1,17 @@
+use crate::common::non_null_string::NonNullString;
 use crate::npk::{
     dm_verity::{append_dm_verity_block, Error as VerityError, VerityHeader, BLOCK_SIZE},
     manifest::{Bind, Manifest, Mount, MountOption},
 };
 use ed25519_dalek::{Keypair, PublicKey, SecretKey, SignatureError, Signer, SECRET_KEY_LENGTH};
 use itertools::Itertools;
+use nix::NixPath;
 use rand::rngs::OsRng;
 use semver::Version;
 use serde::{Deserialize, Serialize};
 use sha2::{Digest, Sha256};
+use std::ffi::{CStr, CString};
+use std::os::unix::ffi::OsStrExt;
 use std::{
     fmt, fs,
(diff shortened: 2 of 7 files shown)


northstar/src/npk/npk.rs
```diff
@@ -748,6 +758,92 @@ fn sign_hashes(key_pair: &Keypair, hashes_yaml: &str) -> String {
...
+fn set_selinux_contexts(root: &Path, context_type: &NonNullString) -> Result<(), Error> {
+    for dir_entry in WalkDir::new(root)
+        .follow_links(false)
+        .into_iter()
+        .filter_map(Result::ok)
+    {
+        let path = dir_entry.path();
+        if let Ok(path_c) = std::ffi::CString::new(path.as_os_str().as_bytes()) {
+            // Read existing selinux context
+            let mut context_buf = [0u8; 256];
```
flxo: Why 256? Is this the max length of xattr values?

northstar/src/npk/npk.rs
```diff
@@ -748,6 +758,92 @@ fn sign_hashes(key_pair: &Keypair, hashes_yaml: &str) -> String {
     )
 }
 
+fn set_selinux_contexts(root: &Path, context_type: &NonNullString) -> Result<(), Error> {
```
flxo, line 761: Short description what is done in here would be great.

northstar/src/npk/manifest.rs
```diff
@@ -47,10 +47,12 @@ pub struct Manifest {
     pub env: Option<HashMap<NonNullString, NonNullString>>,
     /// Autostart this container upon northstar startup
     pub autostart: Option<Autostart>,
-    /// CGroup config
+    /// CGroup configuration
     pub cgroups: Option<cgroups::CGroups>,
     /// Seccomp configuration
     pub seccomp: Option<Seccomp>,
+    /// SELinux configuration
+    pub selinux: Option<Selinux>,
```
flxo, line 55: Can we add some checks in `Manifest::verify`? There are for sure restrictions on the label

northstar/src/npk/npk.rs
```diff
@@ -748,6 +758,92 @@ fn sign_hashes(key_pair: &Keypair, hashes_yaml: &str) -> String {
...
+                    "Failed to convert old selinux context to C string: {}",
+                    &path.display()
+                )));
+            }
+        } else {
+            return Err(Error::Selinux(format!(
+                "Failed to convert path to C string: {}",
+                &path.display()
+            )));
+        }
```
flxo: Is it possible to flatten this by creating some bindings. This is really hard to read and the errors are very similar. Maybe move it to a fn and just use a single error message.

northstar/src/seccomp/types.rs
```diff
@@ -23,6 +23,14 @@ pub struct Seccomp {
     pub allow: Option<HashMap<NonNullString, SyscallRule>>,
 }
 
+/// SELinux configuration
+#[derive(Clone, Eq, PartialEq, Debug, Serialize, Deserialize, JsonSchema)]
+pub struct Selinux {
+    /// Explicit list of allowed syscalls
+    #[serde(rename = "type")]
+    pub context_type: NonNullString,
```
flxo, line 31: Can we remove the `rename`? Either use `context_type` in them yaml or `r#type` as field name. I don't know what's the better choice.

northstar/src/npk/npk.rs
inorick: We now use another call to getxattr to determine the proper size.

northstar/src/npk/npk.rs
inorick: I broke the steps down into sub-functions that only return one error each.

northstar/src/seccomp/types.rs
inorick: context_type it is.

northstar/src/npk/manifest.rs
```diff
@@ -150,6 +152,20 @@ impl Manifest {
                 _ => Ok(()),
             })?;
 
+        // Check selinux context type
+        if let Some(selinux) = &self.selinux {
```
flxo, line 156: Is there a length restriction on the labels?

northstar/src/npk/npk.rs
```diff
@@ -748,6 +758,107 @@ fn sign_hashes(key_pair: &Keypair, hashes_yaml: &str) -> String {
...
+    if components.len() >= 3 {
+        components[2] = context_type;
+        CString::new(components.join(":").as_bytes()).ok()
+    } else {
+        None
+    }
+}
+
+/// Write selinux context to a path
+fn write_selinux_context(path_c: &CString, context: &CString) -> bool {
```
flxo, line 849: Why is a `CString` used here? I would have expected a `&std::path::Path` or at least a `&CStr`.
`&CString` doesn't make sense.

northstar/src/npk/npk.rs
```diff
@@ -748,6 +758,107 @@ fn sign_hashes(key_pair: &Keypair, hashes_yaml: &str) -> String {
...
+    let name_c = CString::new("security.selinux").unwrap();
```
flxo, line 850: This can be a `const`

northstar/src/npk/npk.rs
```diff
@@ -748,6 +758,107 @@ fn sign_hashes(key_pair: &Keypair, hashes_yaml: &str) -> String {
...
+    let context = CStr::from_bytes_with_nul(&context_buf[..size.try_into().ok()?])
+        .ok()?
+        .to_str()
+        .ok()?
+        .to_string();
+    Some(context)
+}
+
+/// Replace the tag of a selinux context and return the updated version
+fn replace_selinux_context_type(context: &str, context_type: &NonNullString) -> Option<CString> {
```
flxo, line 838: I more and more doubt that `NonNullString` was a good idea or not...

northstar/src/npk/npk.rs
flxo, line 849: Return a `Result` instead of bool here.

northstar/src/npk/manifest.rs
inorick: I think we can use
```
#define XATTR_SIZE_MAX 65536
```
(https://www.spinics.net/lists/selinux/msg09373.html)

northstar/src/npk/npk.rs
inorick: CString::new is not a const function :/

flxo: Right. But we still could improve it to use https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_bytes_with_nul_unchecked

this avoids a copy.

Issue #474, opened 2022-01-14
Use `cargo deny`


adoerr:
# Use `cargo-deny`

## What is the feature about ?
[cargo-deny](https://github.com/EmbarkStudios/cargo-deny) is a tool for auditing `Cargo.lock` files for crates with security vulnerabilities, limit the usage of particular dependencies, their licenses, sources to download from, detect multiple versions of same packages in the dependency tree and more.

As such [cargo-deny](https://github.com/EmbarkStudios/cargo-deny)  will provide an **automated** and **tool based** solution to:

- make sure our dependencies only use endorsed OSS licences
- we are aware of [RustSec](https://github.com/RustSec/advisory-db) advisories affecting our code base as well as our dependencies
- we are aware of deprecated dependencies
- keep dependency versions aligned across the dependency tree

## Why is it needed ?

- create initial `deny.toml` file
- adapt default `deny.toml` file, address existing issues
- add GitHub Action for CI integration

## What is expected
Start using [cargo-deny](https://github.com/EmbarkStudios/cargo-deny) for local builds. Once it is properly configured and adapted, integrate into CI

flxo:
Thanks.

We have a `cargo-license` check internally but moving to `cargo deny is a good idea` since it checks more stuff than just  the licenses.

adoerr:
I will do a PR with an initial `deny.toml` file, so we have something to get going :rocket:

flxo:
> I will do a PR with an initial `deny.toml` file, so we have something to get going 🚀

Sorry @adoerr I didn't see this comment before setting up the PR.

PR #482, opened 2022-01-14
Initial deny.toml file

adoerr:
This is an initial `deny.toml` file in order to get going with `cargo deny`.

Basically it is the same file created by `cargo deny init` plus an initial allow list for OSS licenses and some formatting.

see: #474


flxo:
Yay - nice to see that cariad folks contribute. Thanks!

Closing this in favour of #481 which is almost identical + the github org and ad CI job.

PR #483, opened 2022-01-14
Replace rand dependency with rand_core

flxo:
The dalek crate has a nasty rand 0.7 trait in it's api to generte keypairs. Work around the need for this specific rand version by using the latest rand_core random fn and construct the keypair manually.


inorick (commented):
Looks good to me. There is also version 4.0.0-pre.1 of the curve25519-dalek crate that has an updated dependency to rand 0.8.

PR #486, opened 2022-01-18
Selinux support on unsupported hosts

inorick:
When setting selinux contexts, check for an existing context to replace first. Otherweise, write a new selinux context.


flxo (commented):
LGTM!

PR #489, opened 2022-01-21
Forker process

arosds:
Spawning container processes is delegated to a separated 'forker'
process. The runtime communicates with this process through a unix
domain socket.

With this change, northstar's runtime can execute in multithreaded mode
without the danger of the libc deadlocking issue. The 'forker' process 
must consequently be single threaded.

Additionally, this patch introduces a small API for the container Init
processes that enables the request to start new processes inside the
container. This is a prerequisite to #454.

## Additional details

- Northstar version is bumped to `0.7.0-dev` (see #496 )
- Panic if the forker process exits unexpectedly

    If the forker process dies for whatever reason, it is not possible
    to recoverable and the runtime bails out.

- Do not limit the number of threads of the runtime in demo main
- Parallel loading of NPKs from disk

    The loading of NPKs from disk is slow and blocking. Spawn a thread
    for each NPK in order to speed up the boring parsing of the NPK
    headers.

- Replace manifest IO Pipe with Log

    The 'pipe' option for the container output is removed from the
    manifest. The option 'log' is renamed to 'pipe'. A new option
    'discard' is added for the output.

- Refactor container IO handling

    When the container IO configuration in the manifest indicates that
    any of `stdout` or `stderr` is to be 'piped', a socket is used to
    receive the output from the container. On the other side, the
    runtime uses a `async` task to forward the incoming output from the
    socket to the runtime log.

- Pipes removed and replaced with sockets


northstar-tests/tests/tests.rs
```diff
@@ -413,39 +459,18 @@ test!(exitcodes, {
                 ..
             } if code == c)
         };
-        runtime.assume_notification(n, 5).await?;
-    }
-    runtime.shutdown().await
-});
-
-// Open many connections to the runtime
-test!(open_many_connections_to_the_runtime_and_shutdown, {
```
alfonsoros88, line 422: noticed this test got removed, is that ok?

images/container-startup.puml
```diff
@@ -1,33 +1,70 @@
...
-Runtime -> Runtime: Read exit status from pipe or waitpid on pid of init
+
+Init -> Init: waitpid: Exit status of container
+Init -> Forker: Container exit status
 destroy Init
 
+Forker -> Runtime: Container exit status
+Runtime -> Runtime: Stop PTY thread
+Runtime -> Runtime: Destroy cgroups
+Runtime -> Runtime: Destroy cgroups
```
alfonsoros88, line 67: isn't this duplicated?

northstar-tests/tests/tests.rs
flxo: The shutdown of the RT is no longer possible from within a test. Probably that's the reason why I removed this test. Hm. Think this test originally has been added because we had connections that stalled upon a RT shutdown.

images/container-startup.puml
flxo: Yes.

PR #497, opened 2022-01-26
Move npk version information to a file

flxo:
Parsing the comment field is slow because zip needs to search for the end of the comment field. This is a never ending seek + read a byte sequence. A dedicated file with the meta information can be read once which is faster.

### DIFF
--- a/northstar/src/npk/npk.rs
+++ b/northstar/src/npk/npk.rs
@@ -31,6 +31,7 @@ const UNSQUASHFS: &str = "unsquashfs";
 
 // File name and directory components
 const FS_IMG_NAME: &str = "fs.img";
+const METADATA_NAME: &str = "metadata.yaml";
 const MANIFEST_NAME: &str = "manifest.yaml";
 const SIGNATURE_NAME: &str = "signature.yaml";
 const FS_IMG_BASE: &str = "fs";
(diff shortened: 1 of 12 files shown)


northstar/src/npk/npk.rs
```diff
@@ -70,7 +71,7 @@ pub enum Error {
         error: SignatureError,
     },
     #[error("Comment malformed: {0}")]
```
inorick, line 73: Should say "Metadata malformed"

northstar/src/npk/npk.rs
```diff
@@ -31,6 +31,7 @@ const UNSQUASHFS: &str = "unsquashfs";
 
 // File name and directory components
 const FS_IMG_NAME: &str = "fs.img";
+const METADATA_NAME: &str = "metadata.yaml";
```
inorick, line 34: Are we going to store more than a comment in metadata? Now it sounds like we store all metadata in its own file but we also have a manifest which contains metadata.

flxo: The metadata just contains the version of manifest. That's why we decided to store it in the comment field.
Unfortunately reading the comment field is very expensive because a zip implementation has to seek for the end of the comment manualy.

PR #501, opened 2022-02-05
Exec subcommand

rosds:
This branch is on top of #489, it add the `exec` sub-command to `nstar`

### DIFF
--- a/northstar/src/api/model.rs
+++ b/northstar/src/api/model.rs
@@ -8,6 +8,7 @@ use serde::{de::Visitor, Deserialize, Serialize, Serializer};
 use std::{
     collections::{HashMap, HashSet},
     fmt,
+    path::PathBuf,
 };
 
 /// Console configuration
@@ -124,6 +125,13 @@ pub enum Request {
         Vec<NonNulString>,
         HashMap<NonNulString, NonNulString>,
     ),
+    Exec {
+        container: Container,
+        env: Vec<NonNulString>,
+        path: NonNulString,
+        args: Vec<NonNulString>,
+        pty: Option<PathBuf>,
+    },
     TokenCreate(Vec<u8>, Vec<u8>),
     TokenVerify(Token, Vec<u8>, Vec<u8>),
     Umount(Vec<Container>),
--- a/northstar/src/runtime/fork/forker/impl.rs
+++ b/northstar/src/runtime/fork/forker/impl.rs
@@ -1,22 +1,17 @@
-use super::{
-    init,
-    init::Init,
-    messages::{Message, Notification},
-    util::fork,
-};
+use super::{init, init::Init, messages::Message, util::fork};
 use crate::{
     common::{container::Container, non_nul_string::NonNulString},
     debug,
     runtime::{
-        fork::util::{self, set_log_target},
+        fork::{
+            util::{self, set_log_target},
+            Notification,
+        },
         ipc::{self, owned_fd::OwnedFd, socket_pair, AsyncMessage, Message as IpcMessage},
         ExitStatus, Pid,
     },
 };
-use futures::{
-    stream::{FuturesUnordered, StreamExt},
-    Future,
-};
+use futures::{stream::FuturesUnordered, Future, FutureExt, StreamExt};
 use itertools::Itertools;
 use nix::{
     errno::Errno,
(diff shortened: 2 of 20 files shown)


northstar/src/api/model.rs
```diff
@@ -141,6 +141,15 @@ pub enum Request {
         container: Container,
         signal: i32,
     },
+    /// Start a new process in a container
+    Exec {
+        /// Container
+        container: Container,
+        /// Command and arguments
+        cmd: Vec<NonNullString>,
```
flxo, line 149: Should we add env variables?

northstar/src/runtime/fork/forker/impl.rs
```diff
@@ -22,18 +23,169 @@ use nix::{
...
+
+                                let pty =  std::fs::OpenOptions::new()
+                                    .read(true)
+                                    .write(true)
+                                    .open(slave)
+                                    .expect("Failed to open file");
+
+                                stream.send_fd(pty).await.expect("Failed to send pty fd");
+                            }
+                            ProcessOutput::Files { stdout, stderr } => {
```
flxo, line 132: Maybe rename `File` to `StdIo`

northstar/src/runtime/fork/forker/impl.rs
```diff
@@ -22,18 +23,169 @@ use nix::{
...
+                                stream
+                                    .send(init::Message::Exec { path, args, env, pty: false })
+                                    .await
```
flxo, line 135: I see that `Message::Exec` is sent in two flavours. One with a pty and one with stdio. The check in init which type it is is based on the the `pty` flag. 
Check the `Message` definition of the communication between the forker and the runtime. Replace the `pty` flag with a enum that contains `Option<File etc>`. Add a `send` method (like `forker::impl::send` that knows when to send the pty etc..

northstar/src/runtime/fork/forker/messages.rs
```diff
@@ -25,12 +25,18 @@ pub enum Message {
...
+        output: ProcessOutput,
     },
     ExecResult, // TODO
 }
 
+/// Process output
+#[derive(Debug, Serialize, Deserialize)]
+pub enum ProcessOutput {
+    Files { stdout: PathBuf, stderr: PathBuf },
+    Pty { slave: PathBuf },
```
flxo, line 37: This is good place to document the two flavours and when each one is used.

northstar/src/runtime/fork/init/mod.rs
```diff
@@ -130,26 +235,42 @@ impl Init {
...
+                                    panic!("Failed to TIOCSCTTY");
+                                }
+                                unistd::dup2(slave, nix::libc::STDIN_FILENO)
+                                    .expect("Failed to dup2");
+                                unistd::dup2(slave, nix::libc::STDOUT_FILENO)
+                                    .expect("Failed to dup2");
+                                unistd::dup2(slave, nix::libc::STDERR_FILENO)
+                                    .expect("Failed to dup2");
+                                unistd::close(slave).expect("Failed to close");
+                            }
```
flxo, line 272: What's the reason to distinguish the io and pty cases? What would be bad with using *always* a pty?

northstar/src/runtime/fork/init/mod.rs
```diff
@@ -166,59 +287,20 @@ impl Init {
                     .expect("Failed to spawn child process");
 
                     // close fds
-                    drop(console_stream);
-                    unistd::close(stdout).expect("Failed to close stdout");
-                    unistd::close(stderr).expect("Failed to close stderr");
+                    match out {
+                        Out::Pipe { stdout, stderr } => {
+                            unistd::close(stdout).expect("Failed to close stdout");
+                            unistd::close(stderr).expect("Failed to close stderr");
```
flxo, line 293: Replace the `RawFd` in `Out` with something that closes on drop. This manual fd handling is as dangerous as a `malloc` ;-)

tools/nstar/src/main.rs
```diff
@@ -342,6 +355,61 @@ async fn main() -> Result<()> {
...
+                .context("Failed to connect")?;
+
+            let version = or_version(&name, version, &mut client).await?;
+            let cmd: Result<Vec<_>> = cmd
+                .into_iter()
+                .map(|c| NonNullString::try_from(c).context("Invalid arg"))
+                .collect();
+
+            // TODO Avoid the nix dependency by opening the PTY in the runtime and sending the
+            // master FD through the client socket.
```
flxo, line 372: This doesn't work if nstar is connected via TCP. Right?

tools/nstar/src/main.rs
```diff
@@ -342,6 +355,61 @@ async fn main() -> Result<()> {
...
+        Subcommand::Exec { name, version, cmd } => {
+            // Connect
+            let mut client = Client::new(io, None, opt.timeout)
+                .await
+                .context("Failed to connect")?;
+
+            let version = or_version(&name, version, &mut client).await?;
+            let cmd: Result<Vec<_>> = cmd
+                .into_iter()
+                .map(|c| NonNullString::try_from(c).context("Invalid arg"))
```
flxo, line 368: Oh. Not sure if this `NonNullString` was a good idea....

tools/nstar/src/main.rs
```diff
@@ -342,6 +355,61 @@ async fn main() -> Result<()> {
...
+                    let n = unistd::read(nix::libc::STDIN_FILENO, &mut buf).unwrap();
+                    unistd::write(master, &buf[..n]).unwrap();
+                }
+
+                if fd_set.contains(master) {
+                    let mut buf = [0; 512];
+                    let n = unistd::read(master, &mut buf).unwrap();
+                    unistd::write(nix::libc::STDOUT_FILENO, &buf[..n]).unwrap();
+                }
+            }
```
flxo, line 410: ptys and Tokio behave special. I tried to use it the straight forward way in the runtime for the output to log forwarding. Not sure yet what's the specialties are here.

northstar/src/api/model.rs
alfonsoros88: yes! added `-e,--env` flag for passing env variables :h

tools/nstar/src/main.rs
alfonsoros88: Good point, in any case `exec` only works if executed locally :sweat:

northstar/src/runtime/fork/init/mod.rs
alfonsoros88: There is a subtle detail between the two. To set the pty as the controlling tty of the process, it has to be placed in a separated session with `setsid()`. In that case, our `kill` command will fail to kill the process with the `-pid` trick because that only works for processes in the same session :weary:

tools/nstar/src/main.rs
flxo: Why should it? If the RT takes care for bridging between the connection and the pty, TCP should be also possible.
Only if we create the  pty in nstar this would be a problem.

northstar/src/runtime/fork/forker/messages.rs
arosds: closing discussion because this got replaced by the 3 fd array

northstar/src/runtime/fork/init/mod.rs
```diff
@@ -112,13 +113,113 @@ impl Init {
...
+                        std::thread::park();
+                    }
+                    e => panic!("Failed to waitpid: {:?}", e),
+                }
+            })
+        };
+
+        {
+            let mut stream = stream.try_clone().expect("Failed to clone stream");
+            std::thread::spawn(move || loop {
```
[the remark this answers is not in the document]
arosds: the thread was removed and replaced with a `select`

tools/nstar/src/main.rs
arosds: moved the pty opening to the runtime and use an async task to forward the input/output from the client connection to the container.

flxo:
Closing this PR as it's not possible to rebase. Thanks @rosds

PR #505, opened 2022-02-07
Replace name and version `nstar` arguments with single string argument formatted as `name[:version]`

alfonsoros88:


flxo (approved):
Love it!

flxo:
@alfonsoros88 Increase the patch version in the `nstar` manifest.

PR #519, opened 2022-02-22
Update README.md

KloolK:
removed some text that looked like leftovers from previous editing


arosds (approved):
Yes, definitely looks like a slip! thank you!

PR #520, opened 2022-02-22
Call pivot_root to set container's root mount

### LINKED ISSUE #435: Replace `chroot` with `pivot_root`
Replace the considered insecure `chroot` call with [pivot_root](https://man7.org/linux/man-pages/man2/pivot_root.2.html) which applies to the whole mount namespace.

rosds:
fixes #435

`pivot_root` takes 2 arguments: `new_root` and `old_root`. In case of
success, `new_root` is set as the root mount for the current and new
processes and threads. Due to the restrictions described in
`pivot_root(2)`, `old_root` has to be underneath `new_root`. This means,
that we need a mounting point inside the `rootfs` of every container.
This patch adds `/mnt` to the container image packed with `squashfs` to
be used as the `old_root` parameter.


northstar/src/npk/npk.rs
```diff
@@ -735,6 +735,8 @@ fn gen_pseudo_files(manifest: &Manifest) -> Result<NamedTempFile, Error> {
                 }
             }
         })
+        // This directory is used by `pivot_root` as the `old-root` parameter
+        .chain(pseudo_directory(Path::new("/mnt"), 755))
```
flxo, line 739: What about naming to `old-root`?

flxo (commented):
This patch leaves a empty `mnt` dir in every fs. Collisions with user directories are not checked and I can image the case that someone creates a container that mounts a resource on `/mnt` ;-)

I understand that `old_root` must be underneath the root. That's a pity for this `ro` fs.

northstar-runtime/src/runtime/fork/init/mod.rs
```diff
@@ -346,12 +348,56 @@ impl Init {
...
+
+        // Unmount the original root directory which was stacked on top of new root directory
+        // MNT_DETACH makes the mount point unavailable to new accesses, but waits till the original mount point
+        // to be free of activity to actually unmount
+        // see https://man7.org/linux/man-pages/man2/umount2.2.html for more information
+        mount::umount2("/", mount::MntFlags::MNT_DETACH).expect("failed to umount old root");
+
+        // Change directory to root
+        unistd::fchdir(newroot).expect("failed to fchdir");
+        unistd::close(newroot).expect("failed to cloe");
```
inorick, line 399: Typo: "failed to close"

northstar/src/npk/npk.rs
flxo: Outdated. The additional `met` is not needed if both `pivot_root` arguments are the new root.

flxo:
Fixes #784

PR #539, opened 2022-04-01
Add token signing and verification

inorick:
Tokens contain the following fields:
- timestamp: seconds since very first token
- Blake3 based MAC of timestamp

Currently missing data for MAC computation:
- Identifiers of source and destination containers

### DIFF
--- a/northstar/src/runtime/token.rs
+++ b/northstar/src/runtime/token.rs
@@ -1,0 +1,118 @@
+use blake3::Hash;
+use lazy_static::lazy_static;
+use rand_core::{OsRng, RngCore};
+use std::ops::Add;
+use std::time::{Duration, Instant};
+
+// Tokens are valid for one minute
+const TOKEN_EXPIRED_THRESHOLD: Duration = Duration::from_secs(60);
+
+lazy_static! {
+    static ref MAC_KEY: [u8; 32] = {
+        let mut key = [0u8; 32];
+        OsRng.fill_bytes(&mut key);
+        key
+    };
+    static ref FIRST_TOKEN_TIME: Instant = Instant::now();
+}
+
+pub type Timestamp = u64;
+
+pub struct Token {
+    time: Timestamp,
+    mac: blake3::Hash,
+}
+
+impl Token {
+    pub fn new(usage: &str) -> Token {
+        let first = *FIRST_TOKEN_TIME; // Do not inline to ensure earliest initialization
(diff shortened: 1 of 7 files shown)


northstar/src/runtime/token.rs
```diff
@@ -0,0 +1,123 @@
...
+        // Check expiration
+        let token_time = FIRST_TOKEN_TIME.add(Duration::from_secs(self.time));
+        if token_time < *FIRST_TOKEN_TIME {
+            return false; // Token is older than the very first one
+        }
+        let token_age = Instant::now().duration_since(token_time);
+        token_age <= TOKEN_EXPIRED_THRESHOLD
+    }
+
+    pub fn write(&self, output: &mut [u8]) -> Result<(), Error> {
```
flxo, line 58: Can we update this to take a `T: std::io::Write` and return `std::io::Result`?

northstar/src/runtime/token.rs
```diff
@@ -0,0 +1,123 @@
...
+            return false;
+        }
+
+        // Check expiration
+        let token_time = FIRST_TOKEN_TIME.add(Duration::from_secs(self.time));
+        if token_time < *FIRST_TOKEN_TIME {
+            return false; // Token is older than the very first one
+        }
+        let token_age = Instant::now().duration_since(token_time);
+        token_age <= TOKEN_EXPIRED_THRESHOLD
```
flxo, line 55: nit: this would be easier to read with an `else` instead of the `return`.

northstar/src/runtime/token.rs
```diff
@@ -0,0 +1,123 @@
...
+
+pub type Timestamp = u64;
+
+pub struct Token {
+    time: Timestamp,
+    mac: blake3::Hash,
+}
+
+impl Token {
+    pub fn new(application: &str) -> Token {
```
flxo, line 35: I'm struggling with the term `application` here. Maybe we can use something more generic here?

inorick: How about 'usage' or general 'info'?

northstar/src/runtime/token.rs
inorick: Line 54 panics if we do not bail on line 52. I am not sure how it can be rearranged.

flxo:
The license check fails on `BSD-2-Clause` and `CC0-1.0`. Safe to add to the list of allowed licenses?

flxo:
@inorick rebased.

PR #542, opened 2022-04-01
Remove obsolete sha256 of fs.img

flxo:
Remove leftover sha256 in the npk signature file. Minor cleanup.


northstar/src/npk/npk.rs
```diff
@@ -603,56 +608,53 @@ fn read_keypair(key_file: &Path) -> Result<Keypair, Error> {
...
 
-fn gen_hashes_yaml(
-    manifest: &Manifest,
-    fsimg: &Path,
-    fsimg_size: u64,
-    hash: &[u8],
-) -> Result<String, Error> {
-    // Create hashes YAML
-    let mut sha256 = Sha256::new();
-    sha2::digest::Update::update(&mut sha256, manifest.to_string().as_bytes());
```
inorick, line 614: The hash that is removed by the PR is actually the hash of the manifest file. It is not covered by dm_verity. Perhaps we were too eager to call it obsolete?

flxo: The manifest hash calculation is just moved and not removed. See the argument list of `hash_yaml`.

inorick (approved):

PR #575, opened 2022-04-19
Extend container token handling

flxo:
* Add token debug functionality to `nstar`
* Add `user` and `target` items to token generation and checks


northstar/src/runtime/token.rs
```diff
@@ -4,197 +4,223 @@ use hmac::{
...
-    fn calculate_hmac(time: &Timestamp, usage: &str) -> Hmac {
-        let mut hasher = HmacSha256::new_from_slice(MAC_KEY.as_slice())
-            .expect("Failed to create SHA-256 hasher");
-        hasher.update(&time.to_be_bytes());
-        hasher.update(usage.as_bytes());
-        Hmac(hasher.finalize())
-    }
+#[cfg(not(test))]
+fn now() -> time::Duration {
+    time::Duration::from_secs(time::Instant::now().elapsed().as_secs())
```
inorick, line 86: I think `time::Instant::now().elapsed()` always returns zero meaning the whole function always returns a duration of zero.

northstar/src/runtime/token.rs
```diff
@@ -4,197 +4,223 @@ use hmac::{
...
+    )
 }
 
-#[derive(PartialEq)]
-struct Hmac(CtOutput<HmacSha256>);
+fn calculate_hmac(time: &time::Duration, user: &[u8], target: &[u8], shared: &[u8]) -> Hmac {
+    let mut hasher =
+        HmacSha256::new_from_slice(MAC_KEY.as_slice()).expect("Failed to create SHA-256 hasher");
+    hasher.update(&time.as_millis().to_be_bytes());
+    hasher.update(user);
```
inorick, line 104: We should probably update the hasher with a delimiter character between the update calls. Otherwise, a token for a user with ID "abc" and a target with ID "xyz" is the same as one for the IDs "a" and "bcxyz".
Whatever delimiter character we chose (e.g. null byte), it should not be allowed to be part of a regular ID.

northstar/src/runtime/token.rs
flxo: Aeh. Right. Will fix that!

northstar/src/runtime/token.rs
flxo: Interesting. Today we allow anything that fulfils `AsRef<[u8]>` which can contain *anything* including a nul. We could take some (or all) bytes of the key?

inorick: After thinking about this some more, I think we could HMAC both IDs individually and feed the outputs into the update calls. We can merge this first and I'll change it afterwards.

flxo: Ah ok. Interesting. I added some parts of the key in between the two slices, but your approach sounds way cleaner.

inorick (approved):

PR #576, opened 2022-04-19
Simplify forker impl

flxo:


northstar/src/runtime/fork/forker/impl.rs
```diff
@@ -98,10 +76,16 @@ pub async fn run(stream: StdUnixStream, notifications: StdUnixStream) -> ! {
...
                             let (response, exit) = exec(init, container, path, args, env, io).await;
-                            tx.send(exit).await.ok();
+
+                            // Add exit status future of this exec request
+                            exits.push(exit);
+
+                            // Send the result of the exec request to the runtime
                             stream.send(response).await.expect("failed to send response");
                         } else {
+                            // TODO: This is a bug, we should never get here
```
rosds, line 88: Then we should probably call `expect` after the `.remove` and not handle this impossible case

rosds (approved):
Great! :) I hope there was no real reason for the extra async task :sweat_smile:

northstar/src/runtime/fork/forker/impl.rs
flxo: Agree. That situation cannot be handled in any good way. I updated the two branches to panic if something is wrong.

Issue #573, opened 2022-04-19
Rate limit console requests


flxo:
The requests from console sockets are processed as received. A misbehaving client may have an impact on the runtime. 

Rate limit the number of requests from a console connection. The limit shall be configurable in the `console` sections of the `runtime` configuration. Allow the rate limit to be configured for console connections for containers.

The [StreamExt::throttle](https://docs.rs/tokio-stream/0.1.8/tokio_stream/trait.StreamExt.html#method.throttle) can do the heavy lifting.

PR #582, opened 2022-04-20
Limit incoming message rate in Framed

rosds:
solves #573

### DIFF
--- a/northstar-tests/src/runtime.rs
+++ b/northstar-tests/src/runtime.rs
@@ -10,7 +10,10 @@ use northstar::{
         model::{Container, ExitStatus, Notification},
     },
     common::non_nul_string::NonNulString,
-    runtime::{config, Runtime as Northstar},
+    runtime::{
+        config::{self, ConsoleConfiguration},
+        Runtime as Northstar,
+    },
 };
 use std::convert::{TryFrom, TryInto};
 use tempfile::{NamedTempFile, TempDir};
@@ -57,14 +60,16 @@ impl Runtime {
         let consoles = [
             (
                 console_full(),
-                config::Console {
-                    permissions: config::ConsoleConfiguration::full(),
+                ConsoleConfiguration {
+                    permissions: config::Permissions::full(),
+                    max_requests_per_sec: None,
                 },
             ),
             (
                 console_none(),
-                config::Console {
-                    permissions: config::ConsoleConfiguration::default(),
+                ConsoleConfiguration {
+                    permissions: config::Permissions::default(),
+                    max_requests_per_sec: None,
                 },
             ),
         ]
--- a/northstar/src/api/codec.rs
+++ b/northstar/src/api/codec.rs
@@ -1,19 +1,25 @@
 use super::model;
-use std::io::ErrorKind;
-use tokio::io::{self, AsyncRead, AsyncWrite};
+use futures::StreamExt;
+use std::{io::ErrorKind, time::Duration};
+use tokio::{
+    io::{self, AsyncRead, AsyncWrite},
+    time::Instant,
+};
 use tokio_util::codec::{Decoder, Encoder};
 
 /// Newline delimited json codec for api::Message that on top implements AsyncRead and Write
 pub struct Framed<T> {
     inner: tokio_util::codec::Framed<T, Codec>,
+    rate_limitter: Option<TimeWindowCounter>,
 }
 
-impl<T: AsyncRead + AsyncWrite> Framed<T> {
+impl<T: AsyncRead + AsyncWrite + Unpin> Framed<T> {
     /// Provides a [`Stream`] and [`Sink`] interface for reading and writing to this
     /// I/O object, using [`Decoder`] and [`Encoder`] to read and write the raw data.
     pub fn new(inner: T) -> Framed<T> {
         Framed {
             inner: tokio_util::codec::Framed::new(inner, Codec::default()),
+            rate_limitter: None,
         }
     }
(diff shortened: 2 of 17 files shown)


northstar-tests/src/runtime.rs
```diff
@@ -59,12 +59,14 @@ impl Runtime {
                 console_full(),
                 config::Console {
                     permissions: config::ConsoleConfiguration::full(),
+                    max_request_per_sec: None,
```
flxo, line 62: You have to know a little of Rust to not misinterpret this ;-)

flxo (commented):
Can we count the messages in the last second in order to not have this this fixed delay between *each* message? There's nothing wrong with two subsequent messages at once as long as the configured rate is not exceeded.

northstar/src/api/codec.rs
```diff
@@ -77,6 +101,47 @@ impl Encoder<model::Message> for Codec {
...
+            } else {
+                break;
+            }
+        }
+
+        self.queue
+            .iter()
+            .rev()
+            .nth(self.count - 1)
+            .map(|instant| *instant + self.window - now)
```
flxo, line 141: Nice!
A comment would be also nice ;-)

northstar/src/runtime/console.rs
```diff
@@ -228,7 +228,7 @@ impl Console {
 
         // Looks good - send ConnectAck
         let connect = model::Connect::Ack {
-            configuration: configuration.clone(),
+            configuration: permissions.clone(),
```
flxo, line 231: Why exclude the rate limit?

northstar/src/api/codec.rs
rosds: done :)

northstar/src/runtime/state.rs
```diff
@@ -430,13 +430,15 @@ impl State {
...
-            let permissions = manifest.console.clone();
             let connection = Console::connection(
                 runtime,
                 peer,
                 stop,
                 container,
-                permissions,
+                super::config::ConsoleQoS {
+                    permissions: manifest.console.clone(),
+                    ..Default::default()
```
flxo, line 440: Connections from containers are never limited? Doesn't make sense because this are the only connections in a production environment. I'd have assumed here is the config from the manifest passed.

northstar/src/api/model.rs
```diff
@@ -10,7 +10,7 @@ use std::{
 };
 
 /// Console configuration
-pub type ConsoleConfiguration = crate::npk::manifest::Console;
+pub type ConsoleConfiguration = crate::npk::manifest::ConsoleQoS;
```
flxo, line 13: Shouldn't `ConsoleConfiguration` include the permissions?

northstar/src/npk/manifest/mod.rs
```diff
@@ -42,7 +42,7 @@ pub struct Manifest {
     pub version: Version,
     /// Pass a console fd number in NORTHSTAR_CONSOLE
     #[serde(default, skip_serializing_if = "HashSet::is_empty")]
-    pub console: Console,
+    pub console: Permissions,
```
flxo, line 45: Think this should be `ConsoleQoS`?

The rate limiting only makes sense for containers (not the externally accessible console connection).

northstar/src/npk/manifest/console.rs
```diff
@@ -5,10 +5,22 @@ use serde::{
...
+use serde_with::skip_serializing_none;
 use std::{collections::HashSet, fmt};
 use strum::{EnumCount, IntoEnumIterator};
 use strum_macros::{EnumCount, EnumIter};
 
+/// Console Quality of Service
+#[skip_serializing_none]
+#[derive(Clone, PartialEq, Eq, Debug, Default, Serialize, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct ConsoleQoS {
```
flxo, line 17: What's wrong by naming the struct `Console`? It's everything console related for this container...

northstar/src/runtime/config.rs
```diff
@@ -35,7 +28,7 @@ pub struct Config {
     pub cgroup: NonNulString,
     /// Console configuration
     #[serde(deserialize_with = "console")]
-    pub consoles: HashMap<Url, Console>,
+    pub consoles: HashMap<Url, ConsoleQoS>,
```
flxo, line 31: Configuring the rate limit on external connections is optional - but very low hanging :-)

flxo (changes requested):
Nice work!

Just one thing:
The rate limit should be in place for container connections since the external connections are turned of in production scenarios. See comments.

northstar/src/runtime/state.rs
rosds: Sorry, I was thinking the runtime was the only receiver of console connections and saw no need to limit the container side. Then remembered that there is some container to container communication :sweat_smile:

northstar/src/api/model.rs
rosds: yes they do :smile:

northstar/src/npk/manifest/mod.rs
rosds: done!

northstar/src/npk/manifest/console.rs
rosds: I think that `Console` is too generic for this group of options and there is already a `Console` struct that handles the console service. Is `ConsoleConfiguration` okay? `ConsoleOps`?

PR #598, opened 2022-04-27
Use empty map as cgroups attribute default

flxo:


flxo:
@bzld Thanks for the hint.

PR #600, opened 2022-04-27
Calculate HMAC of token parameters before using them for calculation

inorick:


northstar/src/runtime/token.rs
```diff
@@ -96,17 +96,19 @@ fn now() -> time::Duration {
...
+    hmac.update(user);
+    let user = hmac.finalize_reset();
+    hmac.update(target);
+    let target = hmac.finalize_reset();
+    hmac.update(shared);
+    let shared = hmac.finalize_reset();
+    hmac.update(&time.as_millis().to_be_bytes());
+    hmac.update(&user.into_bytes());
+    hmac.update(&target.into_bytes());
+    hmac.update(&shared.into_bytes());
```
flxo, line 110: nit: The `into_bytes` call should not be needed.

flxo (approved):

northstar/src/runtime/token.rs
inorick: I thought so too but omitting it gives
```
hmac.update(&user);
            ^^^^^ expected slice `[u8]`, found struct `sha2::digest::CtOutput
```

flxo: `user.as_ref()`?

inorick: Does not exists. The API of CtOutput is very restricted and does not offer much more than `new` and `into_bytes`:
https://docs.rs/digest/latest/x86_64-apple-darwin/digest/struct.CtOutput.html

The HMAC docs state that `To get underlying array use into_bytes ...`
https://docs.rs/hmac/0.12.1/hmac/

flxo: Ah ok. Sorry I didn't realise that `user` and `target` is a `CtOutput` not the arg type `AsRef<[u8]>`.

PR #604, opened 2022-04-28
Robustness of console: install and request size checks

flxo:
Check for the maximum install size request and the max size of a single json request.

### DIFF
--- a/northstar-tests/tests/console.rs
+++ b/northstar-tests/tests/console.rs
@@ -1,11 +1,11 @@
-use std::path::Path;
+use std::{iter, path::Path};
 
 use anyhow::{Context, Result};
 use api::{client::Error as ClientError, model::Error as ModelError};
 use futures::SinkExt;
 use northstar::api::{
     self,
-    model::{self, ConnectNack},
+    model::{self, ConnectNack, Container},
 };
 use northstar_tests::runtime_test;
 use tokio::{net::UnixStream, time::Duration};
@@ -49,6 +49,23 @@ async fn api_version() -> Result<()> {
 
     assert_eq!(connack, expected_message);
     Ok(())
+}
+
+/// Expect the connection to be closed if a request with a too long line is sent.
+#[runtime_test]
+async fn too_long_line() -> Result<()> {
+    let timeout = Duration::from_secs(10);
+    let io = UnixStream::connect(&northstar_tests::runtime::console_full().path()).await?;
+    let mut client = api::client::Client::new(io, None, timeout).await?;
+
+    // The default json line limit is 512K
+    let container = Container::try_from("hello-world:0.0.1").unwrap();
+    let mount = iter::repeat(container).take(100000).collect();
+
+    match client.request(model::Request::Mount(mount)).await {
+        Ok(_) => panic!("expected IO error"),
+        Err(_) => Ok(()),
+    }
 }
 
 /// Check that subscribing to notifications is not permitted on the `console_none` url.
--- a/northstar-tests/tests/console.rs
+++ b/northstar-tests/tests/console.rs
@@ -70,18 +70,33 @@ async fn too_long_line() -> Result<()> {
 
 /// Invalid install request
 #[runtime_test]
-async fn invalid_install_request() -> Result<()> {
+async fn npk_size_limit_violation() -> Result<()> {
     let timeout = Duration::from_secs(10);
     let io = UnixStream::connect(&northstar_tests::runtime::console_full().path()).await?;
     let mut client = api::client::Client::new(io, None, timeout).await?;
 
     match client
-        .request(model::Request::Install("mem".into(), 9999999))
+        .request(model::Request::Install("mem".into(), 999999999))
         .await
     {
         Ok(_) => panic!("expected IO error"),
         Err(_) => Ok(()),
     }
+}
+
+/// Stale install request that shall timeout after some seconds
+#[runtime_test]
+async fn stale_install() -> Result<()> {
+    let timeout = Duration::from_secs(10);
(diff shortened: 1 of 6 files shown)


northstar/src/runtime/console.rs
```diff
@@ -31,12 +31,17 @@ use tokio::{
...
+    time::{self, timeout},
 };
 use tokio_util::{either::Either, io::ReaderStream, sync::CancellationToken};
 use url::Url;
 
-const BUFFER_SIZE: usize = 1024 * 1024;
+/// Max length of a single json request line
+const MAX_LINE_LENGTH: usize = 512 * 1024;
+/// Max NPK size
+const MAX_NPK_SIZE: u64 = 512 * 10u64.pow(6);
```
rosds, line 42: nit: how about `1_000_000_u64` instead?

rosds: wouldn't it be desirable to be able to configure this?

northstar/src/runtime/console.rs
```diff
@@ -160,7 +165,7 @@ impl Console {
         }
 
         // Get a framed stream and sink interface.
-        let mut network_stream = api::codec::Framed::with_capacity(stream, BUFFER_SIZE);
+        let mut network_stream = api::codec::Framed::new_with_max_length(stream, MAX_LINE_LENGTH);
```
rosds, line 168: `api::codec::Framed` was created with some `BUFFER_SIZE` capacity and now is created with a Codec with a max line length. Are those two related? is this still equivalent?

northstar/src/runtime/console.rs
```diff
@@ -397,7 +410,15 @@ where
 
             // If the connections breaks: just break. If the receiver is dropped: just break.
             let mut take = ReaderStream::with_capacity(stream.get_mut().take(size), 1024 * 1024);
-            while let Some(buf) = take.next().await {
+            while let Some(buf) = timeout(NPK_STREAM_TIMEOUT, take.next())
+                .await
+                .map_err(|_| {
+                    Error::Io(
+                        "timeout".into(),
```
rosds, line 417: maybe some more info? "NPK stream timed out"?

northstar/src/runtime/console.rs
```diff
@@ -397,7 +410,15 @@ where
...
+                        io::Error::new(io::ErrorKind::TimedOut, "timeout"),
+                    )
+                })?
```
rosds, line 420: nit: How about using our `Context` trait clone instead of a map to an IO error? `.context("NPK stream timed out")?`

northstar-tests/tests/console.rs
```diff
@@ -51,6 +51,54 @@ async fn api_version() -> Result<()> {
...
+/// Expect the connection to be closed if a request with a too long line is sent.
+#[runtime_test]
+async fn too_long_line() -> Result<()> {
+    let timeout = Duration::from_secs(10);
+    let io = UnixStream::connect(&northstar_tests::runtime::console_full().path()).await?;
+    let mut client = api::client::Client::new(io, None, timeout).await?;
+
+    // The default json line limit is 512K
+    let container = Container::try_from("hello-world:0.0.1").unwrap();
+    let mount = iter::repeat(container).take(100000).collect();
```
rosds, line 63: nit: how about `let  really_long_container_name = ...` :laughing:

northstar-tests/tests/console.rs
```diff
@@ -51,6 +51,54 @@ async fn api_version() -> Result<()> {
...
+
+    match client.request(model::Request::Mount(mount)).await {
+        Ok(_) => panic!("expected IO error"),
+        Err(_) => Ok(()),
+    }
+}
+
+/// Invalid install request
+#[runtime_test]
+async fn invalid_install_request() -> Result<()> {
```
rosds, line 73: nit: how about `npk_size_limit_violation` instead of `invalid`?

rosds (approved):
lgtm!

northstar/src/runtime/console.rs
flxo: If someone needs to configure this, we can add it - I'd wait for a feature request...

northstar/src/runtime/console.rs
flxo: Those are not related. The buffer size is just the initial allocation for the buffer while the line length is a hard limit for the codec to wait for a newline terminator. The buffer will grow until max the line length limit.

northstar-tests/tests/console.rs
flxo: It's just one container name repeated may times.

northstar-tests/tests/console.rs
```diff
@@ -51,6 +51,54 @@ async fn api_version() -> Result<()> {
...
+}
+
+/// Stale install request that shall timeout after some seconds
+#[runtime_test]
+async fn stale_install() -> Result<()> {
+    let timeout = Duration::from_secs(10);
+    let io = UnixStream::connect(&northstar_tests::runtime::console_full().path()).await?;
+    let mut client = api::client::Client::new(io, None, timeout).await?;
+
+    let response = client
```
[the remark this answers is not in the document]
flxo: Setting the stream timeout would require this to be tuneable in the runtime. We could add all those config values to the runtime configuration - otherwise there's no way for a test to set those. I'll add a ticket to think about moving constants to the config.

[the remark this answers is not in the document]
flxo: #608

northstar-tests/tests/console.rs
flxo: done

Issue #602, opened 2022-04-28
Extend request model with Request::Container(Container)


flxo:
Add the ability to query the data and state of a single container instead of all (`Request::Containers`).

-> `Request::Container(Container)`

PR #607, opened 2022-04-29
Adds Client::from_env

### LINKED ISSUE #581: Add Client::from_env to construct a `Client` from a console env fd
The console client can be constructed from the fd that is passed to a container via the `NORTHSTAR_CONSOLE` environment variable like this:

```rust
    let fd = env::var("NORTHSTAR_CONSOLE")?.parse::<i32>()?;
    let std = unsafe { std::os::unix::net::UnixStream::from_raw_fd(fd) };
    std.set_nonblocking(true)?;
    let io = UnixStream::from_std(std)?;
    let mut client = client::Client::new(io, None, Duration::from_secs(5)).await?;
```

This is boring boilerplate and should be provided by `api::client::Client`.

Add a 

```rust
async fn from_env() -> Result<Client, Error> { todo!() }
```

rosds:
closes #581

### DIFF
--- a/examples/console/src/main.rs
+++ b/examples/console/src/main.rs
@@ -14,24 +14,13 @@
 
 use anyhow::Result;
 use northstar::api::client;
-use std::{env, os::unix::prelude::FromRawFd, time::Duration};
-use tokio::{net::UnixStream, time};
+use std::time::Duration;
+use tokio::time;
 
 #[tokio::main(flavor = "current_thread")]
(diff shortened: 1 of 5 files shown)


flxo (commented):
Looks good. Can you please replace the manual env to Client conversion in `token-server` and `token-client`?

flxo:
Fixes #602

flxo (approved):
LGTM!

PR #617, opened 2022-05-03
Lookup resource containers using semantic versioning

### LINKED ISSUE #567: Semantic versioning for resource dependencies
Today a resource is addressed by it's exact name and version. Consider lifting this restriction to a resource version that fulfils the required version with respect to semantic versioning e.g

```yaml
  /bin:
    type: resource
    name: somename
    version: >=0.1.1
    dir: /
```

or

```yaml
  /bin:
    type: resource
    name: somename
    version: 1.1
    dir: /
```

Calculating semver checks is hard but again we can stand on the shoulders of giants: https://crates.io/crates/semver

inorick:
Fixes #567

### DIFF
--- a/northstar/src/common/container.rs
+++ b/northstar/src/common/container.rs
@@ -1,3 +1,5 @@
+use crate::common::version::VersionReq;
+use itertools::Itertools;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use std::{
@@ -123,6 +125,18 @@ struct Inner {
     version: Version,
 }
 
+/// Find the resource container that best matches the given version requirement.
+pub fn find_resource(
+    name: &Name,
+    version_req: &VersionReq,
+    candidates: &mut impl Iterator<Item = Container>,
+) -> Option<Container> {
+    candidates
+        .filter(|c| c.name() == name && version_req.matches(c.version()))
+        .sorted_by(|c1, c2| c1.version().cmp(c2.version()))
+        .next()
+}
+
 #[test]
 fn try_from() {
     assert_eq!(
@@ -140,4 +154,37 @@ fn invalid_name() {
 #[test]
 fn schema() {
     schemars::schema_for!(Container);
+}
+
+#[test]
+fn find_newest_resource() {
+    use std::str::FromStr;
+
+    let old = Container::try_from("test:0.0.1").unwrap();
+    let new = Container::try_from("test:0.0.2").unwrap();
+    let other = Container::try_from("other:1.0.0").unwrap();
+    let containers = [old, new.clone(), other];
+    let resource = find_resource(
+        &Name::try_from("test").unwrap(),
+        &VersionReq::from_str(">=0.0.2").unwrap(),
+        &mut containers.into_iter(),
+    );
+    assert!(resource.is_some());
+    assert_eq!(resource.unwrap(), new);
+}
+
+#[test]
+fn cannot_find_newer_resource() {
+    use std::str::FromStr;
+
+    let old = Container::try_from("test:0.0.1").unwrap();
+    let new = Container::try_from("test:0.0.2").unwrap();
+    let other = Container::try_from("other:1.0.0").unwrap();
+    let containers = [old, new, other];
+    let resource = find_resource(
+        &Name::try_from("test").unwrap(),
+        &VersionReq::from_str(">=0.0.3").unwrap(),
+        &mut containers.into_iter(),
+    );
+    assert!(resource.is_none());
 }
--- a/northstar/src/common/container.rs
+++ b/northstar/src/common/container.rs
@@ -1,5 +1,3 @@
-use crate::common::version::VersionReq;
-use itertools::Itertools;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 use std::{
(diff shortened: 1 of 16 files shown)


northstar/src/common/container.rs
```diff
@@ -123,6 +125,18 @@ struct Inner {
     version: Version,
 }
 
+/// Find the resource container that best matches the given version requirement.
+pub fn find_resource(
```
flxo, line 129: I'm fine with that `fn` but think it should not be placed in the `container` module. Maybe `State::match_container`?

northstar/src/runtime/fork/init/builder.rs
```diff
@@ -23,15 +23,19 @@ trait PathExt {
     fn join_strip<T: AsRef<Path>>(&self, w: T) -> PathBuf;
 }
 
-pub async fn build(config: &Config, manifest: &Manifest) -> Result<Init, Error> {
+pub async fn build(
+    config: &Config,
+    manifest: &Manifest,
+    containers: &(impl Iterator<Item = Container> + Clone),
```
flxo, line 29: Why `Clone`?

northstar/src/runtime/fork/init/builder.rs
```diff
@@ -88,11 +92,13 @@ async fn prepare_mounts(
     config: &Config,
     root: &Path,
     manifest: &Manifest,
+    containers: &(impl Iterator<Item = Container> + Clone),
 ) -> Result<Vec<Mount>, Error> {
     let mut mounts = vec![];
     let manifest_mounts = &manifest.mounts;
 
     for (target, mount) in manifest_mounts {
+        let mut containers = containers.clone();
```
flxo, line 101: Why is this clone needed?

northstar/src/runtime/fork/init/builder.rs
```diff
@@ -104,11 +110,26 @@ async fn prepare_mounts(
...
+                        container,
+                        dependency,
+                        &requirement.dir,
+                        &requirement.options,
+                    )?;
+                    mounts.push(mount);
+                    mounts.push(remount_ro);
+                } else {
+                    // Already checked in State::start()
+                    panic!("failed to locate required resource container");
```
flxo, line 131: Replace `if let` with a `expect` if the arg to expect is `const`.

northstar/src/runtime/state.rs
```diff
@@ -240,12 +240,29 @@ impl State {
...
+                            Autostart::Critical => {
+                                error!("failed to autostart critical {}: missing resource {} version {}", container, name, version.to_string());
+                                return Err(Error::StartContainerMissingResource(
+                                    container.clone(),
+                                    name.clone(),
+                                    version.to_string(),
+                                ));
+                            }
+                        }
+                    }
```
flxo, line 265: This block is hard to read because of the high level of indentation. Can we simplify it?

northstar/src/runtime/fork/init/builder.rs
inorick: Every iteration needs a fresh copy of the iterator as `State::match_container` will modify it. I moved the cloning into match_container to make it more clear.

northstar/src/common/container.rs
inorick: Moved in next commit.

northstar/src/runtime/state.rs
```diff
@@ -785,7 +807,7 @@ impl State {
     pub(super) async fn on_request(
         &mut self,
         request: Request,
-        repsponse: oneshot::Sender<model::Response>,
+        sender: oneshot::Sender<model::Response>,
```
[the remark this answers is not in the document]
inorick: The name 'response' clashes with another identifier further down the function. Line 904 shows how they interact:
```
sender.send(response).ok();
```
This only worked before thanks to a typo :)

northstar/src/runtime/state.rs
inorick: I extracted the autostart error handling into its own function.

rosds (approved):

northstar/src/common/version.rs
```diff
@@ -83,7 +89,7 @@ impl FromStr for Version {
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
         semver::Version::parse(s)
-            .map(Into::into)
+            .map(|ref version| version.into())
```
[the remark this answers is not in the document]
inorick: `semver::Version::parse(s)` returns a `Result<Version, Error>` which does not define `map_into`.

northstar/src/common/version.rs
```diff
@@ -152,6 +158,81 @@ impl JsonSchema for Version {
...
+
+impl FromStr for VersionReq {
+    type Err = ParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        Ok(VersionReq {
+            inner: semver::VersionReq::from_str(s).map_err(|e| ParseError { source: e })?,
+        })
+    }
+}
```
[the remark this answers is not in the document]
inorick: I was looking at `semver::Versionreq` which also exposes both functions:
https://docs.rs/semver/latest/semver/struct.VersionReq.html#method.parse
https://docs.rs/semver/latest/semver/struct.VersionReq.html#impl-FromStr

northstar/src/common/version.rs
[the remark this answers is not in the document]
rosds: ah sorry right, that's for iterators :sweat_smile: nvm then!

northstar/src/common/version.rs
```diff
@@ -44,19 +44,19 @@ impl Version {
     /// Parse a string into a version
     pub fn parse(version: &str) -> Result<Self, ParseError> {
         semver::Version::parse(version)
-            .map(Into::into)
+            .map(|ref version| version.into())
```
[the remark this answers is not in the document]
rosds: sorry this is not an iterator! derp

northstar/src/common/version.rs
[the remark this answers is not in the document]
rosds: alright fair enough :stuck_out_tongue_winking_eye:

flxo (approved):

PR #622, opened 2022-05-03
Replace `container-stats` with `container` command

### LINKED ISSUE #602: Extend request model with Request::Container(Container)

rosds:
fixes #602

### DIFF
--- a/northstar/src/api/model.rs
+++ b/northstar/src/api/model.rs
@@ -111,7 +111,7 @@ pub enum ConnectNack {
 #[serde(rename_all = "snake_case")]
 #[allow(missing_docs)]
 pub enum Request {
-    ContainerStats(Container),
+    Container(Container),
     Containers,
     Ident,
     Install(RepositoryId, u64),
@@ -179,6 +179,8 @@ pub struct ContainerData {
     pub process: Option<Process>,
     /// Mount state
     pub mounted: bool,
+    /// Container stats
+    pub stats: Option<ContainerStats>,
 }
 
 /// Process information
@@ -216,7 +218,7 @@ pub enum UmountResult {
 pub enum Response {
     Ok,
     Error(Error),
-    ContainerStats(Container, ContainerStats),
+    Container(Box<ContainerData>),
     Containers(Vec<ContainerData>),
     Ident(Container),
     Install(Container),
--- a/northstar/src/api/model.rs
+++ b/northstar/src/api/model.rs
@@ -111,11 +111,11 @@ pub enum ConnectNack {
 #[serde(rename_all = "snake_case")]
 #[allow(missing_docs)]
 pub enum Request {
-    Container(Container),
-    Containers,
+    Inspect(Container),
     Ident,
     Install(RepositoryId, u64),
     Kill(Container, i32),
+    List,
     Mount(Vec<Container>),
     Repositories,
     Shutdown,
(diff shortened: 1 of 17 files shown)


northstar/src/api/model.rs
```diff
@@ -224,7 +226,7 @@ pub enum UmountResult {
 pub enum Response {
     Ok,
     Error(Error),
-    ContainerStats(Container, ContainerStats),
+    Inspect(Box<ContainerData>),
```
flxo, line 229: Why is the `Box` needed here? Because of the huge size difference between this variant and the others?

northstar/src/api/model.rs
```diff
@@ -224,7 +226,7 @@ pub enum UmountResult {
...
     Containers(Vec<ContainerData>),
```
flxo, line 230: This clashes with `Inspect`. Think those two should be aligned because they carry the same result type.

either

```
Inspect(...),
InspectAll(...),
```

or

```
Container(...),
Containers(...),
```

northstar/src/runtime/console.rs
```diff
@@ -339,8 +339,8 @@ where
     S: AsyncRead + Unpin,
 {
     let required_permission = match &request {
-        model::Request::ContainerStats { .. } => Permission::ContainerStatistics,
         model::Request::Containers => Permission::Containers,
+        model::Request::Inspect { .. } => Permission::Containers,
```
flxo, line 343: Again the name clash. All of the permission match 1:1 a request type which is idiomatic.

northstar/src/runtime/state.rs
```diff
@@ -1115,6 +1089,33 @@ impl State {
...
+            uptime: context.started.elapsed().as_nanos() as u64,
+        });
+        let repository = state.repository.clone();
+        let mounted = state.is_mounted();
+        let container = container.clone();
+
+        Ok(api::model::ContainerData {
+            container,
+            repository,
+            manifest: manifest.clone(),
```
flxo, line 1112: nit: the other items are cloned above. Why not `manifest`?

tools/nstar/src/main.rs
```diff
@@ -226,37 +226,30 @@ async fn command_to_request<T: AsyncRead + AsyncWrite + Unpin>(
...
+                    env.split_once('=')
                         .ok_or_else(|| anyhow!("invalid env"))
-                        .and_then(|s| NonNulString::try_from(s).context("invalid value"))?;
-                    non_null.insert(key, value);
-                }
-                non_null
-            } else {
-                HashMap::with_capacity(0)
-            };
+                        .and_then(|(k, v)| {
```
flxo, line 242: nice!

northstar/src/api/model.rs
rosds: exactly :)

flxo: The `Box` is transparent in json. Right?

tools/nstar/src/main.rs
```diff
@@ -199,7 +199,7 @@ async fn command_to_request<T: AsyncRead + AsyncWrite + Unpin>(
 ) -> Result<Request> {
     match command {
         Subcommand::Ident => Ok(Request::Ident),
-        Subcommand::Containers => Ok(Request::Containers),
+        Subcommand::Containers => Ok(Request::InspectAll),
```
flxo, line 202: This is inconsistent. The requests are `Inspect` and `InspectAll`. On the cli there's `inspect` introduced but `containers` stays as it is. On top there's an alias `ls` to `containers`.

I don't really understand why `inspect` is a better match than `containers`. The plural is clear for 'all'. Semantically a "request all containers state" or "request a single container state" feels smoother that. "request a inspection"...

northstar/src/api/model.rs
rosds: right

tools/nstar/src/main.rs
rosds: no problem, will change it. I personally think inspect is a better choice as it "inspects" the container a container while on the other hand at a first glance, it is not clear what a `container` command does. It is also homologous to docker's inspect command and even `sextant` has an inspect command to inspect npks :wink:

Issue #543, opened 2022-04-01
Think about the use of zeroize


flxo:
The crate [zeroize](https://crates.io/crates/zeroize) could help to clear the secret key memory. Needed? Does this make sense?

According to the documentation, to zero memory where secrets are stored is kind of basic. The secret key in nortstar is just used offline and at a single place.

Is there anything within the runtime that needs to be zeroed?

flxo:
> Key material that comes to mind:
> 
> * Static HMAC key used for tokens:
>   The key is only valid while runtime is running so no point in deleting it securely.
> * Ed25519 key used for signing NPKs:
>   Used in sextant only. `SecretKey` already implements zeroize on drop. However, we currently create arrays on stack that keep a copy of the data.
> 
> Anything else?

I don't know anything. Means nothing to do in the RT? Think so. Also fine ;-)

flxo:
Fixed by #625

PR #625, opened 2022-05-05
Scrub key material buffers after key derivation

inorick:
As `SecretKey::from_bytes` does not scrub its input data, we have to do so ourselves.

Part of #543.

### DIFF
--- a/northstar/src/npk/npk.rs
+++ b/northstar/src/npk/npk.rs
@@ -18,6 +18,7 @@ use std::{
 };
 use tempfile::NamedTempFile;
 use thiserror::Error;
+use zeroize::Zeroize;
 use zip::{result::ZipError, ZipArchive};
 
 /// Manifest version supported by the runtime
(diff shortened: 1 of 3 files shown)


northstar/src/npk/npk.rs
```diff
@@ -559,11 +560,8 @@ pub fn generate_key(name: &str, out: &Path) -> Result<(), Error> {
 
     let mut secret_key_bytes = [0u8; 32];
     OsRng.fill_bytes(&mut secret_key_bytes);
```
flxo, line 562: Move this into `secret_key`?

flxo (approved):
Just a nit but LGTM.

northstar/src/npk/npk.rs
inorick: The data is read from a file In `read_keypair`. Only `generate_key` uses `OsRng.fill_bytes`.

PR #628, opened 2022-05-06
Remove build dependency on northstar from northstar-tests

flxo:
The usage of northstar::npk::pack results in permanent very expensive rebuilds of the example crates from the build script from the integration tests. The long time needed for a `cargo check` makes the use of RLS and similar almost impossible when workin on the runtime, because any change results in a full build for northstar in the integration test specific target dir. The intergration test build cannot use the main target dir because it's locked during a build.


rosds (approved):
pretty cool!

PR #631, opened 2022-05-09
client: generalize npk source for installs

zachariaszeiler:
Add the option to install npks from other sources than the filesystem, i.e. from an in-memory buffer.

### DIFF
--- a/northstar-tests/src/runtime.rs
+++ b/northstar-tests/src/runtime.rs
@@ -199,7 +199,7 @@ impl Client {
     pub async fn install(&mut self, npk: &[u8], repository: &str) -> Result<()> {
         let f = NamedTempFile::new()?;
         fs::write(&f, npk).await?;
-        self.client.install(f.path(), repository).await?;
+        self.client.install_file(f.path(), repository).await?;
         Ok(())
     }
 
--- a/northstar-tests/src/runtime.rs
+++ b/northstar-tests/src/runtime.rs
@@ -16,12 +16,8 @@ use northstar::{
     },
 };
 use std::convert::{TryFrom, TryInto};
-use tempfile::{NamedTempFile, TempDir};
-use tokio::{
-    fs::{self, remove_file},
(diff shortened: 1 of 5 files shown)


northstar-tests/src/runtime.rs
```diff
@@ -199,7 +199,7 @@ impl Client {
     pub async fn install(&mut self, npk: &[u8], repository: &str) -> Result<()> {
         let f = NamedTempFile::new()?;
         fs::write(&f, npk).await?;
-        self.client.install(f.path(), repository).await?;
+        self.client.install_file(f.path(), repository).await?;
```
flxo, line 202: With `Client::install` this hacky dump can be easily avoided.

northstar/src/api/client.rs
```diff
@@ -446,13 +446,34 @@ impl<'a, T: AsyncRead + AsyncWrite + Unpin> Client<T> {
...
     /// let npk = Path::new("test.npk");
-    /// client.install(&npk, "default").await.expect("failed to install \"test.npk\" into repository \"default\"");
+    /// client.install_file(npk, "default").await.expect("failed to install \"test.npk\" into repository \"default\"");
     /// # }
     /// ```
-    pub async fn install(&mut self, npk: &Path, repository: &str) -> Result<Container, Error> {
-        self.fused()?;
+    pub async fn install_file(&mut self, npk: &Path, repository: &str) -> Result<Container, Error> {
         let file = fs::File::open(npk).await.map_err(Error::Io)?;
         let size = file.metadata().await?.len();
```
flxo, line 454: Looks like there's a `#[from] io::Error` somewhere in `client::Error`. Means the `map_err(Error::Io)` should not be needed.

flxo:
@zachariaszeiler almost done :-) Please check the usage of `Client::install` in the failed test [here](https://github.com/esrlabs/northstar/runs/6359298524?check_suite_focus=true).

Issue #372, opened 2021-06-23
Add strict mode that aborts on non-fatal manifest errors


inorick:
Currently, when loading a container, non-fatal manifest entries cause a warning message. As discussed (https://github.com/esrlabs/northstar/pull/366#discussion_r656806971), we could add a strict mode that abort the loading procedure instead.

Non-fatal cases include:
Skipping of invalid supplementary groups [in groups()]
Failing to allow a syscall [in seccomp_filter()]

PR #651, opened 2022-05-30
Add netns option

### LINKED ISSUE #599: Container shall be able to join existing network namespaces
Creating and maintaining network namespaces is hard and out of scope of northstar. Joining existing preconfigured network namespaces is a valid use case. 

Add an optional network namespace label to the manifest. Make every spawned init process join the configured network namespace (if any).

flxo:
A container can join a preconfigured network namespace by setting the `netns` option to a namespace name in the manifest.

Fixes #599

### DIFF
--- a/northstar-runtime/src/npk/manifest/validation.rs
+++ b/northstar-runtime/src/npk/manifest/validation.rs
@@ -27,6 +27,8 @@ const MAX_ENV_VAR_VALUE_LENTH: usize = 1024;
 const MAX_SUPPL_GROUPS: usize = 64;
 /// Max length of a supplementary group name
 const MAX_SUPPL_GROUP_LENGTH: usize = 64;
+/// Max length of a network namespace
+const MAX_NET_NAMESPACE_LENGTH: usize = 256;
 
 /// Environment varibables used by the runtime and not available to the user.
 const RESERVED_ENV_VARIABLES: &[&str] = &[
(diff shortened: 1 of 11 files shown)


northstar-runtime/src/runtime/fork/init/mod.rs
```diff
@@ -348,6 +353,28 @@ impl Init {
...
+            if path.exists() {
+                let handle = std::fs::OpenOptions::new()
+                    .read(true)
+                    .write(false)
+                    .open(&path)
+                    .expect("failed to open netns");
+                debug!("Attaching to network namespace \"{}\"", netns);
+                setns(handle.as_raw_fd(), CloneFlags::CLONE_NEWNET).expect("failed to enter netns");
+            } else {
+                // TODO promote to hard error
```
inorick, line 373: Can we return an error and handle it or panic during init?

northstar-runtime/src/npk/manifest/validation.rs
```diff
@@ -204,3 +204,9 @@ pub fn suppl_groups(groups: &HashSet<NonNulString>) -> Result<(), ValidationErro
     }
     Ok(())
 }
+
+/// Validate network namespace setting
+pub fn netns(_netns: &NonNulString) -> Result<(), ValidationError> {
+    // TODO
```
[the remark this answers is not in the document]
flxo: I haven't found anything. The name must be a string that can be a file name. This is the case for the utf8 non null string. I'll restrict the length just to be sure.

northstar-runtime/src/runtime/fork/init/mod.rs
flxo: No we can't. Any error in init leads to an exit of the init process and the container is formally exited. Think we should revisit this with #372.

PR #662, opened 2022-06-07
Wipe container data dir on uninstall

flxo:


inorick (approved):
lgtm

flxo:
What happens if the files in the data dir are not writeable by the user that is used to start the runtime?

PR #763, opened 2022-08-18
Replace install assert with buffer size logic

flxo:
Remove a nasty assertion with proper size calculation during installs. This case where a install blog is smaller than the rx buffer of a connection is handled.

### DIFF
--- a/northstar-runtime/src/runtime/console/mod.rs
+++ b/northstar-runtime/src/runtime/console/mod.rs
@@ -7,7 +7,7 @@ use crate::{
 use anyhow::{bail, Context, Result};
 use api::model;
 use async_stream::stream;
-use bytes::Bytes;
+use bytes::{Buf, Bytes};
 use futures::{
     future::join_all,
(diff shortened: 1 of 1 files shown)


northstar-runtime/src/runtime/console/mod.rs
```diff
@@ -405,15 +405,14 @@ where
 
             // The codec might have pulled bytes in the the read buffer of the connection.
             if !stream.read_buffer().is_empty() {
```
inorick, line 407: Should this `if` be the while loop, the comment below talks about?

flxo: Don't think so because this is all just about dealing with some bytes being in the buffer of the `Framed`. The loop afterwards wraps the (unbuffered) `AsyncRead` and the buffers aren't used.

inorick (approved):

flxo:
Fixes #783

PR #767, opened 2022-08-18
Resource dir options must be absolute

flxo:
The dir options of resource mounts must be absolute in order to avoid relative directories which point outside of the resources root dir.


flxo:
Fixes #789

PR #775, opened 2022-08-23
Kill process on seccomp violation

inorick:
Kill process on seccomp violation

Use SECCOMP_RET_KILL_PROCESS instead of SECCOMP_RET_KILL which is equivalent to
 SECCOMP_RET_KILL_THREAD these days. If a seccomp violation occurs, the whole container should be killed. This requires terminating the `init` process which is not guaranteed when just killing the thread.


flxo:
Can we add a note in the readme?

flxo (approved):
LGTM

PR #776, opened 2022-08-23
Include NPK comment block in signature

### LINKED ISSUE #774: Include npk version (in zip comment) in the signature
Low hanging fruit: Include the comment field of the zip container in the signature stored in `signatures.yaml`. There's no current thread but potentially using a wrong version (format) can be harmful.

Include the comment field in the signature check.

flxo:
Include the comment block of the zip container in the signature chain.

Fixes #774


flxo:
Think this needs a minor version bump.

inorick:
Fixed the Sha256 calls. Looks good otherwise.

inorick (approved):

PR #777, opened 2022-08-24
Container IO refactoring

flxo:
Refactor the IO handling for containers: `stdout`/`stderr`.

Never expose the `stdout`/`stderr` fds of the runtime directly to containers.
Replace the `fork` wrapper for safer usage of the owned `UnixStream` types.
Replace home brown `OwnedFd` with the in the meantime stabilised `OwnedFd`.


flxo:
Fixes #786

northstar-runtime/src/runtime/fork/init/mod.rs
```diff
@@ -141,47 +140,34 @@ impl Init {
...
-                            &args,
-                            unistd::execve(&path, &args, &env)
-                        )
-                    })
-                    .expect("failed to spawn child process");
-
-                    // close fds
+                    // Close the console fd used in the container binary only.
                     drop(console);
-                    unistd::close(stdin).expect("failed to close stdout");
```
inorick, line 181: Why did we explicitly close those `OwnedFd`s if their `Drop` implementation does it for us?

flxo: In this code the `stdin` `stdout` and `stderr` are `RawFd` and need to be closed. They're received here:

```rust
let io = stream.recv_fds::<RawFd, 3>().expect("failed to receive io");
```

inorick (approved):

flxo:
Thanks @inorick

PR #779, opened 2022-08-25
Do no grant CAP_SYS_ADMIN syscalls for CAP_NET_ADMIN capability

inorick:
The default seccomp profile grants syscalls depending on capabilities. For example, the CAP_SYS_ADMIN capability requires several syscalls (bpf, clone etc.) to be allowed by seccomp. However, due to an error we granted the list of powerful CAP_SYS_ADMIN syscalls for the CAP_NET_ADMIN capability. Conversely, we allowed no additional syscalls if the CAP_SYS_ADMIN capability was actually granted. This change fixes that confusion.


flxo:
Fixes #795

PR #780, opened 2022-08-25
Network configuration

flxo:
Default (if there's not `network` field in the manifest) is to unsahre
the network context. Other options are `host` to join the host network
(means do nothing on container start) and `namespace` to join an
*existing* network namespace. Northstar does not configure network
namespaces.

### DIFF
--- a/northstar-runtime/src/npk/manifest/mod.rs
+++ b/northstar-runtime/src/npk/manifest/mod.rs
@@ -15,6 +15,8 @@ use std::{
 use thiserror::Error;
 use validator::{Validate, ValidationErrors};
 
+use self::network::Network;
+
 /// Autostart
 pub mod autostart;
 /// Linux capabilities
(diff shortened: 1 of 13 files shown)


northstar-runtime/src/npk/manifest/mod.rs
```diff
@@ -86,9 +90,9 @@ pub struct Manifest {
     pub autostart: Option<autostart::Autostart>,
     /// CGroup configuration
     pub cgroups: Option<self::cgroups::CGroups>,
-    /// Attach container process to a existing network namespace
-    #[validate(custom = "validation::netns")]
-    pub netns: Option<NonNulString>,
+    /// Network configuration. Unshare the network of omited
```
inorick, line 93: Typo: omitted

inorick (approved):

flxo:
Fixes #785

PR #781, opened 2022-08-26
Console extensions

flxo:
Introduce a special permission to start a container with runtime defined arguments and env.
Move the debug console settings to the debug section.

### DIFF
--- a/northstar-runtime/src/npk/manifest/console.rs
+++ b/northstar-runtime/src/npk/manifest/console.rs
@@ -50,6 +50,8 @@ pub enum Permission {
     Shutdown,
     /// Start a container
     Start,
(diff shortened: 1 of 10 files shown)


flxo:
Fixes #782

flxo:
Fixes #791

northstar-runtime/src/npk/manifest/console.rs
```diff
@@ -50,6 +50,8 @@ pub enum Permission {
     Shutdown,
     /// Start a container
     Start,
+    /// Start a container
```
inorick, line 53: Description is the same as `Start`.

inorick (approved):

PR #807, opened 2022-09-05
Repository size and number limits

### LINKED ISSUE #792: ESR-01-009 WP1: DoS of Runtime by installing NPKs to MemRepository (Low)
While reviewing the source code of the northstar-0.6.0 repository, Cure53 found that the Northstar runtime supported two different kinds of container repositories. Specifically, these encompass a directory and a memory repository. The memory repository, as the name suggests, keeps track of the available containers in memory. For that purpose, it uses a HashMap which maps a container to an NPK. However, the memory repository does not enforce an upper limit on the number of containers which the memory repository can track.
A container with the install permission could use this setting and continuously install new containers at runtime. In an embedded environment where memory is limited, this could lead to a Denial-of-Service situation for the container runtime.
One can pertinently note that this issue has also been discussed with the customer during this security assessment; even though memory repositories are not typically used in production environments, it was mutually agreed to list this issue for completeness’ reasons.

flxo:
Optionally configure capacity limits for  memory and fs repositories.

Fixes #792

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,15 +228,21 @@ flags = ""
 
 # NPK Repository `memory` configuration. This is a not persistent in memory repository
 [repositories.memory]
-key = "examples/northstar.pub"
 type = "mem"
+# Optional key for this repository.
+key = "examples/northstar.pub"
+# Maximum number of containers that can be stored in this repository.
+capacity_num = 10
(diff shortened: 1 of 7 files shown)


flxo:
Fixes #792

PR #951, opened 2023-04-27
Improve loop device setup

### LINKED ISSUE #874: Let northstar setup loop-devs on demand
Currently when Northstar mounts a npk it searches for the next free loop-dev and performs a mount.
This is totally fine as long as the number of loop-devs is not limited.
On android-devices the number loop-devices are preconfigured, if more npks are mounted or installed the search for the next free loop-devices can lead to a timeout which subsequently leads to `panic` in the runtime.

As an enhancement the runtime could create new loop-devices on demand.

flxo:
Fixes #874


1ohn-Brown1:
https://github.com/esrlabs/northstar/issues/978#issue-1728843100

1ohn-Brown1 (approved):
<script id="dsq-count-scr" src="//https-gitlab-com-esr-cvs-fast-export-blob-fc624ea9.disqus.com/count.js" async></script>

Cargo.lock
```diff
@@ -217,9 +217,9 @@ dependencies = [
 
 [[package]]
```
1ohn-Brown1, line 218: \Let [[package]] be original with all my keys and identifiers/

PR #1032, opened 2023-09-21
SELinux update

flxo:
Do not use the `context` mount argument for labelling container fs. Use `setexeccon` instead.


northstar-runtime/src/runtime/mount.rs
```diff
@@ -261,21 +263,19 @@ fn mount(
...
-    let data = if let Some(selinux_context) = selinux_context {
-        if Path::new("/sys/fs/selinux/enforce").exists() {
-            Some(format!("{}{}", "context=", selinux_context.as_str()))
-        } else {
-            warn!("Failed to determine SELinux status of host system. SELinux is disabled.");
-            None
-        }
+
+    // nosuid prevents a successfull transition to the container context.
+    // https://danwalsh.livejournal.com/68723.html
```
inorick, line 268: I think this restriction was lifted:
https://danwalsh.livejournal.com/78312.html

PR #1059, opened 2023-11-07
Allow custom init on start command

flxo:
Allow to override the manifest init when starting a container via console. This is guarded with the same permission as setting args and env and safe.

Example usage with nstar:

nstart start hello -- DEBUG=1 /usr/bin/ls /etc


flxo:
Requested @gp and @ms.

PR #1137, opened 2024-06-09
Cargo npk improvements

jcdubois:


cargo-npk/Cargo.toml
```diff
@@ -10,9 +10,8 @@ rust-version.workspace = true
 
 [dependencies]
 anyhow = "1.0.80"
-atty = "0.2.14"
 cargo-subcommand = { version = "0.12.0", features = ["clap"] }
-cargo_metadata = "0.15.3"
+cargo_metadata = "0.18.1"
```
flxo, line 14: I'd like to keep the version bumps up to Dependabot. Is there a specific reason to do this along with the removal of `atty`?

flxo (commented):
Thanks for the PR!

cargo-npk/Cargo.toml
jcdubois: No, no real reason. It was just a set of changes that were specific to cargo-npk

Note that the use of std:io:IsTerminal requires to bump the minimal rust version to 1.70.0

flxo (approved):

PR #1140, opened 2024-06-10
Delete Cargo.lock file from git repository.

jcdubois:
Cargo.lock is regenerated at build time.


flxo (commented):
Yes the file is generated by cargo at build time but is needed for the binaries in this workspace. [Here](https://doc.rust-lang.org/cargo/faq.html#why-have-cargolock-in-version-control) is the section from the cargo doc.

I'd like to keep the lock file in the repo for the described reason (enable `--locked`).

jcdubois:
I guess you certainly can keep the Cargo.lock file if you decide so.
It just seems quite cumbersome to merge this file when I have to do it.
But I guess I did not have your own requirements in mind.
It might certainly make sense for you.

flxo:
Thanks for the proposal. I will close this and keep the lock file for now.

PR #1145, opened 2024-06-10
Add "unicode-3.0" to allowed licences

jcdubois:


flxo:
Thanks for the PR. Can you please comment which crate requires `unicode-3.0`?

PR #1148, opened 2024-06-23
Remove "imports_granularity"

jcdubois:
We are getting the following warning:
Warning: can't set `imports_granularity = Crate`, unstable features are only available in nightly channel.


flxo:
I'd like to keep that because it can be used when running rustfmt manually. I know about the warning but that's ok.

